PR #149, opened 2020-09-01
[Fix]Move Budget request parameter formation to method

Bingjiling:
Issue #, if available:

Description of changes:

Move Budget API parameter structure to methods in BudgetService class
Remove budget template file 
Fix CF template indent error

Checklist: 


addons/addon-base-raas/packages/base-raas-services/lib/budgets/budgets-service.js
```diff
@@ -177,22 +173,53 @@ class BudgetsService extends Service {
...
+        Subscribers: [
+          {
+            Address: newBudgetConfig.notificationEmail,
+            SubscriptionType: 'EMAIL',
+          },
+        ],
+      };
       notifications.push(notificationCopy);
     });
     return notifications;
```
dpandya, line 225: Minor. Since you are translating (mapping) thresholds to notifications it may be better to use `_.map` instead of `_.forEach`. This way, you don't need to keep `notifications` array and push to it. 

Something like,

```
_formNotificationObjects(newBudgetConfig) {
    const thresholds = _.get(newBudgetConfig, 'thresholds', []);
    return _.map(thresholds, threshold => {
      return {
        Notification: {
          ComparisonOperator: 'GREATER_THAN',
          NotificationType: 'ACTUAL',
          Threshold: threshold,
          ThresholdType: 'PERCENTAGE',
        },
        Subscribers: [
          {
            Address: newBudgetConfig.notificationEmail,
            SubscriptionType: 'EMAIL',
          },
        ],
      };
    });
  }
```

Bingjiling: Good call! Updating it now.

PR #151, opened 2020-09-08
RStudio for SC

SanketD92:
Issue #, if available:
GALI-450

Description of changes:
1. Added the RStudio product template to the default SC portfolio (reused the built-in template and added SC-related output variables)
2. Reuse getRStudioUrl mechanism from the built-in environmentUrlService class, and integrated it into environmentScConnectionService. This provides the 'Connect' button in the UI that opens the RStudio authorized URL tab.

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-connection-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-connection-service.test.js
@@ -42,10 +42,20 @@ const EnvironmentSCServiceMock = require('../environment-sc-service');
 jest.mock('../environment-sc-keypair-service');
 const EnvironmentScKeyPairServiceMock = require('../environment-sc-keypair-service');
 
+jest.mock('request-promise-native', () => jest.fn(() => 'keyExponent:keyModulus'));
+jest.mock('querystring', () => ({
+  encode: jest.fn(() => 'SampleEncodedParams'),
+}));
+jest.mock('@aws-ee/base-services/lib/helpers/rstudio-encryptor', () => ({
+  encrypt: jest.fn(() => 'SampleEncryptedKey'), // This will be used as sample sign-in params for RStudio URL
+}));
+
 const EnvironmentScConnectionService = require('../environment-sc-connection-service');
 
 describe('EnvironmentScConnectionService', () => {
   let service = null;
+  let envDnsService = null;
+
   beforeEach(async () => {
     const container = new ServicesContainer();
     container.register('jsonSchemaValidationService', new JsonSchemaValidationService());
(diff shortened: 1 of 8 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-connection-service.test.js
```diff
@@ -0,0 +1,110 @@
...
+
+jest.mock('../environment-sc-service');
+const EnvironmentSCServiceMock = require('../environment-sc-service');
+
+jest.mock('../environment-sc-keypair-service');
+const EnvironmentScKeyPairServiceMock = require('../environment-sc-keypair-service');
+
+const EnvironmentScConnectionService = require('../environment-sc-connection-service');
+
+describe('EnvironmentScConnectionService', () => {
```
jn1119, line 47: I was expecting to see an assert where we check the RStudioUrl after mocking a bunch of calls. Do we not care about what the final RStudio URL looks like?

pnpm-lock.yaml
```diff
@@ -4873,7 +4875,7 @@ packages:
   /@typescript-eslint/typescript-estree/2.27.0:
     dependencies:
       debug: 4.1.1
-      eslint-visitor-keys: 1.3.0
+      eslint-visitor-keys: 1.1.0
```
jn1119, line 4878: Why was this downgrade required?

pnpm-lock.yaml
```diff
@@ -19331,7 +19332,7 @@ packages:
       integrity: sha512-i/6DQjL8Xf3be4K/E6Wgpekn5Qasl1usyw++dAA35Ue5orEn65VIxOA+YvNNl9HV3qv70T7CNwjODHZrLwvd1Q==
   /tsutils/3.17.1:
     dependencies:
-      tslib: 1.13.0
+      tslib: 1.11.1
```
jn1119, line 19335: Why was this downgraded?

pnpm-lock.yaml
SanketD92: The lock file changes automatically once you run ./scripts/install.sh, and since there were no changes made to the eslint config manually I confirmed linting still works as expected by running 
```
pnpm recursive run format
```

pnpm-lock.yaml
SanketD92: No changes were made manually, this was automatic (same as above)

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-connection-service.test.js
SanketD92: I was considering that to be tested manually with an actual certified domain name, but adding a unit test to make sure the URL structure stays as expected makes sense. I'll add a test for that.

jn1119 (approved):

PR #156, opened 2020-09-15
Added accessType property at study level and also added route for upd…

jn1119:
Issue #, if available: GALI-456

Description of changes:
* Added accessType enum in study schema for update and create calls
* Added route for Study PUT for updates
* Added check to disallow ReadWrite accessType on Open Data Studies

Note: These are just backend changes- it is expected that UI will set accessType property when we integrate. Additionally, for backwards compatibility missing accessType will be assumed to be 'ReadOnly'

Checklist: 

[✓] Have you successfully deployed to an AWS account with your changes? 
[✓] Have you linted your code locally prior to submission?
[✓] Have you written new tests for your core changes, as applicable?
[✓] Have you successfully ran unit tests and manual tests with your changes locally?
[ ] If you had to run manual tests, have you considered automating those tests by adding them to [end-to-end tests](../main/end-to-end-tests/README.md)? Didn't run manual tests since the changes are made to Study API which isn't added yet. Additionally, these are backend changes and not being called from UI yet, will tackle this bit when we do the UI integration.

### UI testing
* Verified that create study works
* Verified that all the studies are being displayed correctly

### Testing done via Postman
* CREATE Study call with valid and invalid accessType
* CREATE Study call without accessType
* CREATE Study call on Open Data Study with ReadWrite accessType
* Above pattern of calls with PUT calls
* Verified GET Study call on Studies with and without accessType

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
@@ -88,6 +88,9 @@ class StudyService extends Service {
     // For now, we assume that 'createdBy' and 'updatedBy' are always users and not groups
     const by = _.get(requestContext, 'principalIdentifier'); // principalIdentifier shape is { username, ns: user.ns }
 
+    // validate if study can be read/write
+    this.validateStudyType(rawData.accessType, rawData.category);
+
     // The open data studies do not need to be associated to any project
     // for everything else make sure projectId is specified
     if (rawData.category !== 'Open Data') {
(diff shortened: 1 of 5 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
```diff
@@ -388,6 +401,13 @@ class StudyService extends Service {
     // return auditWriterService.write(requestContext, auditEvent);
     return auditWriterService.writeAndForget(requestContext, auditEvent);
   }
+
+  // ensure that study accessType is read/write for Open Data category
+  validateStudyType(accessType, studyCategory) {
```
SanketD92, line 406: Probably could have a better name, since it's only checking if accessType is ReadWrite. Maybe something like validateReadWrite() ? 

If you're planning to reuse this for other accessType checks in the future, I'll ignore this.

SanketD92 (changes requested):
Method name change request. Everything else LGTM

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
jn1119: Yeah, I just created a validation method where we can dump every thing. The validation turned out to be pretty straight forward for now- planning on leaving it as it is unless you have strong objections.

SanketD92 (dismissed):

SanketD92 (approved):
Looks good.

PR #157, opened 2020-09-16
Bugfix: Exceptions going unseen

SanketD92:
Issue #, if available:
Unexpected exceptions were being ignored

Description of changes:
Unknown/unexpected exceptions should be thrown.

Checklist: 


addons/addon-base-rest-api/packages/services/lib/authentication-providers/built-in-providers/cogito-user-pool/provisioner-service.js
```diff
@@ -424,6 +424,9 @@ class ProvisionerService extends Service {
       if (err.code === 'InvalidParameterException' && err.message.indexOf('already exists') >= 0) {
         // The domain already exists so nothing to do. Just log and move on.
         this.log.info(`The Cognito User Pool Domain with Prefix "${userPoolDomain}" already exists. Nothing to do.`);
+      } else {
+        // Throw other error messages, so it doesn't silently fail
```
dpandya, line 428: Very minor but you are not throwing "error messages", you are re-throwing the caught "error" that was already thrown before. Suggest to change the comment to say something like follows

```
// Re-throw any other error, so it doesn't fail silently
```

dpandya (approved):

PR #163, opened 2020-09-24
Bugfix: metaconnection check for rstudio

SanketD92:
Issue #, if available:
Not all workspace types have metaconnection output

Description of changes:
Divided metaconnection outputValue retrieval for creating/deleting CNAME DNS record for RStudio.

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/plugins/env-provisioning-plugin.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/plugins/env-provisioning-plugin.js
@@ -118,11 +118,15 @@ async function updateEnvOnProvisioningSuccess({
   const existingEnvRecord = await environmentScService.mustFind(requestContext, { id: envId, fields: ['rev'] });
 
   // Create DNS record for RStudio workspaces
-  const connectionType = _.find(outputs, o => o.OutputKey === 'MetaConnection1Type').OutputValue;
-  if (connectionType.toLowerCase() === 'rstudio') {
-    const dnsName = _.find(outputs, o => o.OutputKey === 'Ec2WorkspaceDnsName').OutputValue;
-    const environmentDnsService = await container.find('environmentDnsService');
-    environmentDnsService.createRecord('rstudio', envId, dnsName);
+  const connectionType = _.find(outputs, o => o.OutputKey === 'MetaConnection1Type');
+  let connectionTypeValue;
+  if (connectionType) {
+    connectionTypeValue = connectionType.OutputValue;
+    if (connectionTypeValue.toLowerCase() === 'rstudio') {
+      const dnsName = _.find(outputs, o => o.OutputKey === 'Ec2WorkspaceDnsName').OutputValue;
+      const environmentDnsService = await container.find('environmentDnsService');
+      environmentDnsService.createRecord('rstudio', envId, dnsName);
+    }
   }
 
   const environment = {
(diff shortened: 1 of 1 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/plugins/env-provisioning-plugin.js
```diff
@@ -118,11 +118,15 @@ async function updateEnvOnProvisioningSuccess({
   const existingEnvRecord = await environmentScService.mustFind(requestContext, { id: envId, fields: ['rev'] });
 
   // Create DNS record for RStudio workspaces
-  const connectionType = _.find(outputs, o => o.OutputKey === 'MetaConnection1Type').OutputValue;
-  if (connectionType.toLowerCase() === 'rstudio') {
-    const dnsName = _.find(outputs, o => o.OutputKey === 'Ec2WorkspaceDnsName').OutputValue;
-    const environmentDnsService = await container.find('environmentDnsService');
-    environmentDnsService.createRecord('rstudio', envId, dnsName);
+  const connectionType = _.find(outputs, o => o.OutputKey === 'MetaConnection1Type');
```
Bingjiling, line 121: Looks like these two blocks are exactly the same except for the last line, maybe we can extract it to a function and pass in 'delete' or 'create' as a variable?

SanketD92: I didn't choose to do that since all the other methods in the plugin are linked to workflow steps. Creating a "private" method in this plugin seemed like an anti-pattern.

SanketD92: Also, I'm not sure if we'll need special logic in updateEnvOnTerminationSuccess() since RStudio will soon have auto-stop enabled, and the changed DNS may not be found in the outputs anymore.

Bingjiling: Got it. Yea, that makes sense. Thanks for explaining this 👍

Bingjiling (approved):
LGTM!

addons/addon-base-raas/packages/base-raas-services/lib/plugins/env-provisioning-plugin.js
jn1119: I think a private method can be created in order to avoid duplicate code and it shouldn't be treated as an anti-pattern even though other public methods are linked to workflow steps. The decision to not move the code for determining `dnsName` due to `updateEnvOnTerminationSuccess` changing later is entirely up to you though.

addons/addon-base-raas/packages/base-raas-services/lib/plugins/env-provisioning-plugin.js
```diff
@@ -212,11 +216,15 @@ async function updateEnvOnTerminationSuccess({ requestContext, container, status
...
   // Delete DNS record for RStudio workspaces
-  const connectionType = _.find(updatedEnvironment.outputs, x => x.OutputKey === 'MetaConnection1Type').OutputValue;
-  if (connectionType.toLowerCase() === 'rstudio') {
-    const dnsName = _.find(updatedEnvironment.outputs, x => x.OutputKey === 'Ec2WorkspaceDnsName').OutputValue;
-    const environmentDnsService = await container.find('environmentDnsService');
-    environmentDnsService.deleteRecord('rstudio', envId, dnsName);
+  const connectionType = _.find(updatedEnvironment.outputs, o => o.OutputKey === 'MetaConnection1Type');
+  let connectionTypeValue;
+  if (connectionType) {
+    connectionTypeValue = connectionType.OutputValue;
```
jn1119, line 222: Can we add a couple of unit tests for this branch for both create and delete?

SanketD92: Per our conversation, the underlying service would see some changes during the auto-stop feature addition. Implementing unit tests then.

PR #172, opened 2020-10-06
feat: autostop SM and rstudio

SanketD92:
Issue #, if available:
GALI-476
GALI-477

Description of changes:

Added auto-stop scripts for Sagemaker and RStudio
Added security enhancement for RStudio (environments do not need to be open to internet anymore)

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/sagemaker-notebook-instance.cfn.yml
+++ b/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/sagemaker-notebook-instance.cfn.yml
@@ -30,6 +30,9 @@ Parameters:
   EncryptionKeyArn:
     Type: String
     Description: The ARN of the KMS encryption Key used to encrypt data in the notebook
+  AutoStopIdleTimeInMinutes:
+    Type: Number
+    Description: Number of idle minutes for auto stop to shutdown the instance (0 to disable auto-stop)
 
 Conditions:
   IamPolicyEmpty: !Equals [!Ref IamPolicyDocument, '{}']
@@ -94,6 +97,25 @@ Resources:
                 - logs:CreateLogGroup
               Resource:
                 - !Sub arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/sagemaker/*
+
+  # This policy is attached to the role after the instance is created
+  # so that the instance can be referenced in the resource section
+  NotebookStopPolicy:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: NotebookStop
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Effect: Allow
+            Action:
+              - sagemaker:DescribeNotebookInstance
+              - sagemaker:StopNotebookInstance
+            Resource:
+              - !Ref BasicNotebookInstance
+      Roles:
+        - !Ref IAMRole
+
   # TODO: Consider also passing DefaultCodeRepository to allow persisting notebook data
   BasicNotebookInstance:
     Type: 'AWS::SageMaker::NotebookInstance'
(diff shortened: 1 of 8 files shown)


addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/sagemaker-notebook-instance.cfn.yml
```diff
@@ -94,6 +97,25 @@ Resources:
                 - logs:CreateLogGroup
               Resource:
                 - !Sub arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/sagemaker/*
+
+  # This policy is attached to the role after the instance is created
```
Bingjiling, line 101: Nice call! Can we rename 'IAMRole' to something more specific?

addons/addon-base-raas/packages/base-raas-services/lib/plugins/env-provisioning-plugin.js
```diff
@@ -253,6 +244,39 @@ async function updateEnvOnTerminationSuccess({ requestContext, container, status
...
+// SSM public kay parameter created during environment's provisioning
+async function rstudioCleanup(requestContext, updatedEnvironment, container) {
+  const connectionType = _.find(updatedEnvironment.outputs, o => o.OutputKey === 'MetaConnection1Type');
+  let connectionTypeValue;
+  if (connectionType) {
+    connectionTypeValue = connectionType.OutputValue;
+    if (connectionTypeValue.toLowerCase() === 'rstudio') {
+      const dnsName = _.find(updatedEnvironment.outputs, x => x.OutputKey === 'Ec2WorkspaceDnsName').OutputValue;
+      const instanceId = _.find(updatedEnvironment.outputs, x => x.OutputKey === 'Ec2WorkspaceInstanceId').OutputValue;
+      const environmentDnsService = await container.find('environmentDnsService');
```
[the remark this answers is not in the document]
SanketD92: Can't do that, I tried it. `container.find()` is different than the way we use `this.dependency(['service1','service2'])`

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/sagemaker-notebook-instance.cfn.yml
SanketD92: The RoleName is assigned according to the namespace, which makes it deployment-specific appended by the workspace type. (See line 51)

Bingjiling: Yes! I was talking about the name of the resource in the template itself? '- !Ref IAMRole' seems generic and I spent sometime trying to figure out how this policy gets attached to the SageMaker role after the instance is created. If it says something like '- !Ref SageMakerIAMRole', it would be self-explanatory and easier to grasp. Just some random thoughts, it's very minor.

main/solution/machine-images/config/infra/files/rstudio/set-password
```diff
@@ -1,7 +1,12 @@
 #!/usr/bin/env bash
 
-instance_id=$(curl -s "http://169.254.169.254/1.0/meta-data/instance-id")
+instance_id=$(curl -s "http://169.254.169.254/latest/meta-data/instance-id")
```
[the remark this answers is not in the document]
SanketD92: This command is run by the EC2 instance on it's own bash terminal and will never be available to the user directly. As to whether it poses a the security threat, that's a negative. 
When we try it out on the browser, the instance data we see is probably from EC2 instances that have allowed wider access to themselves. 

I've confirmed the security group assigned to the RStudio EC2 instances is not as wide, and therefore will not be available anywhere but from within the EC2 instance backing it.

Bingjiling (approved):
LGTM!

PR #177, opened 2020-10-12
fix (End to End test): Creating workspace with correct project

nguyen102:
Description of changes:
There's an error in the end to end tests when is is testing creating new workspaces. During creation of new workspace, it fails to select the correct project because the test is looking at the wrong CSS element class. Screenshot below. All elements in the drop down menu will have the class `item`, but not all elements will have the class `selected`

project-id

Checklist: 


SanketD92 (approved):
LGTM

PR #178, opened 2020-10-13
Apply version name to products out of the box

SanketD92:
Issue #, if available:
GALI-481

Description of changes:
Currently when an existing SC product template is updated and deployed to SWB, the auto-create code does not assign a name to the new provisioning artifact. This displays the new product version on SWB UI as "<ProductName> undefined".

To avoid this, the PR achieves the following:

1. Change the default "autoCreateVersion" variable to something simpler like "v1"
2. Add "Name" field in the params used in createProductArtifact().
3. Before attempting to create a new product artifact, call servicecatalog.ListProvisioningArtifacts() to retrieve latest version of product already in SC.
4. Assign the next logical version name to the "Name" field

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
+++ b/addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
@@ -320,7 +320,7 @@ class CreateServiceCatalogPortfolio extends Service {
     if (latestVersionInSC) {
       const parsedOutput = latestVersionInSC.match(pattern);
       if (parsedOutput && parsedOutput.length > 0) {
-        returnVal = `v${parsedOutput[1]}`;
+        returnVal = `v${parseInt(parsedOutput[1], 10) + 1}`;
       }
     }
     return returnVal;
--- a/addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
+++ b/addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
@@ -63,7 +63,7 @@ const productsToCreate = [
   },
 ];
 
-const autoCreateVersion = 'V1.0.0';
+const autoCreateVersion = 'v1';
 const autoCreateDesc = 'Auto-created during post deployment';
 const deploymentItemId = 'default-SC-portfolio-1';
 
@@ -209,6 +209,7 @@ class CreateServiceCatalogPortfolio extends Service {
     const scAvailableProducts = _.map(envTypesAvailable, obj => ({
       productName: obj.product.name,
       provisioningArtifactId: obj.provisioningArtifact.id,
+      provisioningArtifactName: obj.provisioningArtifact.name,
     }));
     const scAvailableProductNames = _.map(scAvailableProducts, p => p.productName);
 
@@ -239,9 +240,12 @@ class CreateServiceCatalogPortfolio extends Service {
               this.log.info(
                 `A previously used CfN file was updated with new CfN product template data for ${productName}`,
               );
+              // Get all provisioning artifacts for this product
+              const latestVersionInSC = ScProduct.provisioningArtifactName;
               const newProvisioningId = await this.createProductArtifact(
                 productFound.productId,
                 productToCreate.filename,
+                latestVersionInSC,
               );
 
               productDetails.provisioningArtifactId = newProvisioningId;
@@ -282,11 +286,12 @@ class CreateServiceCatalogPortfolio extends Service {
     return portfolioToUpdate;
   }
 
-  async createProductArtifact(productId, productFileName) {
+  async createProductArtifact(productId, productFileName, latestVersionInSC) {
     const aws = await this.service('aws');
     const servicecatalog = new aws.sdk.ServiceCatalog({ apiVersion: '2015-12-10' });
     const s3BucketName = this.settings.get(settingKeys.deploymentBucketName);
     const productToCreate = productsToCreate.find(p => p.filename === productFileName);
+    const nextProvArtifact = await this.getNextArtifactVersion(latestVersionInSC);
     const params = {
       Parameters: {
         Info: {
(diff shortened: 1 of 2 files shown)


addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
```diff
@@ -295,13 +300,55 @@ class CreateServiceCatalogPortfolio extends Service {
...
+    // Only finds version strings that match patterns:
+    // "V<n>", "V<n>.0", "V<n>.0.0", "v<n>", "v<n>.0", "v<n>.0.0"
+    const pattern = /^(v|V)(\d+\.)?(\d+\.)?(\*|\d+)$/;
+    let latestVersionNo;
+    if (latestVersionInSC) {
+      const parsedOutput = latestVersionInSC.match(pattern);
+      this.log.info(`${parsedOutput}`);
+      if (parsedOutput && parsedOutput.length > 1) {
+        const longestMatch = parsedOutput.sort(function(a, b) {
+          return b.length - a.length;
```
Bingjiling, line 324: A lot of efforts are made to accommodate different type of version strings "V<n>", "V<n>.0", "V<n>.0.0", "v<n>", "v<n>.0", "v<n>.0.0". Since these products are created and updated by the system, can we define the version number to only follow one convention? Maybe v1, v2, v3? This way we can avoid the complication accommodating different dots and digits.

SanketD92: We initially decided to have the initial product version to be "V1.0.0", but realized the minor and patch versions don't really make sense in template versions and furthermore difficult to determine if upcoming changes are minor or patches (thereby rendering them useless).

Now we have changed `autoCreateVersion` (assigned to the first version of the product) to "v1". The regex pattern is intentionally kept to match both the previous version string format (eg. V1.0.0) as well as the new ones (eg. v1). In the long run, new product versions would stick to the "v<n>" format.

addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
```diff
@@ -295,13 +300,55 @@ class CreateServiceCatalogPortfolio extends Service {
...
+    // For product artifacts in Service Catalog, versions and names are the same thing
+    // If no previous version found, assign 'v2' (this will help in next update cycle)
+    let returnVal = 'v2';
+    // Only finds version strings that match patterns:
+    // "V<n>", "V<n>.0", "V<n>.0.0", "v<n>", "v<n>.0", "v<n>.0.0"
+    const pattern = /^(v|V)(\d+\.)?(\d+\.)?(\*|\d+)$/;
+    let latestVersionNo;
+    if (latestVersionInSC) {
+      const parsedOutput = latestVersionInSC.match(pattern);
+      this.log.info(`${parsedOutput}`);
```
rsmayda, line 321: Do we want to log this? doesn't seem to have context around it

addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
```diff
@@ -295,13 +300,55 @@ class CreateServiceCatalogPortfolio extends Service {
...
+      if (parsedOutput && parsedOutput.length > 1) {
+        const longestMatch = parsedOutput.sort(function(a, b) {
+          return b.length - a.length;
+        })[0];
+        latestVersionNo = longestMatch.substring(1);
+        // Increment major version of the string and assign to the new artifact
+        const versionToCompute = [latestVersionNo]
```
rsmayda, line 328: I may be missing something but if we know the format is V#.#.# couldn't we just do:
`input.substring(1).split('.')[0]` to get the major string?

Or just use the regex grouping?

addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
SanketD92: For testing only, needs to be taken out. Thanks for pointing it out.

addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
SanketD92: I think we'll need to cover all the mentioned format, since it could be one of these:
1. V#.#.# (The initially decided default version)
2. v# (The version format from here on)
3. Any other format/string that users could have used while manually adding an artifact

SanketD92: You're right. That works too.

addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
carvantes: You may want to add a summary of this as a comment in code. Just explaining that it supports "legacy" versions  like v.1.2.3 but that new versions are simplified to look like v2

addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
```diff
@@ -295,13 +300,35 @@ class CreateServiceCatalogPortfolio extends Service {
...
+    const pattern = /^(v|V)(\d+\.)?(\d+\.)?(\*|\d+)$/;
+    let latestVersionNo;
+    if (latestVersionInSC) {
+      const parsedOutput = latestVersionInSC.match(pattern);
+      if (parsedOutput && parsedOutput.length > 1) {
+        const longestMatch = parsedOutput.sort(function(a, b) {
+          return b.length - a.length;
+        })[0];
+        latestVersionNo = longestMatch.substring(1);
+        returnVal = `v${parseInt(latestVersionNo.split('.')[0], 10) + 1}`;
```
carvantes, line 326: Using regex AND string methods (i.e. substring, split) together is often an anti pattern. It leads to fragmented/duplicated string parsing logic. 

A regex with well placed capture groups can find the version number you are looking for. 

Try something like
```regex
^(?:v|V)(\d+)?(?:\.\d+)?(?:\.(?:\*|\d+))?$
```
https://regexr.com/5e53g

addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
```diff
@@ -295,13 +300,35 @@ class CreateServiceCatalogPortfolio extends Service {
...
     return data.ProvisioningArtifactDetail.Id;
   }
 
+  async getNextArtifactVersion(latestVersionInSC) {
+    // For product artifacts in Service Catalog, versions and names are the same thing
+    // If no previous version found, assign 'v2' (this will help in next update cycle)
+    let returnVal = 'v2';
+    // Only finds version strings that match patterns:
+    // "V<n>", "V<n>.0", "V<n>.0.0", "v<n>", "v<n>.0", "v<n>.0.0"
+    const pattern = /^(v|V)(\d+\.)?(\d+\.)?(\*|\d+)$/;
```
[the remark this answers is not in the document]
SanketD92: No particular need for the `*` symbol, need to remove it. Thanks for catching that.

addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
```diff
@@ -295,13 +300,35 @@ class CreateServiceCatalogPortfolio extends Service {
...
+    let latestVersionNo;
+    if (latestVersionInSC) {
+      const parsedOutput = latestVersionInSC.match(pattern);
+      if (parsedOutput && parsedOutput.length > 1) {
+        const longestMatch = parsedOutput.sort(function(a, b) {
```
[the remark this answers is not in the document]
SanketD92: That's good to know, I'll update the code accordingly. Thanks!

addons/addon-base-raas/packages/base-raas-post-deployment/lib/steps/create-service-catalog-portfolio.js
SanketD92: Made the changes per our conversation. Thanks for this suggestion 👍

carvantes (approved):

PR #179, opened 2020-10-14
chore: Env delete logs

Ludo171:
Description of changes:

The script `scripts/environment-delete.sh` can take incredible time to end. It is mostly due to the operations of cleaning the S3 buckets before trying to delete them (in order to avoid cloudformation errors). Last time it took 1 hour to run completely (because the version of SWB was up for several months and accumulated a lot of logs in its buckets). Several times I thought there might be an error with the script freezing my terminal whereas the script actually worked fine, but the operations take time and no progress was displayed. Also during this scenario I interrupted the script after it removed the backend stack. From then on it would crash at the beginning when trying to remove non-existing stacks.

The following commits aim at making things clear :
- commit 94dd532 : display the progression of the S3 operations.
- commit 7ec953d  : avoid script interruptions on serverless operations that throw errors in case a stack does not exist.

Checklist: 

* [x ] Have you successfully deployed to an AWS account with your changes? 
* [x ] Have you linted your code locally prior to submission?
* [x ] Have you successfully ran unit tests and manual tests with your changes locally?

Result : for each S3, terminal display progress in one single line updated alongside the operations. 
For instance one line could be updated with the following content displayed successively :
` <S3_bucket> <Operation_1> ...  $progress/$total ...` 
` <S3_bucket> <Operation_2> ...  $progress/$total ...` 
` <S3_bucket> <Operation_2> ...  Done !.` 
Screenshot from 2020-10-13 11-45-06

### DIFF
--- a/scripts/environment-delete.sh
+++ b/scripts/environment-delete.sh
@@ -138,13 +138,17 @@ function emptyS3Bucket() {
     local bucket=$1
     local region=$2
     local delete_option=$3
-    printf "\n- Emptying bucket $bucket ... "
+
+    blank="                                                                                                                        "
+    message="\\r$blank\\r- Emptying bucket $bucket ... "
+    printf "\n$message"
 
     # Remove Versions for all objects
     versions=$(aws s3api list-object-versions --bucket $bucket | jq '.Versions')
     let count=$(echo $versions | jq 'length')-1
     if [ $count -gt -1 ]; then
         for i in $(seq 0 $count); do
+            printf "$message Removing objects versions : $i/$count"
             key=$(echo $versions | jq .[$i].Key | sed -e 's/\"//g')
             versionId=$(echo $versions | jq .[$i].VersionId | sed -e 's/\"//g')
             cmd=$(aws s3api delete-object --bucket $bucket --key $key --version-id $versionId)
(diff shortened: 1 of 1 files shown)


scripts/environment-delete.sh
```diff
@@ -138,13 +142,17 @@ function emptyS3Bucket() {
     local bucket=$1
     local region=$2
     local delete_option=$3
-    printf "\n- Emptying bucket $bucket ... "
+
+    blank="                                                                                                                        "
```
rsmayda, line 146: What is this var and why do we need it?

Ludo171: To make display the operation progress in a smart way I am using the escape character `\r` that allow me to print a message on the same line as the previous message.
In order to update the progress displayed in one single line, I am printing as follow :  
`printf "\\r$blank\\r- Emptying bucket $bucket ... $i/$count" `

Precisely this instruction prints `$blanck` from the start of the current line of the console (it clears the line), then it displays the updated progress message.

rsmayda: perfect, thanks for explaining! yeah in the picture I just noticed at the bottom of the screenshot the progression

rsmayda (approved):

Issue #185, opened 2020-10-14
SM & RStudio autoStop does not work if user does not click on button 'close the session' before closing its web browser


Ludo171:
#### Description :
Related to the new feature proposed by PR #171 (commit 1e8b067f96c463a4b3b27f8b625d2f2f1591d452). We expect that this new feature would turn RStudio workspaces to sleep after 1hour of inactivity, and Sagemaker Notebooks after a custom idletime specified in the workspace configuration. I used the Sagemaker Notebook and RStudio configurations coming out-of-the-box from 81c60ef537cb45b55dc701b7d0a0dfbfbb26c81c.

The behavior works as expected if I do not try to connect on the workspaces at all. In contrast, if I connect on the workspaces and execute a few instructions before closing the window, I realized that they were still up and running 5 hours afterwards.

#### Reproduce the bug :
- Create an RStudio workspace
- Connect onto it and execute a few instructions ('variable <- 4 +4') suffices.
- Close the window, and your web browser.
- Connect on Service Workbench more than 60min afterwards
--> Assess that the workspace is still up and running
--> On the AWS console, assess that the EC2 instance is still up and running
--> The same behavior goes for Sagemaker notebooks if you specify 60 min of AutoStopIdleTime in the configurations.

On the opposite, if I do not perform step 2 (connect and execute instructions), then the workspace automatically stops after 1 hour.

SanketD92:
1. What type of instructions did you run on the workspaces? Would they be expected to run in the background on the instances?
2. Could you confirm if you "Quit" the RStudio session before exiting the portal?

FYI - As for the Sagemaker auto-stop feature, we're using this [script](https://raw.githubusercontent.com/aws-samples/amazon-sagemaker-notebook-instance-lifecycle-config-samples/master/scripts/auto-stop-idle/autostop.py) to check for its idle state from the last activity.

Ludo171:
Hi @SanketD92 

1) Here are the very basic instructions that I ran on RStudio and Sagemaker :
![CaptureRstudio](https://user-images.githubusercontent.com/33128425/96127529-b8319b80-0ef4-11eb-9143-a15a9ae85d25.PNG)
![CaptureSagemakr](https://user-images.githubusercontent.com/33128425/96127554-b9fb5f00-0ef4-11eb-803f-67ea869f3977.PNG)  

[...]

PR #188, opened 2020-10-16
fix: changing rstudio check-idle logic

SanketD92:
Issue #, if available:
https://github.com/awslabs/service-workbench-on-aws/issues/185

Description of changes:
Changing check-idle logic for RStudio.

Checklist: 

### DIFF
--- a/main/solution/machine-images/config/infra/files/rstudio/check-idle
+++ b/main/solution/machine-images/config/infra/files/rstudio/check-idle
@@ -1,42 +1,53 @@
 #!/usr/bin/env python
 
-
 import os
-import time
+import datetime as dt
+import dateutil.parser as parser
 
 
 LAST_ACTIVE_TIME_FILENAME = '/tmp/last-active-rsession-time'
+ACCESS_LOG_FILENAME = '/var/log/nginx/access.log'
 
-MAX_IDLE_SECONDS = 3600
-
-
-def get_uptime():
-    return float(open('/proc/uptime').read().split()[0])
+# To disable auto-stop for RStudio, enter 0 as the value below for MAX_IDLE_MINUTES
+MAX_IDLE_MINUTES = 60
 
 
-def is_rsession(pid):
-    try:
-        cmdline = open(os.path.join('/proc', pid, 'cmdline'), 'rb').read()
-        return 'rsession' in cmdline
-    except IOError:
-        return False
+def get_activity_count():
+    non_ping_count = 0
+    with open(ACCESS_LOG_FILENAME) as access_log:
+        for line in access_log:
+            # nginx sometimes pings the RStudio instance
+            # which does not indicate user activity
+            if "POST /events/get_events" not in line:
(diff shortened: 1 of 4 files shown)


main/solution/machine-images/config/infra/files/rstudio/check-idle
```diff
@@ -1,42 +1,53 @@
...
+    for line in fh:
+        pass
+    last_active_time = parser.parse(line)
 
-if num_sessions or not os.path.exists(LAST_ACTIVE_TIME_FILENAME):
-    open(LAST_ACTIVE_TIME_FILENAME, 'w').write(str(current_time))
-    print('{0} RStudio sessions active at {1}'.format(num_sessions, current_time))
+# Compare with old_activity_count
+# If higher than old count, update file with time and count
+if latest_activity_count > last_activity_count:
```
Bingjiling, line 45: Why is the condition here higher than old count? Not latest_activity_count > 0?

main/solution/machine-images/config/infra/files/rstudio/check-idle
```diff
@@ -1,42 +1,53 @@
...
+        str(latest_activity_count) + '\n' + str(current_time))
 else:
-    last_active_time = float(open(LAST_ACTIVE_TIME_FILENAME).read())
-    idle_time = current_time - last_active_time
-    uptime = get_uptime()
-    print('RStudio has been idle for {0} seconds and system uptime is {1}'.format(idle_time, uptime))
-    if idle_time >= MAX_IDLE_SECONDS and uptime >= MAX_IDLE_SECONDS:
+    # If time difference between file time and now exceeds MAX_IDLE_MINUTES, shutdown
+    idle_minutes = (current_time - last_active_time).total_seconds()/60
+    if MAX_IDLE_MINUTES != 0 and idle_minutes >= MAX_IDLE_MINUTES:
```
Bingjiling, line 51: Minor - Is there a way to quit the file execution early on? If MAX_IDLE_MINUTES == 0, the file don't really need to be ran.

main/solution/machine-images/config/infra/files/rstudio/check-idle
SanketD92: If there is any activity on the RStudio platform, the count would increase. Even if the instance is rebooted this count, and the log file from which the count is derived, do not change; the count and logs are simply updated and appended respectively.

main/solution/machine-images/config/infra/files/rstudio/check-idle
SanketD92: Good point. Making that change.

main/solution/machine-images/config/infra/files/rstudio/check-idle
Bingjiling: Ah! Got it. Thanks for the explanation. :)

Bingjiling (approved):
LGTM!

Issue #187, opened 2020-10-15
Enable admins to change CIDR block of a workspace


Ludo171:
#### [Feature Suggestion]

Yesterday we felt in a situation that would have been solved by this suggested feature. 
Some user was working on a workspace. After the workspace had been stopped, he left it alone for several days. One of our admins wanted to check the content of the workspace before terminating it, but the CIDR block only pointed to the original creator's IP. The admin had no way of accessing the workspace.

It would benefit multiple use-cases :
- Handle IP change for one user (changing internet provider, moving into a new house)
- Share access with additional users, by updating the IP.

As a work-around to those missing use-cases, we specify the CIDR block of our VPN. But for sharing with external users, it is enough.

(We would place this feature as a medium-priority feature)

nguyen102:
Thanks for the suggestion Ludo. I've created a back log ticket for this. As we prioritize items, we will keep this request in mind.

nguyen102:
I added a PR to fix this bug.

nguyen102:
Changes have been merged

nguyen102:
Sorry the PR linked here is incorrect. Please ignore that.

carvantes:
This was fixed as part of https://github.com/awslabs/service-workbench-on-aws/pull/269

PR #189, opened 2020-10-16
fix: Sagemaker instances respect CIDR blocks that are provided to the instance

nguyen102:
Issue #, if available:
https://github.com/awslabs/service-workbench-on-aws/issues/187

Description of changes:
Sagemaker instances respect CIDR blocks that are provided to the instance. This change adds the CIDR block to Sagemaker Notebook instance's Security Group Ingress rule. 

This will restrict access to that Sagemaker instance to clients from that CIDR.

I tested trying to access the instance while on VPN and with the correct IP address and while off VPN with the wrong address. I was able to access the instance on VPN but not while off of VPN. This worked as expected.
Checklist: 


Bingjiling (approved):
LGTM! Nice work!

PR #191, opened 2020-10-20
fix: Query string parameters were getting duplicated in the url

dpandya:
Query string parameters were getting duplicated in the url upon retries. This was causing intermittent failures on the Workspace Types page.

Issue #, if available:

Description of changes:
- Delete merged.params after they are added to the url to stop them from getting duplicated upon recursive retry calls
- This fixes an issue where the `Workspace Types` screen intermittently shows `Invalid version specified for filter. Valid values for version are *,latest` error under `AWS Service Catalog Products` section. That error happens when the initial call times out (because fetching Service Catalog Products is relatively slow operation) and the call is retried. 
image

Checklist: 

### DIFF
--- a/addons/addon-base-ui/packages/base-ui/src/helpers/api.js
+++ b/addons/addon-base-ui/packages/base-ui/src/helpers/api.js
@@ -74,6 +74,7 @@ function fetchJson(url, options = {}, retryCount = 0) {
 
   if (merged.method === 'GET') delete merged.body; // otherwise fetch will throw an error
 
+  let retryOptions = options;
   if (merged.params) {
     // if query string parameters are specified then add them to the URL
     // The merged.params here is just a plain JavaScript object with key and value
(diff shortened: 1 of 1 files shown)


addons/addon-base-ui/packages/base-ui/src/helpers/api.js
```diff
@@ -89,6 +89,16 @@ function fetchJson(url, options = {}, retryCount = 0) {
...
+
+    // Delete merged.params after they are added to the url as query string params
+    // This is required otherwise, if the call fails for some reason (e.g., time out) the same query string params
+    // will be added once again to the URL causing duplicate params being passed in.
+    // For example, if the merge.params = { param1: 'value1', param2: 'value2' }
+    // The url will become something like `https://some-host/some-path?param1=value1&param2=value2`
+    // If we do not delete "merged.params" here and if the call is retried (with a recursive call to "fetchJson") due
+    // to timeout or any other issue, the url will then become
+    // `https://some-host/some-path?param1=value1&param2=value2?param1=value1&param2=value2`
+    delete merged.params;
```
rsmayda, line 101: isn't `merge.params` a local variable? I don't see how changing a local variable will change the outcome of a future invocation. Maybe I missed something or miss understand it?

dpandya: Of course, not sure what I was thinking. It should be `options` and not `merged`, when fixing this locally at one point I had changed the retry call to pass `merged` instead of `options` so the recursive call will receive new options without the `params` but then I switched back to passing the `options` in the retry calls. Anyways, I will fix this by creating new `retryOptions` and passing that in the recursive call in case of retries. (Don't want to just do `delete options.params` as that would end up mutating the object being passed)

rsmayda (approved):
Looks good!

PR #199, opened 2020-10-29
docs:  add specific info to enable cost breakdowns

Ludo171:
**Issue** : Cost breakdowns were always empty in a version of SWB deployed in a test account.

**Description of changes:**
Cost allocation tags need to be activated in order to get actual cost reports.
We figured that this information was not included in the documentation yet. 
Feel free to correct it if you think of more exact formulations.

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

### DIFF
--- a/docs/docs/deployment/pre_deployment/prerequisites.md
+++ b/docs/docs/deployment/pre_deployment/prerequisites.md
@@ -17,4 +17,4 @@ During the installation, we will require access to the billing account (Created 
 ## Cost Explorer
 
 In order to see any actual cost in dashboards and workspaces, the master account must have Cost Explorer set up. 
-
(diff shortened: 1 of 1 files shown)


docs/docs/deployment/pre_deployment/prerequisites.md
```diff
@@ -16,5 +16,4 @@ During the installation, we will require access to the billing account (Created
 
 ## Cost Explorer
 
-In order to see any actual cost in dashboards and workspaces, the master account must have Cost Explorer set up.
```
SanketD92, line 19: This line would be a prerequisite to the one you've added, so it should still exist.

Ludo171:
@SanketD92 I added back the first sentence specifying that Cost Explorer should be set up.

SanketD92 (approved):
LGTM

PR #200, opened 2020-10-29
ci: Only download the env settings file from S3 if not present in the rep…

adrian-grajdeanu:
ci: better handling of the environment settings file
* ci: Only download the env settings file from S3 if not present in the repo
* ci: Upload env settings to S3 during pipeline deploy only if not tracked in repo

Description of changes:
During CICD pipeline execution, the environment settings file used to always be downloaded from S3, overwriting any local copy that is tracked (and might have been updated) in repository. This present change makes it so that the environment settings file is only uploads to S3 (at CICD pipeline deploy time) if not tracked in repository. Furthermore, during pipeline execution, only downloads it from S3 if not found locally.


main/cicd/cicd-pipeline/scripts/upload-env-config-if-not-versioned.sh
```diff
@@ -0,0 +1,29 @@
...
+for environment_name in "$@"; do
+  config_filename="../../config/settings/${environment_name}.yml"
+
+  if [[ -f "${config_filename}" ]]; then
+    git ls-files --error-unmatch "${config_filename}" >/dev/null 2>&1 && (
+      echo "File '${config_filename}' is tracked in version control; not uploaded to S3"
+    ) || (
+      echo "File '${config_filename}' is not tracked in version control"
+      # no need to echo the upload, as the aws command does that anyway
+      # echo "Uploading '${config_filename}' to '${config_s3_path}'"
```
jn1119, line 23: nit: can be removed

main/cicd/cicd-pipeline/scripts/upload-env-config-if-not-versioned.sh
```diff
@@ -0,0 +1,29 @@
...
+      aws --profile "${aws_profile}" s3 cp --sse aws:kms "${config_filename}" "${config_s3_path}"
+    )
+  else
+    echo "File not found: '${config_filename}'; condition ignored!"
```
jn1119, line 27: nit: ~~condition ignored~~ nothing to upload

jn1119 (approved):
LGTM

main/cicd/cicd-pipeline/scripts/upload-env-config-if-not-versioned.sh
adrian-grajdeanu: Line 20 cannot be removed.
Line 21 Could, but it is useful to be informative
Line 22-23 can be removed, but they don't hurt explaining why no need to announce what we're about to do (upload).

main/cicd/cicd-pipeline/scripts/upload-env-config-if-not-versioned.sh
adrian-grajdeanu: Lines 24-26 got to be in there.
If you nit is about line 27, I agree and will change to "echo \"File not found: '${config_filename}'; ignored!\""
Your proposal 'nothing to upload' is not appropriate in all cases.

jn1119: Yeah, I just meant that we should indicate that that particular file doesn't exist so won't be uploaded. Ignored messaging sounds good.  Comment was meant for just line 27.  Something funky happened when I was adding my comments.

main/cicd/cicd-pipeline/scripts/upload-env-config-if-not-versioned.sh
jn1119: Yeah, comment was meant for just line 23.  Something funky happened when I was adding my comments. Seems like a redundant line to have.

jn1119 (approved):

PR #202, opened 2020-11-05
feat: NFS File Share Allowlist Update on EC2 start and stop

Bingjiling:
Issue #, if available:

Description of changes:
* Update file share creation to save file share ARN to study table 
* Update file share creation to account for situation where DDB data save fails 
* Update EC2 start, stop and RStudio start to call for IP allowlist update 
* Create methods to find associated studies and file arns 
* Create methods to update IP allowlist

Checklist: 


addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -93,6 +93,30 @@ class EnvironmentMountService extends Service {
...
+    const fileShareARNs = studiesList.map(study => study.resources[0].fileShareArn).filter(arn => !_.isUndefined(arn));
+    if (!_.isEmpty(fileShareARNs)) {
+      let ip;
+      // If IP is in ipAllowListAction, use that, if not, find it in existingEnvironment
+      if ('ip' in ipAllowListAction) {
+        ip = ipAllowListAction.ip;
+      } else {
+        ip = existingEnvironment.outputs.filter(output => output.OutputKey === 'Ec2WorkspaceInstanceId')[0].OutputValue;
+      }
+      await storageGatewayService.updateFileShareIPAllowedList(fileShareARNs, ip, ipAllowListAction.action);
```
jn1119, line 116: Didn't see this method anywhere. Was it supposed to be part of the service?

addons/addon-base-raas/packages/base-raas-services/lib/plugins/env-provisioning-plugin.js
```diff
@@ -211,7 +211,7 @@ async function updateEnvOnTerminationSuccess({ requestContext, container, status
     status,
     inWorkflow: 'false',
   };
-  const updatedEnvironment = await environmentScService.update(requestContext, environment);
+  const updatedEnvironment = await environmentScService.update(requestContext, environment, { action: 'REMOVE' });
```
jn1119, line 214: Just for my knowledge- is this supposed to run for just RStudio on stop as well? The filename seemed like this was pretty generic, so can you explain where this fits in. Also wondering why isn't there a rstudio-stop js file where we can add this?

jn1119 (commented):
I am assuming unit tests will come later?

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
Bingjiling: Good catch! Had some merge conflict yesterday since I used an existing branch, the changes for storageGatewayService was override by accident. Just added that part of the change.

addons/addon-base-raas/packages/base-raas-services/lib/plugins/env-provisioning-plugin.js
Bingjiling: This plugin method is only used for environment termination, since the environment termination workflow uses this plugin to handle all actions after termination success, this is where it fits in. rstudio-stop is reusing ec2-stop js file, so no extra update is needed for rsdutio-sop.

jn1119: One clarifying question- Is this redundant piece of code if ec2-stop already has the ip removal code?

Bingjiling: Yes! The environment DB record update can be reused by all the workflow with some small twist in pass in parameters, and use plugin to invoke the method. I'll create a tech debt ticket to centralize the DB record update part. The ip removal code is essentially pass in one extra parameter when calling the environmentScService, so it shouldn't add up to existing tech debt.

Bingjiling: Actually we already have a ticket for it - https://sim.amazon.com/issues/GALI-491

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -93,6 +93,30 @@ class EnvironmentMountService extends Service {
     return this._updateResourcePolicies({ updateAwsPrincipals, workspaceRoleArn, s3Prefixes });
   }
 
+  async updateStudyFileMountIPAllowList(requestContext, existingEnvironment, ipAllowListAction) {
```
jn1119, line 96: Where is this method being tested?

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -93,6 +93,30 @@ class EnvironmentMountService extends Service {
...
+        return { id };
+      }),
+    );
+
+    // If yes, get the file share ARNs and call to update IP allow list
+    const fileShareARNs = studiesList.map(study => study.resources[0].fileShareArn).filter(arn => !_.isUndefined(arn));
+    if (!_.isEmpty(fileShareARNs)) {
+      let ip;
+      // If IP is in ipAllowListAction, use that, if not, find it in existingEnvironment
+      if ('ip' in ipAllowListAction) {
```
jn1119, line 111: Is it deliberate that string 'ip' needs to be in ipAllowListAction

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
Bingjiling: Ah! Missed this one. Adding it now.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
Bingjiling: When it's add new IP to the allow list, the calling method usually knows about the new IP that should be added, so it's passed in here to avoid the trouble of downstream method having the find out the new IP again; and when it's remove old IP, the IP should be the one that's in the old DDB record.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -93,6 +93,30 @@ class EnvironmentMountService extends Service {
...
+        ip = ipAllowListAction.ip;
+      } else {
+        ip = existingEnvironment.outputs.filter(output => output.OutputKey === 'Ec2WorkspaceInstanceId')[0].OutputValue;
```
jn1119, line 114: Can we add a unit test for this branch as well. The key should probably reflect the 'ip' address.

jn1119 (approved):

PR #203, opened 2020-11-06
feat: Study Read/Write and Permission propagation (Goofys)

SanketD92:
Issue #, if available:
GALI-455

Description of changes:
This feature has added a new permission level for studies - Read/Write. Study admins can now choose to add/remove/switch permission levels for users having access to these studies, and the changes will be propagated to the workspaces owned by those users.
The logic takes into consideration what changes have been made on the SWB UI, and leverages the same API call that is used for study-permission update for updating the workspace's IAM role of all the users' owned workspaces.

Note: 
1. Being an admin still does not guarantee study access on your workspaces, it simply means you can add/edit/remove users who have access to that study. In order for the admin themself to have study access permission on their owned workspaces, they need to be in one of the permission levels (R/O or R/W).
2. Studies will not be mounted/unmounted according to the permission level changes. The mounts will remain as they were selected at the time of workspace provisioning.
3. In order to list any study's contents in your workspace, you need to have access to _some_ permission level (R/O or R/W)

**What to expect**:
Let's say a user has Read/Write access to an organization study named "AnotherOrgStudy". The user selects this study and provisions an EC2 Linux workspace with it.
On the workspace, the user should be able to list, read and write file contents inside this study

```
[ec2-user@ip-xx-x-xx-xxx AnotherOrgStudy]$ ls
example.yml   test.txt
[ec2-user@ip-xx-x-xx-xxx AnotherOrgStudy]$ echo "some test line" > test2.txt
[ec2-user@ip-xx-x-xx-xxx AnotherOrgStudy]$ ls
example.yml  test2.txt  test.txt
[ec2-user@ip-xx-x-xx-xxx AnotherOrgStudy]$ cat test2.txt 
some test line
```

Once the study admin changes the user's permission level on this study to R/O, the user won't be able to make any changes on the file contents of the study but will continue to have list and read permissions

```
[ec2-user@ip-xx-x-xx-xxx AnotherOrgStudy]$ ls
example.yml  test2.txt  test.txt
[ec2-user@ip-xx-x-xx-xxx AnotherOrgStudy]$ echo "another test line" > test3.txt
-bash: test3.txt: Permission denied
[ec2-user@ip-xx-x-xx-xxx AnotherOrgStudy]$ cat test2.txt 
some test line
```
If all permissions (R/O and R/W) are taken away from the study, the user won't be able to list/read/write contents of that study.

```
[ec2-user@ip-xx-x-xx-xxx AnotherOrgStudy]$ ls
ls: reading directory .: Permission denied
```

This holds true for any workspace type.

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
+++ b/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
@@ -171,7 +171,7 @@ async function configure(context) {
       }
 
       const result = await studyPermissionService.update(requestContext, studyId, updateRequest);
-      await environmentMountService.applyWorkspacePermissions(studyId, updateRequest);
+      await environmentMountService.applyWorkspacePermissions(studyId, updateRequest); // Nice-to-have: Add number of workspaces updated to result object
       res.status(200).json(result);
     }),
   );
--- a/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
+++ b/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
@@ -171,7 +171,7 @@ async function configure(context) {
       }
 
       const result = await studyPermissionService.update(requestContext, studyId, updateRequest);
-      await environmentMountService.manageWorkspacePermissions(studyId, updateRequest);
+      await environmentMountService.applyWorkspacePermissions(studyId, updateRequest);
       res.status(200).json(result);
     }),
   );
--- a/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
+++ b/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
@@ -162,6 +162,7 @@ async function configure(context) {
       const updateRequest = req.body;
 
       // Validate permissions and usage
+      // For future - Move authZ logic inside service (or new service) & move away from verifyRequestorAccess
       await studyPermissionService.verifyRequestorAccess(requestContext, studyId, req.method);
       const study = await studyService.mustFind(requestContext, studyId);
       if (study.category === 'My Studies') {
@@ -171,12 +172,13 @@ async function configure(context) {
         throw context.boom.forbidden('Permissions cannot be set for studies in the "Open Data" category', true);
       }
 
-      lockService.tryWriteLockAndRun({ id: `${studyId}-workspaces` }, async () => {
-        const result = await studyPermissionService.update(requestContext, studyId, updateRequest);
+      const result = await lockService.tryWriteLockAndRun({ id: `${studyId}-workspaces` }, async () => {
+        const updateOutcome = await studyPermissionService.update(requestContext, studyId, updateRequest);
         await environmentMountService.applyWorkspacePermissions(studyId, updateRequest);
         // Nice-to-have: Add number of workspaces updated to result object
-        res.status(200).json(result);
+        return updateOutcome;
       });
+      res.status(200).json(result);
     }),
   );
 
--- a/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
+++ b/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
@@ -17,7 +17,11 @@ async function configure(context) {
   const router = context.router();
   const wrap = context.wrap;
 
-  const [studyService, studyPermissionService] = await context.service(['studyService', 'studyPermissionService']);
+  const [studyService, studyPermissionService, environmentMountService] = await context.service([
+    'studyService',
+    'studyPermissionService',
+    'environmentMountService',
+  ]);
 
   // ===============================================================
   //  GET / (mounted to /api/studies)
@@ -167,6 +171,7 @@ async function configure(context) {
       }
 
       const result = await studyPermissionService.update(requestContext, studyId, updateRequest);
+      await environmentMountService.updateExistingPermissions(studyId, updateRequest);
       res.status(200).json(result);
     }),
   );
--- a/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
+++ b/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
@@ -171,7 +171,7 @@ async function configure(context) {
       }
 
       const result = await studyPermissionService.update(requestContext, studyId, updateRequest);
-      await environmentMountService.updateExistingPermissions(studyId, updateRequest);
+      await environmentMountService.manageWorkspacePermissions(studyId, updateRequest);
       res.status(200).json(result);
     }),
   );
--- a/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
+++ b/addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
@@ -17,10 +17,11 @@ async function configure(context) {
   const router = context.router();
   const wrap = context.wrap;
 
-  const [studyService, studyPermissionService, environmentMountService] = await context.service([
+  const [studyService, studyPermissionService, environmentMountService, lockService] = await context.service([
     'studyService',
     'studyPermissionService',
     'environmentMountService',
+    'lockService',
   ]);
 
   // ===============================================================
@@ -170,9 +171,12 @@ async function configure(context) {
         throw context.boom.forbidden('Permissions cannot be set for studies in the "Open Data" category', true);
       }
 
-      const result = await studyPermissionService.update(requestContext, studyId, updateRequest);
-      await environmentMountService.applyWorkspacePermissions(studyId, updateRequest); // Nice-to-have: Add number of workspaces updated to result object
-      res.status(200).json(result);
+      lockService.tryWriteLockAndRun({ id: `${studyId}-workspaces` }, async () => {
+        const result = await studyPermissionService.update(requestContext, studyId, updateRequest);
+        await environmentMountService.applyWorkspacePermissions(studyId, updateRequest);
+        // Nice-to-have: Add number of workspaces updated to result object
+        res.status(200).json(result);
+      });
     }),
   );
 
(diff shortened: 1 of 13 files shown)


addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
```diff
@@ -167,6 +171,7 @@ async function configure(context) {
       }
 
       const result = await studyPermissionService.update(requestContext, studyId, updateRequest);
+      await environmentMountService.manageWorkspacePermissions(studyId, updateRequest);
```
hatimkhan, line 174: I think renaming the method to applyWorkspacePermissions() might make more sense.
Regarding the design,  it might be worth having a class diagram that shows the relationship between  studyPermissionService, environmentMountService and environmentScService.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -212,15 +235,406 @@ class EnvironmentMountService extends Service {
...
   }
 
+  /**
+   * Function that calculates which users need should and should not have access permissions for the given studyId
+   * Accordingly calls methods that ensure/remove/update permissions for said users
+   *
+   * @param {String} studyId
+   * @param {Object} updateRequest - permission add/remove/update requests coming from SWB UI
+   */
+  async manageWorkspacePermissions(studyId, updateRequest) {
```
hatimkhan, line 245: I have a few design comments here, specially regarding the handling of exceptions and how you want to async  this.   It might be better to discuss this over chime.

hatimkhan: You might also need to pass the requestContext because the applyWorkspacePermissions() method might be used somewhere else and we need to check that permissions are enforced there too.

addons/addon-user-id-upgrade/packages/user-id-upgrade-post-deployment/lib/data/study-permissions-transforms.js
```diff
@@ -65,6 +65,7 @@ module.exports = [
   addUid,
   updateId,
   listValue('readonlyUsers'),
+  listValue('readwriteUsers'),
```
hatimkhan, line 68: There is no need to make any update to the user-id-upgrade addon.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
SanketD92: Sounds good, thanks.

addons/addon-user-id-upgrade/packages/user-id-upgrade-post-deployment/lib/data/study-permissions-transforms.js
SanketD92: True, since it wasn't available before anyways.

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
SanketD92: Sounds good. I'll work on the class diagram and add it to the Quip doc.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -138,30 +147,44 @@ class EnvironmentMountService extends Service {
...
             Resource: [`arn:aws:s3:::${s3BucketName}/${prefix}*`],
           };
+          // Write Permission
+          let putStatement = {
+            Sid: putSid,
+            Effect: 'Allow',
+            Principal: { AWS: [] },
+            Action: ['s3:PutObject'],
+            Resource: [`arn:aws:s3:::${s3BucketName}/${prefix}*`],
+          };
```
jn1119, line 165: Nit: Probably put a comment here stating why PutObjectAcl is not required. It seems very counter intuitive that the IAM policy requires it but the bucket policy doesn't.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -212,15 +235,406 @@ class EnvironmentMountService extends Service {
...
+              studyPathArn,
+              policyDoc,
+              roleName,
+              studyDataPolicyName,
+            } = await this._getIamUpdateParams(env, studyId);
+
+            const statementSidToUse = this._getStatementSidToUse(user.permissionLevel);
+            policyDoc.Statement = this._getStatementsAfterAddition(policyDoc, studyPathArn, statementSidToUse);
+            policyDoc.Statement = this._ensureListAccess(policyDoc, studyPathArn);
+            await iamService.putRolePolicy(roleName, studyDataPolicyName, JSON.stringify(policyDoc), iamClient);
```
jn1119: For all the role policies which are being updated, do we need to hold a lock in order to avoid race conditions and simultaneous updates to the policy. We do something similar when updating S3 bucket policies.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
SanketD92: Exception handling has been added. User permissions are being checked on studyPermissionService.verifyRequestorAccess() as we discovered :)

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -247,14 +247,46 @@ class EnvironmentMountService extends Service {
...
+        `Cannot update ${total} workspace permissions at a time. Please reduce the number of user permission changes and try again`,
+      );
     }
-    if (permissionChangeUsers.length) {
-      await this.updatePermissions(permissionChangeUsers, studyId, updateRequest);
+
+    const errors = [];
+
+    if (!_.isEmpty(allowedUsers)) {
+      const errorsforAdd = await this.addPermissions(allowedUsers, studyId);
```
hatimkhan, line 264: You can do this differently.  You can make the this.addPermissions throws an exception that contains the errors.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
SanketD92: Will do, good call.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
SanketD92: I was wondering if that should be P1 or P2. I guess it makes sense for P1 too.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
SanketD92: Ah, yes. That makes more sense.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -212,15 +237,463 @@ class EnvironmentMountService extends Service {
...
+
+    _.forEach(policyDoc.Statement, statement => {
+      if (
+        statement.Sid.startsWith('studyListS3Access') &&
+        statement.Effect === 'Allow' &&
+        statement.Resource === s3BucketName &&
+        !_.includes(statement.Condition.StringLike['s3:prefix'], studyPrefix)
+      ) {
+        statement.Condition.StringLike['s3:prefix'].push(studyPrefix);
+      }
```
jn1119, line 592: Shouldn't the if condition be evaluated after iterating through all the statements and then if all the statements didn't include the studyPrefix, only then we should push the studyPrefix.

addons/addon-base-raas/packages/base-raas-services/lib/environment/__tests__/environment-mount-service.test.js
```diff
@@ -0,0 +1,213 @@
...
+    it('should call remove when only removals are requested', async () => {
+      // BUILD
+      const updateRequest = {
+        usersToAdd: [{ uid: 'User1-UID', permissionLevel: 'admin' }],
+        usersToRemove: [{ uid: 'User2-UID', permissionLevel: 'readwrite' }],
+      };
+      const studyId = 'StudyA';
+      service.addPermissions = jest.fn();
+      service.removePermissions = jest.fn();
+      service.updatePermissions = jest.fn();
```
jn1119, line 119: Can we add tests for these methods- add, remove and update or not mock them out in some of the tests. Not sure if we are covering these tests in unit tests.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
SanketD92: There is only one studyListS3Access statement per bucket which holds all the studyPrefixes for that bucket, therefore only one in that list that would need that change. 

Even if access to all the studies in a certain bucket are taken away access from this policy, the studyListS3Access statement would still be present for this bucket with Condition.StringLike being an empty list.

addons/addon-base-raas/packages/base-raas-services/lib/environment/__tests__/environment-mount-service.test.js
SanketD92: I believe those would be prime candidates for integration tests.

SanketD92: As discussed, I'll add JSON schema validation.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
jn1119: Got it, thanks.

addons/addon-base-raas/packages/base-raas-services/lib/environment/__tests__/environment-mount-service.test.js
SanketD92: As discussed, pushed more unit tests that ensure logic is intact.

jn1119 (approved):
```
              |    |    |                 
             )_)  )_)  )_)              
            )___))___))___)\            
           )____)____)_____)\\
         _____|____|____|____\\\__
---------\                   /---------
  ^^^^^ ^^^^^^^^^^^^^^^^^^^^^
    ^^^^      ^^^^     ^^^    ^^
         ^^^^      ^^^
```

PR #204, opened 2020-11-10
feat: support ongoing sync (add, update, delete) from S3 to local file system - EC2 Windows

dpandya:
Issue #, if available:

Description of changes:
- Added support for ongoing sync (recurring downloads) from S3 to local EBS volumes for Service Catalog based EC2 Windows workspaces
- Modified the S3 synchronizer Go program to support the recurring sync as follows
    - If the `recurringDownloads` flag is set to `false`, the program will download data from S3 only once and further changes in S3 will not be synchronized locally.
    - If the `recurringDownloads` flag is set to `true`, the program will trigger a thread to periodically (controlled by `downloadInterval`) synchronize the changes from S3 to local file system as follows.
        - Any files present in S3 but not present locally will be downloaded
        - Any existing files updated in S3 will be re-downloaded and local files will be overwritten. If the files had any local changes then those changes will be lost. The program uses S3 object's `ETag` value to determine if the object has changed in S3 since the last download. The program will re-download only updated files.
        - Any files deleted from S3 but present locally will be deleted from local file system as well
    - `stopRecurringDownloadsAfter` can be passed to automatically stop recurring downloads after certain period. This can be passed as `-1` to NOT stop the recurring sync (i.e., keep thread performing the recurring sync indefinitely)

> :warning:  **Need to re-import the EC2 Windows Workspace Type to get these changes**

Checklist: 


addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
```diff
@@ -124,34 +282,49 @@ func getObjectsAll(
...
-			d.Concurrency = concurrency
-		})
+		shouldDownload := true
+
+		// Note that we cannot use os.IsExist(fileError) to check for file's existence
+		// os.IsExist and os.IsNotExist are error checker functions and only work when error is not nil
+		// The correct way to check if file exists is using !os.IsNotExist(fileError)
+		if _, fileError := os.Stat(destFilePath); !os.IsNotExist(fileError) {
+			// If the file has not changed in S3 since last download then skip downloading it
+			shouldDownload = hasFileChangedInS3(item)
```
jn1119, line 295: If the study was read only, should we disallow any modifications/writes to this local directory just like Goofys does. I know we aren't implementing Goofys here but thought we should discuss this out. Another option would be to have another flag which doesn't force the download if the file was modified and leaves it as it is. Do you know what is the expected behavior or what we have is what is desirable?

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
```diff
@@ -124,34 +282,49 @@ func getObjectsAll(
...
-			if debug == true {
+			if debug {
 				log.Println("Create file error: ", err.Error())
 			}
 			stats.errorPrefixes = append(stats.errorPrefixes, item.Key)
 			continue
 		}
 
+		downloader := s3manager.NewDownloader(sess, func(d *s3manager.Downloader) {
+			d.PartSize = 100 * 1024 * 1024
```
jn1119, line 318: nit: probably a comment here stating the units. I am guessing its 100 MB

addons/addon-raas-s3-copy/packages/s3-synchronizer/README.md
```diff
@@ -24,8 +32,12 @@ Usage of bin/s3-synchronizer-darwin-amd64:
         Whether to print debug information
   -destination string
         The directory to download to (default "./")
-  -pollInterval int
-        The delay (in seconds) between api http requests (default 60)
+  -recurringDownloads 
+        Whether to periodically download changes from S3 (default false)
+  -stopRecurringDownloadsAfter int
+        Stop recurring downloads after certain number of seconds. ZERO or Negative value means continue indefinitely. (default -1 i.e., indefinitely)
```
jn1119, line 38: What is the motivation behind having this flag for customers. I understand the need for this for say unit tests, but are there cases when customers would want to stop downloads if recurringDownloads is set to true?

addons/addon-raas-s3-copy/packages/s3-synchronizer/README.md
```diff
@@ -24,8 +32,12 @@ Usage of bin/s3-synchronizer-darwin-amd64:
...
+  -downloadInterval int
+        The interval at which to re-download changes from S3 in seconds. This is only applicable when recurringDownloads is true. (default 60)
```
jn1119, line 40: nit: Probably include that download/sync time from S3 is not counted towards wait time, and regardless of how much time the re-sync took we will wait for these many seconds.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-synchronizer_test.go
```diff
@@ -142,14 +148,273 @@ func TestMainImplForInitialDownloadEmptyMounts(t *testing.T) {
...
 	// ---- Run code under test ----
-	err := mainImpl(testAwsSession, debug, concurrency, testMountsJson, destinationBase)
+	err := mainImpl(testAwsSession, debug, false, -1, 60, concurrency, testMountsJson, destinationBase)
+	if err == nil {
+		// Fail test in case of no errors since we are expecting errors when passing invalid json for mounting
+		t.Logf("Expecting error when running the main s3-synchronizer with invalid testMountsJson but it ran fine")
+	}
+}
+
+// ######### Tests for Recurring Downloads #########
```
jn1119, line 164: Can we add one test for error handling in recurring downloads. Say S3 had some transient issue and sync method fails the first time but the expectation is that a re-sync would happen after downloadInterval is passed and hopefully succeed.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
```diff
@@ -86,24 +133,135 @@ func getBucketObjects(sess *session.Session, config *mountConfiguration, concurr
...
+
+	continueRecurringDownloads := true
+	setupStartTime := time.Now()
+
+	// Kick off thread for recurring download for this mount configuration
+	// This thread will push download stats to stats channel and the reporter thread will receive stats from the
+	// stats channel and report (print) the stats
+	go func() {
+		for continueRecurringDownloads {
+			stats := syncS3ToLocal(sess, config, concurrency, debug)
```
jn1119, line 230: I know you handle probably all the errors in sync method. I was wondering though if we should have some error handling here. Just for my education- in case sync method throws an error, is this thread still active or just dies out silently?

jn1119 (commented):
Implementation looks good to me. Had some minor comments.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
dpandya: In case of ReadOnly, the S3 changes should win. The function `"hasFileChangedInS3"` checks if the file has changed in S3 and needs re-download. This is not about changes in local directory.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
dpandya: In Go, there is no concept of "throwing an error". Since Go functions can return multiple values, the errors are just returned along with other values just like regular return values. Unless we explicitly read error and exit, returning an error does not automatically exit the thread.

jn1119: Thanks, good to know. Thanks for clarifying

addons/addon-raas-s3-copy/packages/s3-synchronizer/README.md
dpandya: Customer may have some temporarily running job pushing changes to S3 and keep monitoring S3 beyond that duration may be waste of CPU/Memory on the workspace instance if no more changes are expected to happen in S3. Having such a flag would allow for auto-stopping of the monitoring thread. 

Another use case may be some studies (say open data studies) which are not expected to change at all. There is no need to keep monitoring that bucket on running workspace. Having this parameter allows for stopping the monitoring thread pretty much immediately after the initial download in those cases.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
jn1119: Makes sense. Just for my clarification though- is the local study mount/directory read-only though? Or existing files can be modified or new files be added by the local user.

addons/addon-raas-s3-copy/packages/s3-synchronizer/README.md
jn1119: I see, let's keep it in that case. If you think these details will be user, feel free to add them to documentation as well.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
dpandya: No the directory will not be ReadOnly on the file system. User's changes will be overwritten by the synchronizer in case of read-only. Cannot make the directory/files read-only on the file system because the researchers are also logging in as the "Administrator" user - i.e., the same user the synchronizer is running as. Making them read-only will make the synchronizer also fail. 
Will add a separate ticket to investigate this further.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
dpandya: Done. Added the comment

addons/addon-raas-s3-copy/packages/s3-synchronizer/README.md
dpandya: Done. Added the following line

> Note that this does not include the download time. This specifies the duration in seconds to wait before initiating the next download after the previous one completes.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-synchronizer_test.go
dpandya: Good suggestion. Would be certainly be a nice test to have. Unfortunately, I couldn't find any straight forward way to simulate intermittent S3 failures. The test is using [gofakes3](https://github.com/johannesboyne/gofakes3) for simulating AWS S3 APIs. Between mocking S3 SDK and simulating S3 APIs, I had chosen "simulating" due to it being non intrusive and its ease of implementation.

For simulating intermittent S3 failures, here is what I tried without luck
1. Stop (Close) and Start the gofakes3 server  -- Result: gofakes3 server cannot be restarted.
2. Close and completely replace gofakes3 server with a new server -- Result: new gofakes3 server starts on another port but S3 session is already pinned to previous endpoint port.
3. Change the Endpoint in the AWS SDK Session to point to some non-existent URL and then swap it back to the gofakes3 url: The Endpoint swap never materializes after first usage. AWS Session caches it. 

Due to these, I will not invest more time on adding this test at the moment. I have tested S3 failure manually (not intermittent though), I have manually verified that the synchronizer thread does not exit and it backs-off for about 30 seconds and then attempts to synchronize again.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
jn1119: Separate ticket sounds good. Once you do that, mark this conversation as resolved.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-synchronizer_test.go
jn1119: Thanks for trying all the various approaches. I think the only way to make this work is probably to inject a new S3 session every time we synchronize. Is that feasible? On a related note re-using s3 session is credential rotation safe, right?

dpandya: Creating new AWS Session every time is an anti-pattern. AWS SDKs in almost all languages have some concept of "initialization" where you "initialize" the SDK with some configuration once and then keep creating service clients and interacting with those clients. 

One way to make this work would be to refactor the program especially the s3-downloader and s3-synchronizer to hold the session in a state (e.g., some struct) and then make the functions read the session from that state instead of passing the session as function argument. We can then swap out the session in that state in the test code temporarily to simulate failures and then swap it back. Since that would require some non-trivial refactoring. I will add a ticket for that.

jn1119: Sounds good to me, you can resolve this conversation once the ticket is created.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-download.go
dpandya: Done. Created ticket.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-synchronizer_test.go
dpandya: Done. Created a ticket.

jn1119 (approved):

PR #206, opened 2020-11-12
feat: Bidirectional S3/Local sync for Service Catalog based EC2 Windows workspaces

dpandya:
Issue #, if available:

Description of changes:

- Adds read/write support for Service Catalog based EC2 Windows workspaces. At high level this is implemented as two main threads (go routines) one that periodically syncs S3 changes to local machine and another that watches file system events and syncs local changes back to S3. 
- Modified the S3 synchronizer Go program to support the bi-directional sync as follows
    - Kick off a thread to watch changes in mapped directories and handle file create, update, delete, rename events
    - Dynamically watch child directories as they get created for the changes
    - Crawl the child directories when they are created to sync then up to S3. This is to handle an edge case when a file and the nested directories are created in one operation and the file gets created before the directory is registered for watching changes.
    - Maintain a persistent map (currently as a state file under user home directory) of S3 keys and their ETag values. This map is used for the followings
       - To not download files from S3 if they are not changed in S3
       - To avoid deleting file that is added locally during download synchronization

Currently there are following limitations to this implementation:

**LIMITATIONS:**

**S3 to Local Sync Limitations:**
- If the selected study is Read-Only, any changes made under the locally mapped study directory and it's subdirectories will be **LOST** after the periodic sync. No local changes will persist.
- There will be delay of at least the duration equal to the periodic download interval plus the download time for the S3 changes to reflect on local EBS volumes.
- Deleting a subdirectory in studies S3 location will leave the corresponding subdirectory as empty directory on local EBS volume. 

**Local to S3 Sync Limitations:**
- Will not upload changes from local to S3 if there is no change in file size (bytes)
- Will not upload changes from local to S3 if the file is empty (i.e., zero bytes)
- **Conflict resolution is undefined:** i.e., if a file is modified in S3 and locally at the same time, the behavior is undefined. Whichever change gets synchronized first may win.

**MANUAL TESTING:**

The following scenarios are manually tested (also added automated unit tests for them - except for the cross region studies)

**Read-Only**
* [x] Initial download (tested for study within the same region and also cross region - e.g., Open Data - 1000 Genomes study) 
* [x] Add to S3 — adds in local 
* [x] Update in S3 — updates in local
* [x] Delete from S3 — deletes from local

**Read-Write**
* [x] Add to S3 — adds in local 
* [x] Add to local — gets uploaded to S3
* [x] Update in S3  — updates in local
* [x] Update in local — updates in S3
* [x] Delete from S3 — deletes from local
* [x] Delete from local  — deletes in S3
* [x] Rename in S3 — renames in local
* [x] Rename in local — renames in S3
* [x] Create object in nested folders S3  — nested folders/files are created in local
* [x] Create nested folders/files in local — nested folders/files are uploaded to S3
* [x] Rename nested folders in S3 - nested folders are renamed in local
* [x] Rename nested folders locally — nested folders are renamed in S3

**Checklist:** 


addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-upload-file-watcher.go
```diff
@@ -2,132 +2,295 @@
...
+					log.Println("error:", err)
+				}
+			} else {
+				select {
+				case event := <-watcher.Events:
+					processFileWatcherEvent(&event)
+				case dirToUpload := <-dirRequiringCrawlCh:
+					uploadDir(dirToUpload, debug)
+				case err := <-watcher.Errors:
+					log.Println("error:", err)
```
jn1119, line 161: Can this duplicate code be moved to a method so that it can be referenced in stopUploadWatchersAfter and non stopUploadWatchersAfter cases.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-upload-file-watcher.go
```diff
@@ -2,132 +2,295 @@
...
+		if debug {
+			log.Println(filename, " size has not changed since last upload, skipping upload this time")
+		}
+	}
+
+	return nil
+}
+
+// Checks if the file's sizes are different on disk and in S3
+func areSizesDifferent(sess *session.Session, bucket string, fileKeyInS3 string, file *os.File) bool {
```
jn1119, line 254: Was looking at `aws s3 sync` documentation and they also include the criteria of the last modified time of the local file being newer than last modified time of the s3 object. Should that be added as well? If the size is same but the content is different, we shouldn't upload the file if it is older than one in S3.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-upload-file-watcher.go
dpandya: Not that I know of. They are two different `select` statements one with a timeout and other one without timeout. These `select` statements are different from the usual `switch` statements we see in other languages. The `select` here is waiting to receive events from 3 channels vs 4 channels (one channel being the one for timeout). To minimize duplication I had already moved all case handling logic to separate methods the only duplication at the moment is two `select` statements receiving from 3 vs 4 channels.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-upload-file-watcher.go
dpandya: I don't think it is necessary in our case because of the following reasons,
1. A file is uploaded to S3 only if there is CREATE or WRITE event at file system level. So the only reason why we attempt to upload is because either the file is created locally or modified locally and we received event from file system. So the case of local file being older than S3 and us trying to upload never happens.
2. If the file is modified directly on S3 (hence leaving us with older version on file system), the file on file system will be updated during the periodic download sync

In case of `aws s3 sync` it may be necessary to check file timestamps because the user specifies a source and a target and sync is explicitly triggered by the user and not happening in reaction to file system events. So a user may call `aws s3 sync` and specify a local directory as source and S3 location as target and the source may contain files older than the ones in target.

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-synchronizer.go
```diff
@@ -19,10 +19,13 @@ func main() {
 
 	sess := makeSession(profile, region)
 
-	mainImpl(sess, debug, recurringDownloads, stopRecurringDownloadsAfter, downloadInterval, concurrency, defaultS3Mounts, destinationBase)
+	// Passing stopUploadWatchersAfter as -1 to let file watchers continue indefinitely if mount is writeable
+	stopUploadWatchersAfter := -1
```
jn1119, line 23: Can you remind me the use of it in read-only use case as it seems like we are disabling it for read/write? Is it for unit testing mostly?

jn1119 (approved):

addons/addon-raas-s3-copy/packages/s3-synchronizer/src/s3-synchronizer.go
dpandya: stopUploadWatchersAfter is for stopping the writer thread. This is only for unit tests and is NOT exposed via the program arguments.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -783,16 +789,22 @@ class EnvironmentMountService extends Service {
...
       mounts = studyInfo.reduce(
         (result, { id, resources, category, writeable }) =>
           result.concat(
             resources.map(resource => {
               const { bucket, prefix } = parseS3Arn(resource.arn);
-              return { id, bucket, prefix, category, writeable };
+              const mount = { id, bucket, prefix, category, writeable };
+              if (category !== 'Open Data') {
+                mount.kmsKeyId = studyDataKmsKeyArn;
+              }
```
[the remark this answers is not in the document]
dpandya: The `getStudyAccessInfo` was not adding the KMS key arn to the mounts. This is not a bug but the kms key arn will be required when we work on BYOB. Since I was in this part of the code I quickly added this code and tested the generated mounts manually. There are no existing tests for that method. Adding test for this case would first require adding tests for that method in general.

jn1119 (approved):

jn1119 (approved):

PR #210, opened 2020-11-17
fix: handling policy names for windows envs

SanketD92:
Issue #, if available:
N/A

Description of changes:
Added logic to handle EC2 Windows permission propagation scenarios, with known S3 permission policy names.

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
@@ -667,16 +667,47 @@ class EnvironmentMountService extends Service {
     return policyDoc.Statement;
   }
 
-  async _getWorkspacePolicy(iamClient, env) {
-    if (!env.outputs) {
-      throw new Error('Environment outputs are not ready yet. Please make sure environment is in Completed status');
-    }
+  async _getPolicyDoc(possiblePolicyNames, roleName, workspaceRoleArn, iamClient) {
     const iamService = await this.service('iamService');
-    const workspaceRoleArn = _.find(env.outputs, { OutputKey: 'WorkspaceInstanceRoleArn' }).OutputValue;
+    const policyNamePrefix = `analysis-${workspaceRoleArn.split('-')[1]}`;
+
+    _.forEach(possiblePolicyNames, async possiblePolicyName => {
+      try {
+        const studyDataPolicyName = `${policyNamePrefix}-${possiblePolicyName}`;
+        const { PolicyDocument: policyDocStr } = await iamService.getRolePolicy(
+          roleName,
+          studyDataPolicyName,
+          iamClient,
+        );
+        if (!policyDocStr) {
+          return { policyDocStr, studyDataPolicyName };
+        }
+      } catch (err) {
+        // Do nothing.
+        // This is intentional, since we are just trying the possible known policy names that could exist in this IAM role
+      }
+    });
+
+    return { policyDocStr: undefined, studyDataPolicyName: undefined };
+  }
+
+  async _getWorkspacePolicy(iamClient, env) {
+    const workspaceRoleObject = _.find(env.outputs, { OutputKey: 'WorkspaceInstanceRoleArn' });
+    if (!workspaceRoleObject) {
+      throw new Error(
+        'Workspace IAM Role is not ready yet. Please make sure environment is in Completed status and retry the operation',
+      );
+    }
+    const workspaceRoleArn = workspaceRoleObject.OutputValue;
     const roleName = workspaceRoleArn.split('role/')[1];
-    const studyDataPolicyName = `analysis-${workspaceRoleArn.split('-')[1]}-s3-studydata-policy`;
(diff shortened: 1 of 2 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -667,16 +667,53 @@ class EnvironmentMountService extends Service {
     return policyDoc.Statement;
   }
 
+  async _getPolicyDoc(possiblePolicyNames, roleName, workspaceRoleArn, iamClient) {
```
dpandya, line 670: nit: would suggest adding JSDocs to this method since this method some non-obvious work, something like
```
This method looks for inline policy based on the given array of "possiblePolicyNames". The method returns as soon as it finds an inline policy in the given role (identified by the "roleName") with a matching name from the "possiblePolicyNames". If no inline policy is found with any of the names from the "possiblePolicyNames", the method returns undefined. 
```

dpandya, line 670: nit: Since you are already passing "iamClient", would suggest accepting "policyNamePrefix" instead of the "workspaceRoleArn".

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -667,16 +667,53 @@ class EnvironmentMountService extends Service {
...
+    const policyNamePrefix = `analysis-${workspaceRoleArn.split('-')[1]}`;
+
+    const policyDocsInfos = await Promise.all(
+      _.map(possiblePolicyNames, async possiblePolicyName => {
+        let policyDocStr = '';
+        let studyDataPolicyName = '';
+        try {
+          studyDataPolicyName = `${policyNamePrefix}-${possiblePolicyName}`;
+          policyDocStr = await iamService.getRolePolicy(roleName, studyDataPolicyName, iamClient).PolicyDocument;
+        } catch (err) {
```
dpandya, line 681: The "getRolePolicy" method internally already handles `NoSuchEntity` and `ResourceNotFoundException` exceptions and returns empty object so you will not get any error here when the policy does not exist.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
dpandya: nit: just realized `possiblePolicyNames` is more like  `possiblePolicyNameSuffixes`

SanketD92: Made the changes per our discussion

SanketD92: Sounds good. Done

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
SanketD92: Got you. Thanks for mentioning that. Made the changes per our discussion.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -667,17 +667,50 @@ class EnvironmentMountService extends Service {
     return policyDoc.Statement;
   }
 
+  /**
+   * This method looks for inline policy based on the given array of "possibleInlinePolicyNames".
+   * The method returns as soon as it finds an inline policy in the given role (identified by the "roleName")
+   * with a matching name from the "possibleInlinePolicyNames".
+   * If no inline policy is found with any of the names from the "possibleInlinePolicyNames", the method returns undefined.
+   *
+   * @param {Object} possibleInlinePolicyNames - Known possible policy role names we have used for out of the box CfN templates
```
dpandya, line 676: nit: `Known possible policy role names we have used for out of the box CfN templates` seems incorrect. This arg is not about role names, it's possible inline policy names (and not role names).

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -667,17 +667,50 @@ class EnvironmentMountService extends Service {
...
+   * * @param {Object} roleName - Name of the IAM role for the given workspace
+   * * @param {Object} iamClient
+   * @returns {Object} - Returns policyDocStr and studyDataPolicyName that was found on this role
```
dpandya, line 679: nit: This is outdated. We are not returning `studyDataPolicyName` anymore

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
SanketD92: Ah, that's right. Good catch

dpandya:
LGTM

dpandya (approved):
LGTM

PR #213, opened 2020-11-20
fix: Manage S3 bucket policy for Study Access

SanketD92:
Issue #, if available:
GALI-461

Description of changes:
_Bug synopsis:_
If you launch one workspace in project 1 with study 1
then you launch another workspace in same project (project 1) with same study (study 1)
then terminate one of the workspaces
the other workspace will also loose its access to the study

This PR prevents the above by managing the studies being used by various workspaces within each member account in a new DB table "EnvironmentResourceUsages". When the first workspace of a member account is mounted with a particular study, it creates a row for that study-account combination, and adds the permission to access the study in the S3 bucket policy. When the last workspace to use a given study within a member account is terminated, the permission is removed from the S3 bucket policy.

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
@@ -229,17 +229,17 @@ class EnvironmentMountService extends Service {
           listStatement.Principal.AWS = updateAwsPrincipals(
             listStatement.Principal.AWS,
             workspaceRoleArn,
-            `arn:aws:s3:::${s3BucketName}/${prefix}`,
+            `${s3BucketName}/${prefix}`,
           );
           getStatement.Principal.AWS = updateAwsPrincipals(
             getStatement.Principal.AWS,
             workspaceRoleArn,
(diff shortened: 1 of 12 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -79,34 +80,97 @@ class EnvironmentMountService extends Service {
...
+        principal: newPrincipal,
+        envId,
+      });
+      // If this is the first environment that requires sharing the resource with the principal
+      // then update
+      if (result.count === 1) {
+        if (Array.isArray(awsPrincipals)) {
+          awsPrincipals.push(newPrincipal);
+        } else {
+          awsPrincipals = [awsPrincipals, newPrincipal];
```
hatimkhan, line 101: shouldn't this be
`awsPrincipals = [...awsPrincipals, newPrincipal];`

hatimkhan, line 101: This might be problematic, because you are changing the reference of a passed parameter.

PR #215, opened 2020-11-23
fix: adding validation for study update route

SanketD92:
Issue #, if available:
N/A

Description of changes:
Fix an authz bug on update study API

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
@@ -222,7 +222,7 @@ describe('studyService', () => {
     it('should get the correct allowed studies ONLY (admin, R/O, R/W)', async () => {
       // BUILD, OPERATE and CHECK
       expect(
-        service._getAllowedStudies({
+        service.getAllowedStudies({
           adminAccess: ['studyA'],
           readonlyAccess: ['studyB'],
           readwriteAccess: ['studyC'],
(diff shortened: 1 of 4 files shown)


nguyen102:
Can we add a note in the `CHANGELOG.md` for this change?

Maybe something like
```
## [1.4.2] - 2020-11-23

### Added 
- fix: Fix a bug on the update study API

We recommend to apply this patch as soon as possible
```

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
```diff
@@ -251,6 +251,50 @@ describe('studyService', () => {
...
+      // OPERATE
+      try {
+        await service.update(
+          { principal: { userRole: 'researcher' }, principalIdentifier: { uid: 'someRandomUserUid' } },
+          dataIpt,
+        );
+        expect.hasAssertions();
+      } catch (err) {
+        // CHECK
+        expect(err.message).toEqual('Resources can only be updated for Open Data study category');
```
hatimkhan, line 272: For future tests, you can use expect(<>).rejects.toThrow(..)

hatimkhan (dismissed):

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
```diff
@@ -251,6 +251,50 @@ describe('studyService', () => {
...
+      };
+      service.audit = jest.fn();
+
+      // OPERATE
+      try {
+        await service.update(
+          { principal: { userRole: 'researcher' }, principalIdentifier: { uid: 'someRandomUserUid' } },
+          dataIpt,
+        );
+        expect.hasAssertions();
```
nguyen102, line 269: Do we expect atleast one assertion for all tests in this file? If yes, we can put ` expect.hasAssertions();` in the `beforeEach`.

jn1119 (dismissed):

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
SanketD92: Will do. Thanks

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
SanketD92: Hmm, not sure if we should. Most of the tests are built are on a case by case basis. Some require the assertion check since they depend on async calls, while some don't.

hatimkhan (approved):

PR #216, opened 2020-11-25
fix: changed RStudio CSP headers to allow uploads from same-origin

Ludo171:
**Object of this PR**
This PR aims to fix an error in the default RStudio workspaces. This error prevents users to upload data from their computer to the RStudio server.

**Description of the error :**
When a user clicks on the "upload" button available on the RStudio interface, the following error is raised.  
The "upload" pop-up frizzes the whole page, and no upload is performed.  
It concerns too-restrictive CSP headers that prevent uploads from any domain name.  
image

**Description of changes:**
I modified the CSP headers and the X-Frame-Options headers of the nginx server to allow uploads from the same own domain name (via the HTTPS RStudio interface).
I tried those changes. It erases the error, and allows to upload data as expected without errors.

**Checklist:** 

### DIFF
--- a/main/solution/machine-images/config/infra/files/rstudio/nginx.conf
+++ b/main/solution/machine-images/config/infra/files/rstudio/nginx.conf
@@ -51,11 +51,11 @@ http {
     ssl_trusted_certificate cert.pem;
 
     add_header Referrer-Policy same-origin always;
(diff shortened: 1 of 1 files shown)


SanketD92 (approved):
LGTM

PR #217, opened 2020-11-25
fix: file not found error in download-env-config script

sebastianplesciuc:
### Issue #, if available:
None

### Description of changes:

Tested the CICD Pipeline provisioning and deployment in a new AWS account and it failed with the following error:
```
./scripts/download-env-config.sh: line 19: +config_target_path=main/config/settings/demo.yml: No such file or directory
```
Fixed by removing the `+` which caused the file to be checked before the actual `if`, forcing the user to commit the `demo.yml` file to the repository.

Refactored the `scripts/download-env-config.sh` script with the following things in mind:
- use of `util.sh` sourcing as the other scripts in folder
- prefer absolute path building versus relying on relative paths
- prefer UPPERCASE for variable names (Unix shell variables convention)
- use `-e` to check for file existence instead of `-f` (in case of symlinks)

### Checklist: 

* [x] If you had to run manual tests, have you considered automating those tests by adding them to [end-to-end tests](../main/end-to-end-tests/README.md)?


SanketD92 (approved):
LGTM

PR #218, opened 2020-11-30
feat: bring your own bucket - registration part

hatimkhan:
This PR is one of many to come.  It implements part of the bucket and study registration logic. The following is a list of everything this PR introduces or modifies.

**Terminology reminder**
* **Entity** (a.k.a **data object**) represents the model objects in in the backend, with one important observation, an entity is a data object that does not have any methods or functions.  This is intentional to make it straightforward to pass entities between services and across boundaries.  However, in many cases we need to ask questions about the state of an entity and we capture the logic to help us ask/answer these questions in functions that we keep under helper/entities/<entity-type>-methods.js. Entities can only be obtained from services that manage them. A service that manages an entity is responsible for persisting and retrieving the entity from the database (if needed).
* **dbEntity** (a.k.a **db record**/**db object**) is the database item representation of the entity.  It might contain a few attributes that are only needed so that we can store the entity correctly in the database. The service that is responsible for managing the entity is in charge of transforming this dbEntity to the expected Entity object

**New RESTful APIs**

*New collection resources:*

* data-sources
* data-sources/accounts
* data-sources/accounts/:id/buckets
* data-sources/accounts/:id/buckets/:name/studies

*New end points:*

* POST /api/data-sources/accounts/
* GET /api/data-sources/accounts/
* PUT /api/data-sources/accounts/:id/
* POST /api/data-sources/accounts/:id/buckets
* POST /api/data-sources/accounts/:id/buckets/:name/studies

**New Services**

* DataSourceAccountService
    * This service is responsible for persisting the data source account information, (DsAccountEntity).
* DataSourceBucketService
    * This service is responsible for persisting the data source bucket information, (DsBucketEntity)
* DataSourceRegistrationService
    * This service encapsulates the logic involved when a study is registered. This includes calling the StudyService to create the study entity in the database and giving a chance for plugins to participate in this logic. We have one plugin right now that implement the ‘roles-only’ strategy. This plugin allocated an application role entity in the database and stores the appRoleArn as an attribute in the study entity.
* ApplicationRoleService
    * WIP - should not be considered as part of the PR review

**Update To Existing Services**

* StudyService
    * New methods are added to implement the facade pattern so that there is no need for other services to directly interact with the the StudyPermissionService.  This allows us to enforce invariants in the StudyService rather than having that logic spread in many places in the code.
    * A new method **register** is introduced. The purpose of this method is to create the study entity in the database including the study permissions entity and the user permissions entities.
    * Removed StudyService.delete method, it is not being used and should not be used this way.
    * Removed accessType from the update-study.json schema since this is not supported for this release.
    * See before and after diagrams.
* StudyPermissionsService
    * This service is now made to be a strict delegate for the StudyService, this means that there is no longer a need for other services to interact directly with this service, instead the StudyService has all the facade needed for other services to use.
    * As a result, this service needed to be rewritten to capture all the logic for access types and to make this service a strict delegate service as mentioned above.
    * New unit tests are created for the rewritten service.
    * See before and after diagrams

* EnvironmentMountService
    * A few methods were removed since they are no longer needed after the StudyPermissionsService rewrite.
    * The getStudyAccessInfo() method was updated to account for the registered studies. However, the creation of the iam document policy has not been addressed yet.
   * The S3 mounts information that is created in environment-mount-service.js and given to cloudformation as  the value for the s3Mounts param has a slightly different shape when internal studies are used vs. external studies are used.
      * For internal studies, a study entry in the S3Mounts has this shape:
        * id
        * bucket
        * prefix
        * writeable
        * kmsKeyId  (we should rename this to kmsArn to match the one from external studies)
      * For external studies, a study entry in the S3Mounts has this shape:
        * id
        * bucket
        * region
        * awsPartition
        * prefix
        * kmsArn
        * readable
        * writeable
        * roleArn
    * The getStudyAcccessInfo() needs to be updated to accept the environment id too. In fact, we might need to rethink this logic a bit and refactor it.

**New DB Tables**

* DsAccounts
    * Stores information about the data source account and the buckets, such as the account id, contact info and name
* RoleAllocations
    * WIP - not to be reviewed as part of the PR

**Update To Existing DB Tables**

* Studies
    * The following attributes are introduced: folder, accountId, status, statusMsg statusAt, qualifier, bucket, awsPartition, region, kmsArn, kmsScope, accessType and appRoleArn

**New Extension Point**

* study-access-strategy
    * Methods:
        * onStudyRegistration
* ds-account-authz
    * Allows plugins to participate in the authorization logic when data source accounts are being created/updated/deleted
* study-authz
    * Allows plugins to participate in the authorization logic for the study permissions.
    * Actions:
        * get-study-permissions
        * get-users-permissions
        * create-study-permissions
        * update-study-permissions
* application-role-authz
    * Allows plugins to participate in the authorization logic for the application role
    * Actions:
        * read
        * allocate
        * update-status

Checklist:


addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -767,83 +801,128 @@ class EnvironmentMountService extends Service {
...
+        id,
+        category,
+        bucket,
+        region,
+        awsPartition,
+        prefix: folder,
+        resources,
+        // roleArn, this not available from the study entity, it needs to be checked out
+        kmsArn,
+        readable: read,
```
SanketD92, line 837: Where is the `readable` parameter being used in the services?

addons/addon-base-raas/packages/base-raas-services/lib/study/study-permission-service.js
```diff
@@ -16,361 +17,543 @@
...
-
     // Setup DB helpers
-    const dbService = await this.service('dbService');
-    this.tableName = this.settings.get(settingKeys.tableName);
-    this._getter = () => dbService.helper.getter().table(this.tableName);
-    this._updater = () => dbService.helper.updater().table(this.tableName);
-    this._query = () => dbService.helper.query().table(this.tableName);
-    this._deleter = () => dbService.helper.deleter().table(this.tableName);
-    this._scanner = () => dbService.helper.scanner().table(this.tableName);
+    const [dbService] = await this.service(['dbService']);
```
SanketD92, line 119: Why do we need a list to initialize a single service? Just curious..

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
```diff
@@ -157,23 +140,17 @@ async function configure(context) {
...
-      if (study.category === 'My Studies') {
-        throw context.boom.forbidden('Permissions cannot be set for studies in the "My Studies" category', true);
-      }
-      if (study.category === 'Open Data') {
-        throw context.boom.forbidden('Permissions cannot be set for studies in the "Open Data" category', true);
-      }
-
       const result = await lockService.tryWriteLockAndRun({ id: `${studyId}-workspaces` }, async () => {
-        const updateOutcome = await studyPermissionService.update(requestContext, studyId, updateRequest);
+        const updateOutcome = await studyService.updatePermissions(requestContext, studyId, updateRequest);
```
SanketD92, line 153: Sanity check - Given the changes to the getStudyAccessInfo() method, does permission propagation still work as expected for upgrade scenarios?

addons/addon-base-raas/packages/base-raas-services/lib/study/helpers/entities/study-methods.js
```diff
@@ -0,0 +1,195 @@
...
+  return studyEntity.category === 'My Studies';
+}
+
+function isReadonly(studyEntity = {}) {
+  // Due to backward compatibility, not all studyEntity have the property 'accessType', therefore,
+  // by default the accessType if not found is readwrite. Remember the accessType represents the
+  // the max permissions allowed on the study.
+  const accessType = _.get(studyEntity, 'accessType', 'readwrite');
+  return accessType === 'readonly';
+}
```
jn1119, line 134: Where do we decide the accessType of the study? Will that happen from the UI going forward?

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
hatimkhan: This is mainly done to be explicit, consistent and handle the accessType = 'writeonly', even though we don't have code that handles 'writeonly' all the way. It is better to be explicit in this case.  The purpose of this method is to tell us information about the mount, with this new code, we are explicitly list out the permissions.

addons/addon-base-raas/packages/base-raas-services/lib/study/study-permission-service.js
hatimkhan: It was just a left over from refactoring, there used to be more than one service.
There is no harm from leaving it the way it is, but will change it anyway.

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
hatimkhan: I didn't do an end to end test for the permission propagation.  Most of the existing logic will need to be refactored now that we have BYOB in the picture.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/application-role-service.js
```diff
@@ -0,0 +1,263 @@
...
+    // - Get all the application roles entities by the account and the bucket
+    // - Do we have the study already part of an existing application role?
+    //   - if yes, then return the application role
+    // - For each application roles, do the following:
+    //   - add study to the role entity
+    //   - check if size has reached the max
+    //    - if not then we found a place for the study and we update the application role
+    //      entity and return it
+    // - If none is found, create a new application role, add study to it, store it
+    //   in the database  and return the application role entity
```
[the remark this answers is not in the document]
hatimkhan: Correct. I intend to obtain a lock in the registration service just before calling this service.  Thanks for the heads up.  I will add the lock in the registration service in an upcoming PR.

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
```diff
@@ -44,12 +39,11 @@ async function configure(context) {
...
     '/:id',
     wrap(async (req, res) => {
+      const studyService = await context.service('studyService');
       const id = req.params.id;
       const requestContext = res.locals.requestContext;
 
-      await studyPermissionService.verifyRequestorAccess(requestContext, id, req.method);
-
-      const result = await studyService.mustFind(requestContext, id);
+      const result = await studyService.getStudyPermissions(requestContext, id);
```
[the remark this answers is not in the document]
hatimkhan: Correct. The previous code was doing a look up for permissions inside the studyPermissionService.verifyRequestorAccess.  It was doing by calling the database twice, one to find the study entity and then another time to find the study permissions entity.  Then, it was looking up the study once more time to return it :) 

With the new code, getStudyPermissions() is taking care of that.  Maybe a better name could have been getStudyWithPermissions()?

addons/addon-base-raas/packages/base-raas-services/lib/environment/__tests__/environment-mount-service.test.js
```diff
@@ -77,102 +73,103 @@ describe('EnvironmentMountService', () => {
...
-      expect(s3Mounts[1].kmsKeyId).toEqual(kmsKeyArn);
-
-      expect(s3Mounts[2].id).toEqual(openDataStudyId);
-      expect(s3Mounts[2].bucket).toEqual(bucket);
-      expect(s3Mounts[2].prefix).toEqual(`${prefix}/${openDataStudyId}/`);
-      expect(s3Mounts[2].writeable).toEqual(false);
-      expect(s3Mounts[2].kmsKeyId).toBeUndefined(); // kmsKeyId should not be set for open data studies
-    });
-  });
+  // TODO - redo it so that it works with BYOB cases
```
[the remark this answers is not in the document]
hatimkhan: I should have removed all the commented out code, but I thought it is better to leave it there to remind me of rewriting the test cases for it.  The whole environment-mount-service.js will need to be refactored but this is still WIP.  FYI, I keep a list of todos in my Quiver notes, so I will add this item to it.  Thanks.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -767,83 +801,128 @@ class EnvironmentMountService extends Service {
...
+        writeable: write, // TODO - fix the spelling mistake in Go too.  It should be 'writable'.
```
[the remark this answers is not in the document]
hatimkhan: agree

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -63,25 +64,58 @@ class EnvironmentMountService extends Service {
...
-          bucket,
-          prefix,
-          writeable,
-          kmsKeyId,
-        })),
-      ),
+      s3Mounts: JSON.stringify(s3Mounts),
       iamPolicyDocument: JSON.stringify(iamPolicyDocument),
       environmentInstanceFiles: this.settings.get(settingKeys.environmentInstanceFiles),
-      s3Prefixes: s3Mounts.filter(({ category }) => category !== 'Open Data').map(mount => mount.prefix),
```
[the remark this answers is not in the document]
hatimkhan: We still do this logic in line 106.  The previous code was doing the filtering and determination in a few places, I was able to consolidate all that logic in lines 94 to 112.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
```diff
@@ -15,6 +15,8 @@
 
 const ServicesContainer = require('@aws-ee/base-services-container/lib/services-container');
 const JsonSchemaValidationService = require('@aws-ee/base-services/lib/json-schema-validation-service');
+
+jest.mock('@aws-ee/base-services/lib/logger/logger-service');
```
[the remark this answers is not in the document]
hatimkhan: Without it, we have too many distracting console messages that are being printed during the tests.

addons/addon-base-raas/packages/base-raas-services/lib/schema/register-data-source-bucket.json
```diff
@@ -0,0 +1,37 @@
...
+      "maxLength": 100
+    },
+    "awsPartition": {
+      "type": "string",
+      "enum": ["aws", "aws-cn", "aws-us-gov"]
+    },
+    "kmsArn": {
+      "type": "string",
+      "maxLength": 90
+    },
```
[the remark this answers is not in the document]
hatimkhan: It can be different.  When an admin is registering a study, the admin is asked to specify if the bucket has a default kms and then for each study the admin is also asked if the study is using the bucket default kms or has its own kms key.  We capture enough information in both the bucket entity and the study entity to be able to determine if a study should be using the default bucket kms (when that is the case, the study kmsArn will be empty).  Let me know if this answered your question.

addons/addon-base-raas/packages/base-raas-services/lib/study/helpers/entities/study-methods.js
hatimkhan: It will only happen when a study is being registered via the api  POST /api/data-sources/accounts/:id/buckets/:name/studies is called.

Existing study creation api is not expected to change.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/application-role-service.js
```diff
@@ -0,0 +1,263 @@
...
+    if (!_.isUndefined(appRoleEntity)) return appRoleEntity;
+
+    appRoleEntity = undefined;
+    // eslint-disable-next-line consistent-return
+    _.forEach(appRoleEntities, entity => {
+      if (!maxReached(addStudy(entity, studyEntity))) {
+        appRoleEntity = entity;
+        return false; // To stop the loop since we found a role
+      }
+    });
```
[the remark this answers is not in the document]
hatimkhan: Sure, I can do that. Let me first see if I no longer need any more logic inside line 145, since this is still WIP.

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
SanketD92: I'm guessing that's a problem, since we'd be potentially breaking R/W functionality for users wishing to switch permissions back and forth on existing environments.

SanketD92 (approved):
LGTM

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/study-controller.js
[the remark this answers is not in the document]
jn1119: getStudyWithPermissions sounds good

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/application-role-service.js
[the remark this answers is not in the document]
jn1119: So can we resolve this for now and it will be addressed in the follow up PR.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/application-role-service.js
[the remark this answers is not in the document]
jn1119: So can we resolve this for now and it will be addressed in the follow up PR.

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-permission-service.test.js
```diff
@@ -12,924 +12,575 @@
...
+    container.register('lockService', new LockService());
+    container.register('userService', new UserService());
+    container.register('pluginRegistryService', new PluginRegistryService());
+    container.register('authorizationService', new AuthService());
+    container.register('auditWriterService', new AuditService());
     container.register('studyPermissionService', new StudyPermissionService());
-
     await container.initServices();
 
     service = await container.find('studyPermissionService');
```
jn1119, line 63: Is there a test covering the part where a readOnly study cannot have readWrite or writeOnly users. Perhaps I missed it.

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
```diff
@@ -522,126 +602,5 @@ describe('studyService', () => {
...
-
-      // OPERATE
-      await service.update({}, dataIpt);
-
-      // CHECK
-      expect(dbService.table.update).toHaveBeenCalled();
-      expect(service.audit).toHaveBeenCalledWith({}, { action: 'update-study', body: undefined });
-    });
-
-    it('should succeed with readwrite accessType', async () => {
```
jn1119, line 578: I think there is value in keeping one of these update tests which were changing accessType and assert that we cannot modify accessType any more. Perhaps I missed the test where we were doing this.

jn1119 (commented):
LGTM overall. I have a few open questions but if you want to address them as a follow up in next PR we can close this PR since we aren't merging into master right away.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/application-role-service.js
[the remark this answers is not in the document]
hatimkhan: When I looked at the logic again, I realized that it might be better to leave it as a foreach loop rather than find. The main difference is that when I think of find, I think of a method that does not mutate anything.  While in my case, I need to add the study to the role entity before I can determine if the max size has been reached. I am thinking it might be better to leave it as a foreach loop.

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-permission-service.test.js
hatimkhan: Yes, I have many tests to cover this case in addons/addon-base-raas/packages/base-raas-services/lib/study/helpers/entities/__tests__/study-methods.test.js

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
hatimkhan: Good point.  Will do that.

PR #234, opened 2020-12-08
Feat: BYOB Reachability Check

SanketD92:
Issue #, if available:
GALI-569

Description of changes:
Controller routes API requests for reachability to dataSourceReachabilityService. This picks the logic to evaluate reachability and subsequently, a workflow is triggered based on the provided id, status, and type.

Checklist: 


addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
```diff
@@ -526,6 +527,21 @@ class StudyService extends Service {
     return result;
   }
 
+  /**
+   * IMPORTANT: Do NOT call this method directly from a controller, this is because
+   * this method does not do any authorization check.  It will return the study given
+   * a study id no matter who the requestContext principal is.
+   */
+  async listBucketsForAccount(requestContext, { accountId }) {
```
hatimkhan, line 535: Are you trying to list buckets that belong to a data source account?
If so, the dataSourceAccountService already has this method.  The study service does not know how to get a list of buckets given a data source account.

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/ds-account-status-change/ds-account-status-change.js
```diff
@@ -0,0 +1,45 @@
...
+    // Get common payload params and pull environment info
+    const requestContext = await this.payload.object('requestContext');
+    const accountId = await this.payload.object('id');
+
+    // For dsAccount, find all (not just unreachable) studies
+    const studies = await studyService.listStudiesForAccount(requestContext, accountId);
+
+    // For each study, reach out
+    await Promise.all(
+      _.map(studies, async study => {
```
hatimkhan, line 38: We don't want to issue thousands of call at the same time,   you can use 'processInBatches'  from @aws-ee/base-services/lib/helpers/utils.  An example of how to use it can be found here: https://github.com/awslabs/service-workbench-on-aws/blob/69d207341aeb85b8108328dc689f4ce2d41bde71/addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/legacy/environment-resource-service.js#L82

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-registration-service.js
```diff
@@ -79,6 +87,150 @@ class DataSourceRegistrationService extends Service {
...
+    // Write audit event
+    await this.audit(requestContext, {
+      action: 'check-study-reachability',
+      body: { id, type },
+    });
+
+    return outputVal;
+  }
+
+  async _assumeAppRole(requestContext, appRoleArn) {
```
hatimkhan, line 184: Have you considered the method in aws-service.js? 
Take a look at (if you haven't already) https://github.com/awslabs/service-workbench-on-aws/blob/e63120d7ebaf442b81b144cce6962ea085d0aa66/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-service.js#L177

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-registration-service.js
```diff
@@ -79,6 +87,150 @@ class DataSourceRegistrationService extends Service {
     return result.studyEntity;
   }
 
+  async bulkReach(requestContext, { status }) {
```
hatimkhan, line 90: Can you remove all the reachability methods to a new service?  Maybe we can name the service 'StudyReachabilityService'?   This is because reachability testing is not limited to study registration.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-registration-service.js
hatimkhan, line 184: Also, I made it so that, we can assume the application role to check if the study is readable, can you add that logic? so you can do something like
`const s3Client = await aws.getClientSdkForRole({ roleArn, appRoleArn, clientName: 'S3' }); 
// then use s3Client to read the head of an object`

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
SanketD92: Sorry, the name for this method is listStudiesForAccount. This will get all studies that are linked to a given accountId.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-registration-service.js
SanketD92: Will do. Thanks

SanketD92: Done

SanketD92: I've named it 'DataSourceReachabilityService' though, since study reachability depends on the DS Accounts being reachable first.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-registration-service.js
SanketD92: Done.

SanketD92: Done

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/ds-account-status-change/ds-account-status-change.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
```diff
@@ -0,0 +1,197 @@
...
+    this.dependency([
+      'auditWriterService',
+      'dataSourceAccountService',
+      'studyService',
+      'workflowTriggerService',
+      'aws',
+    ]);
+  }
+
+  async bulkReach(requestContext, { status }, forceCheckAll = false) {
```
hatimkhan, line 37: What do you think about making third argument as an option object? This will allow us to add more options in the future. So something like this:
`async bulkReach(requestContext, { status }, { forceCheckAll = false } = {}) `

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
```diff
@@ -0,0 +1,197 @@
...
+      action: 'bulk-check-reachability',
+      body: { status },
+    });
+  }
+
+  async reachDsAccount(requestContext, { id, type }, forceCheck = false) {
+    const accountService = await this.service('dataSourceAccountService');
+    const dataSourceAccount = await accountService.mustFind(requestContext, { id });
+
+    if (!dataSourceAccount.status) {
```
hatimkhan, line 59: What is the purpose of this?  DataSourceAccountService will only return data source accounts.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
```diff
@@ -0,0 +1,197 @@
...
+    await accountService.updateStatus(requestContext, dataSourceAccount, { status: newStatus, statusMsg });
+    const outputVal = newStatus;
+
+    // Write audit event
+    await this.audit(requestContext, {
+      action: 'check-dsAccount-reachability',
+      body: { id, type },
+    });
+
+    return outputVal;
```
hatimkhan, line 100: How about we return an object rather than just a string?  something like
{ status, statusMsg }
This will allow us in the future to also return the workflow instance id, which will help the caller of the API to check on the workflow status later.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
```diff
@@ -0,0 +1,197 @@
...
+    await studyService.updateStatus(requestContext, studyEntity, { status: newStatus, statusMsg });
+    const outputVal = newStatus;
+
+    // Write audit event
+    await this.audit(requestContext, {
+      action: 'check-study-reachability',
+      body: { id, type },
+    });
+
+    return outputVal;
```
hatimkhan, line 135: See my comments in the reachDsAccount() method regarding the output shape

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
```diff
@@ -0,0 +1,197 @@
...
+      await s3Client.headObject({ Bucket: studyEntity.bucket, Key: studyEntity.folder });
+      reachable = true;
+    } catch (err) {
+      // Error is expected if assuming role is not successful yet
+      reachable = false;
+    }
+    return reachable;
+  }
+
+  async attemptReach(requestContext, { id, status, type }, forceCheckAll = false) {
```
hatimkhan, line 157: See my other comment regarding the option object

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
```diff
@@ -0,0 +1,197 @@
...
+
+    if (id === '*') {
+      await this.bulkReach(requestContext, { status }, forceCheckAll);
+    } else if (type === 'dsAccount') {
+      outputVal = await this.reachDsAccount(requestContext, { id, type }, forceCheckAll);
+    } else if (type === 'study') {
+      outputVal = await this.reachstudy(requestContext, { id, type });
+    }
+
+    return outputVal;
```
hatimkhan, line 184: See my comments regarding the output shape

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
```diff
@@ -526,6 +527,21 @@ class StudyService extends Service {
...
+   * a study id no matter who the requestContext principal is.
+   */
+  async listStudiesForAccount(requestContext, { accountId }) {
+    const result = await this._query()
+      .index(this.accountIdIndex)
+      .key('accountId', accountId)
+      .limit(1000)
+      .query();
+
+    return result;
```
hatimkhan, line 542: We are missing the logic to convert dbEntities to just studyEntities.
So, you want to do something like
`return _.map(result, toStudyEntity);`
Look for an example in the same service.

SanketD92: Sure, but only the `id` field is used from the list returned from this method.

hatimkhan: For the use of 'toStudyEntity', this is not related to how many properties you need.  This is more related to that fact that dbEntities are never exposed to any methods,  because sometime dbEntities will include properties that are purely for database storage, such as composite keys with prefixes like 'ACT#', etc.

Aside from the need to use toStudyEntity, do you only need the id or you also need the bucket and region information stored in the study entity?

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
```diff
@@ -0,0 +1,262 @@
...
+        requestContext,
+        { workflowId: workflowIds.accountStatusChange },
+        {
+          requestContext,
+          id,
+          type,
+        },
+      );
+    }
+    if (_.isArray(unreachableAppRoles) && unreachableAppRoles.length > 0) {
```
hatimkhan, line 98: Is there a reason why you are checking if the unreachableAppRoles is an array? 
Maybe you just wanted to do !_.isEmpty(unreachableAppRoles)?

main/solution/backend/src/lambdas/data-source-reachability-daily/handler.js
```diff
@@ -0,0 +1,34 @@
...
+
+const handler = async () => {
+  const container = new ServicesContainer(['settings', 'log']);
+  // registerServices - Registers services by calling each service registration plugin in order.
+  await registerServices(container, pluginRegistry);
+  await container.initServices();
+  const dataSourceReachabilityService = await container.find('dataSourceReachabilityService');
+  const userContext = getSystemRequestContext();
+  await dataSourceReachabilityService.attemptReach(userContext, { id: '*' }, { forceCheckAll: true });
+  return { statusCode: 200 };
```
hatimkhan, line 30: Can you remove the status code?

SanketD92: Yup

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
SanketD92: Yeah, I can do that

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/plugins/workflow-steps-plugin.js
```diff
@@ -30,6 +30,10 @@ const stopEC2Environment = require('../steps/stop-ec2-environment/stop-ec2-envir
 const stopEC2EnvironmentYaml = require('../steps/stop-ec2-environment/stop-ec2-environment.yml');
 const networkInfra = require('../steps/storage-gateway/create-network-infrastructure');
 const networkInfraYaml = require('../steps/storage-gateway/create-network-infrastructure.yml');
+const bulkReachabilityInfra = require('../steps/bulk-reachability-check/bulk-reachability-check');
```
hatimkhan, line 33: Any reason why you are calling it bulkReachability**Infra** ,  why 'Infra'?

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/plugins/workflow-steps-plugin.js
```diff
@@ -30,6 +30,10 @@ const stopEC2Environment = require('../steps/stop-ec2-environment/stop-ec2-envir
...
+const bulkReachabilityYaml = require('../steps/bulk-reachability-check/bulk-reachability-check.yml');
+const dsAccountStatusChangeInfra = require('../steps/ds-account-status-change/ds-account-status-change');
```
hatimkhan, line 35: why 'Infra'?

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/bulk-reachability-check/bulk-reachability-check.js
```diff
@@ -0,0 +1,45 @@
...
+    this.print('start pinging data source accounts');
+
+    // Get services
+    const dataSourceReachabilityService = await this.mustFindServices('dataSourceReachabilityService');
+
+    // Get common payload params and pull environment info
+    const requestContext = await this.payload.object('requestContext');
+    // If you specify an id, you can’t specify a status filter
+    const forceCheck = await this.payload.boolean('forceCheckAll');
+    const dsAccountIds = await this.payload.object('dsAccountIds');
```
hatimkhan, line 30: Do you know how many ids we might have?  With StepFunction we can only use 32K (which we share with all the other payload items)?

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
```diff
@@ -0,0 +1,262 @@
...
+    const dsAccountIds = await this._getDsAccountsWithStatus(requestContext, status);
+
+    if (!_.isEmpty(dsAccountIds)) {
+      await workflowTriggerService.triggerWorkflow(
+        requestContext,
+        { workflowId: workflowIds.bulkCheck },
+        {
+          requestContext,
+          status,
+          dsAccountIds,
```
hatimkhan, line 51: You might want to leave a TODO comment here stating that if we have hundreds of account, we need to store the account ids in an s3 object and pass the s3 url to the workflow and the workflow step will have to get the object from S3.  This is because a workflow payload has a size limit of 32k.

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/plugins/workflow-steps-plugin.js
SanketD92: Fixed

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/plugins/workflow-steps-plugin.js
SanketD92: Fixed

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/bulk-reachability-check/bulk-reachability-check.js
SanketD92: Good point. Although for now we can bypass that scenario by adding a guard to only bulk check for a max of 1000 account Ids. We can add chunking eventually.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
SanketD92: For now, I'm adding a guard to not trigger a workflow if number of DS Accounts needing a check exceeds 1000.

hatimkhan (dismissed):
LGTM.  Please take a look at the failed test.  You most likely need to run install and commit the latest pnpm lock file.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-account-service.js
```diff
@@ -238,13 +238,39 @@ class DataSourceAccountService extends Service {
...
       async () => {
-        throw this.boom.notFound(`The data source account entity "${arn}" does not exist`, true);
+        throw this.boom.notFound(`The data source account entity "${id}" does not exist`, true);
       },
     );
 
     return toDsAccountEntity(dbEntity);
   }
 
+  async updateStackCreated(requestContext, { stackCreated, dataSourceAccount } = {}) {
```
hatimkhan, line 248: To be consistent with the rest of the file,  can you rename dataSourceAccount to dsAccountEntity?

hatimkhan: Or you can just pass { id, stackCreated } = {}
Then in the accountIdCompositeKey.encode(dataSourceAccount) you can do this instead
accountIdCompositeKey.encode({ id } )

SanketD92: Done

hatimkhan (approved):
LGTM

PR #236, opened 2020-12-10
feat: bring your own bucket - role allocation implementation

hatimkhan:
This PR wraps up the work related to the implementation of the role allocation core logic.  This PR includes the file system service implementation.  Together with the application role service (submitted in a previous PR), these two services are the core pieces for the role allocation logic.  The file system service code is heavily documented to aid with understanding this nontrivial allocation logic.  The following is a primer and a high level logic of how the file system service role works.

**Primer**

- We capture the relations between four items: fs role, study, member account and environment.
- We keep the relationship information in three database items across two tables.
    1. Study to fs roles, this is a one-to-many relationship (but we need to make it many-to-many in the future if we introduce storage gateway). This relationship is captured in the ResourceUsages table.
       `Database item => resource: <studyId>, setName: <studyId>, items: [<fs arn>, ...]`
     2. Study to member account to environment using the study. This relationship is also captured in the ResourceUsages table.
        `Database item => resource: <studyId>, setName: <memberAccountId>, items: [<envId>, ...]`
     3. fs role to member account ids. This relationship is captured in the RoleAllocations table.
         `Database item => composite key: <accountId, fs role arn>, trust: [<memberAccountId>, ...]`

**The logic**
- Get the application role entity from the study.appRoleArn
- Using the resource usage service, get all existing usages of fs roles for the study, see item _i_ above.
- For each found fs role arn, load its fs role entity,
  - Does the entity include the member account in its trust and include the study with the same envPermission,
    if so, then we don't need to create an fs role
  - If not, can we fit the member account in the trust policy, if so then update the actual trust document
    in the data source account for the fs role, then update the database, see item _iii_ above.
- If no fs role entity is found, or the existing on has its trust property full, we need to create a new
  fs role as follows:
  - Create the fs role entity and add the member account to the trust property, see item _iii_ above.
    - At this point, we don't store the entity in the data base until after we are able to create
      it in the data source account
  - Assume the app role and create the fs role in the data source account
  - Store the entity in the database
  - Add usage of the study by this role, see item _i_ above.

Checklist: 

* [X] Have you written new tests for your core changes, as applicable?
* [X] Have you successfully ran unit tests and manual tests with your changes locally?

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
@@ -32,7 +32,7 @@ const AuditService = require('@aws-ee/base-services/lib/audit/audit-writer-servi
 const JsonSchemaValidationService = require('@aws-ee/base-services/lib/json-schema-validation-service');
 const ApplicationRoleService = require('../application-role-service');
 const ResourceUsageService = require('../../../../usage/resource-usage-service');
-const FileSystemRoleService = require('../filesystem-role-service');
+const FilesystemRoleService = require('../filesystem-role-service');
 
 const createStudy = ({
   id = 'study-1',
@@ -119,10 +119,10 @@ describe('DataSourceBucketService', () => {
     container.register('auditWriterService', new AuditService());
     container.register('roles-only/applicationRoleService', new ApplicationRoleService());
     container.register('resourceUsageService', new ResourceUsageService());
-    container.register('roles-only/fileSystemRoleService', new FileSystemRoleService());
+    container.register('roles-only/filesystemRoleService', new FilesystemRoleService());
     await container.initServices();
 
-    service = await container.find('roles-only/fileSystemRoleService');
+    service = await container.find('roles-only/filesystemRoleService');
     appRoleService = await container.find('roles-only/applicationRoleService');
     usageService = await container.find('resourceUsageService');
   });
--- a/addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
@@ -1,0 +1,366 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License").
+ *  You may not use this file except in compliance with the License.
+ *  A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+const ServicesContainer = require('@aws-ee/base-services-container/lib/services-container');
+
+// Mocked services
+jest.mock('@aws-ee/base-services/lib/db-service');
+jest.mock('@aws-ee/base-services/lib/logger/logger-service');
+jest.mock('@aws-ee/base-services/lib/settings/env-settings-service');
+jest.mock('@aws-ee/base-services/lib/plugin-registry/plugin-registry-service');
+jest.mock('@aws-ee/base-services/lib/audit/audit-writer-service');
+jest.mock('../application-role-service');
+
+const Aws = require('@aws-ee/base-services/lib/aws/aws-service');
+const Logger = require('@aws-ee/base-services/lib/logger/logger-service');
+const DbService = require('@aws-ee/base-services/lib/db-service');
+const PluginRegistryService = require('@aws-ee/base-services/lib/plugin-registry/plugin-registry-service');
+const SettingsService = require('@aws-ee/base-services/lib/settings/env-settings-service');
+const AuthService = require('@aws-ee/base-services/lib/authorization/authorization-service');
+const AuditService = require('@aws-ee/base-services/lib/audit/audit-writer-service');
+const JsonSchemaValidationService = require('@aws-ee/base-services/lib/json-schema-validation-service');
+const ApplicationRoleService = require('../application-role-service');
+const ResourceUsageService = require('../../../../usage/resource-usage-service');
+const FileSystemRoleService = require('../filesystem-role-service');
+
+const createStudy = ({
+  id = 'study-1',
+  category = 'Organization',
+  accountId = '1122334455',
+  awsPartition = 'aws',
+  bucketAccess = 'roles',
+  bucket = 'bucket-1',
+  qualifier = 'swb-IhsKhN8GsLneiis11ujlb8',
+  appRoleArn = 'arn:aws:iam::684277579687:role/swb-IhsKhN8GsLneiis11ujlb8-app-1607537845811',
+  accessType = 'readwrite',
+  envPermission = { read: true, write: true },
+  folder = '/',
+  kmsScope = 'none',
+} = {}) => ({
+  id,
+  category,
+  accountId,
+  awsPartition,
+  bucketAccess,
+  bucket,
+  qualifier,
+  appRoleArn,
+  accessType,
+  envPermission,
+  folder,
+  kmsScope,
+});
+
+const createAppRole = ({
+  arn = 'arn:aws:iam::684277579687:role/swb-IhsKhN8GsLneiis11ujlb8-app-1607537845811',
+  accountId = '1122334455',
+  mainRegion = 'us-east-1',
+  awsPartition = 'aws',
+  bucket = 'bucket-1',
+  bucketRegion = 'us-east-1',
+  status = 'pending',
+  name = 'swb-IhsKhN8GsLneiis11ujlb8-app-1607537845811',
+  qualifier = 'swb-IhsKhN8GsLneiis11ujlb8',
+  boundaryPolicyArn = 'arn:aws:iam::684277579687:policy/swb-IhsKhN8GsLneiis11ujlb8-app-1607537845811',
+  studies = {
+    'study-1': {
+      accessType: 'readonly',
+      kmsScope: 'none',
+      folder: '/',
+    },
+  },
+} = {}) => ({
+  accountId,
+  arn,
+  name,
+  mainRegion,
+  awsPartition,
+  bucket,
+  bucketRegion,
+  status,
+  qualifier,
+  boundaryPolicyArn,
+  studies,
+});
+
+const createAdminContext = ({ uid = 'uid-admin' } = {}) => ({
+  principalIdentifier: { uid },
+  principal: { isAdmin: true, userRole: 'admin', status: 'active' },
+});
+
+describe('DataSourceBucketService', () => {
+  let container;
+  let service;
+  let appRoleService;
+  let usageService;
+
+  beforeEach(async () => {
+    // Initialize services container and register dependencies
+    container = new ServicesContainer();
+
+    container.register('dbService', new DbService());
+    container.register('jsonSchemaValidationService', new JsonSchemaValidationService());
+    container.register('settings', new SettingsService());
+    container.register('aws', new Aws());
+    container.register('log', new Logger());
+    container.register('pluginRegistryService', new PluginRegistryService());
+    container.register('authorizationService', new AuthService());
+    container.register('auditWriterService', new AuditService());
+    container.register('roles-only/applicationRoleService', new ApplicationRoleService());
+    container.register('resourceUsageService', new ResourceUsageService());
+    container.register('roles-only/fileSystemRoleService', new FileSystemRoleService());
+    await container.initServices();
+
+    service = await container.find('roles-only/fileSystemRoleService');
+    appRoleService = await container.find('roles-only/applicationRoleService');
+    usageService = await container.find('resourceUsageService');
+  });
+
+  describe('when allocating a role', () => {
+    it('skip if study bucket access is not roles', async () => {
+      const uid = 'u-currentUserId';
+      const requestContext = { principalIdentifier: { uid } };
+      const study = { bucketAccess: 'something else' };
+      const env = {};
+      const accountId = '1234456789012';
+
+      usageService.getResourceUsage = jest.fn(() => {
+        throw new Error('I got here');
+      });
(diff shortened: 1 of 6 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -0,0 +1,553 @@
...
+   *
+   * Note: this method creates an IAM role resource in an AWS account, if needed.
+   *
+   * @param requestContext The standard requestContext
+   * @param studyEntity The study entity. IMPORTANT, it is expected that this study entity will have
+   * an additional property named 'envPermission', it is an object with the following shape:
+   * { read: true/false, write: true/false }
+   * @param environmentEntity The environment entity that will use the filesystem role
+   */
+  async allocateRole(requestContext, studyEntity = {}, environmentEntity = {}, memberAccountId) {
```
SanketD92, line 132: Nit: Optional params should come after required params

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -0,0 +1,553 @@
...
+   *
+   * Note: this method might delete the IAM role resource in the data source AWS account, if needed.
+   *
+   * @param requestContext The standard requestContext
+   * @param studyEntity The study entity. IMPORTANT, it is expected that this study entity will have
+   * an additional property named 'envPermission', it is an object with the following shape:
+   * { read: true/false, write: true/false }
+   * @param environmentEntity The environment entity that was using the filesystem role
+   */
+  async deallocateRole(requestContext, fsRoleArn, studyEntity = {}, environmentEntity = {}, memberAccountId) {
```
SanketD92, line 283: Same comment as above

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/helpers/entities/filesystem-role-methods.js
```diff
@@ -0,0 +1,200 @@
...
+  });
+
+  return !_.isUndefined(foundStudy);
+}
+
+function removeStudy(fsRoleEntity = {}, studyEntity = {}) {
+  delete fsRoleEntity.studies[studyEntity.id];
+}
+
+function addMemberAccount(fsRoleEntity = {}, memberAccountId) {
```
SanketD92, line 57: Same comment as above

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
hatimkhan: Let me add memberAccountId = '',   studyeEntity, envrionmentEntity, memberAccountId  all are not optional.  I do the studyEntity = {} trick to help avoid null pointer exceptions while I check the values.  In my @param I don't specify that these are optional.  So, once we produce jsdoc out of the code, you will see that all my jdoc have @param and if they are optional, I will "hopefully" remember to mention that :)

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
```diff
@@ -0,0 +1,366 @@
...
+
+  describe('when allocating a role', () => {
+    it('skip if study bucket access is not roles', async () => {
+      const uid = 'u-currentUserId';
+      const requestContext = { principalIdentifier: { uid } };
+      const study = { bucketAccess: 'something else' };
+      const env = {};
+      const accountId = '1234456789012';
+
+      usageService.getResourceUsage = jest.fn(() => {
```
Bingjiling, line 138: Nit: This can be done by `expect(usageService.getResourceUsage).not.toHaveBeenCalled();`

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -0,0 +1,553 @@
...
+      'roles-only/applicationRoleService',
+      'resourceUsageService',
+    ]);
+
+    // Primer
+    // We capture the relations between four items: fs role, study, member account and environment.
+    // We keep the relationship information in three database items across two tables.
+    // 1. Study to fs roles, this is a one-to-many relationship (but we need to make it many-to-many in the future
+    //    if we introduce storage gateway). This relationship is captured in the ResourceUsages table.
+    //    Database item => resource: <studyId>, setName: <studyId>, items: [<fs arn>, ...]
```
Bingjiling, line 157: I'm a bit confused by the key structure of ResourceUsages table. Is <studyId> the primary key? Or <studyId, setName> as composite key?

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -0,0 +1,553 @@
...
+      setName: fsUsageSetName,
+    });
+    const fsRoleEntities = _.get(fsRoleUsages, fsUsageSetName, []);
+
+    // Start a loop to see if existing fs roles for the study have the member account already
+    // or can take it without exceeding the trust doc limit?
+    let fsRoleEntity;
+
+    // eslint-disable-next-line no-restricted-syntax
+    for (const fsRoleArn of fsRoleEntities) {
```
Bingjiling, line 196: Out of curiosity, why is the for loop written this way? We can do a `Promise.all` on the first await call to find entities for all fsRole, then use `filter` to find if there's an existing fs role.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
hatimkhan: Oops, sorry, this was not meant to be here. I did that while I was doing some debugging and forget to remove it.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -0,0 +1,553 @@
...
+        // Add environment usage
+        await resourceUsageService.addUsage(requestContext, {
+          resource: studyResourceId,
+          setName: memberAccountId,
+          item: envId,
+        });
+
+        // We need to save the fsRoleEntity (even though it is already in the database), because
+        // we added the member account id to the trust doc.
+        fsRoleEntity = await this.saveEntity(requestContext, entity);
```
Bingjiling, line 234: If this save fails, can the method be retired successfully to save again?

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
hatimkhan: It is a composite key (resource, setName).  I wish I can attached a diagram here.  For a description of the tables, you can take a look at the dev quip doc that we have.

The comments in the file system role service are not meant to introduce the resource usage service.  The resource usage service was introduced in a previous PR.  Hopefully the documentation for that service is good enough.  Let me know.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
hatimkhan: The logic itself does not lend itself as a filter logic as it contains mutation, multiple decisions and side effects, such as calling mutating AWS IAM APIs.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
hatimkhan: I agonized over this when I was writing the code :)
Introducing a transactional behavior to this method means a much much more complex approach, because we are doing mutating calls using AWS IAM roles, it means that we need to consider the command pattern with undo logic.  For now, I went with this approach while acknowledging its shortcomings.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -0,0 +1,553 @@
...
+
+    // Create in the fs role in the data source account
+    await this.provisionRole(fsRoleEntity);
+
+    // Add environment usage
+    await resourceUsageService.addUsage(requestContext, {
+      resource: studyResourceId,
+      setName: memberAccountId,
+      item: envId,
+    });
```
jn1119, line 255: nit: Does adding environment usage really belong here given that this method and file is just about file system roles.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -0,0 +1,553 @@
...
+      setName: fsUsageSetName,
+    });
+    const fsRoleEntities = _.get(fsRoleUsages, fsUsageSetName, []);
+
+    // Start a loop to see if existing fs roles for the study have the member account already
+    // or can take it without exceeding the trust doc limit?
+    let fsRoleEntity;
+
+    // eslint-disable-next-line no-restricted-syntax
+    for (const fsRoleArn of fsRoleEntities) {
```
jn1119, line 196: Do we need a lock on fsRoleArn or at top level on appRoleArn given that there can be concurrent requests for using the same fsRoleArn or appRoleArn?

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
hatimkhan: This is actually core to the filesystem **allocateRole** logic.  The logic is a 'pool manager' logic (checkin/checkout), it needs to keep track of what roles were issued and do a clean up after that.  If we move this logic outside of the filesystem role service, we need another service called 'filesystem role pool service'.   When I thought about doing that, I realized that we can combine them and there is no need for an additional service.  Let me know what you think.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
hatimkhan: I have it in the environment resource service where I lock at the application role level.  This code is coming soon in a PR.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -0,0 +1,553 @@
...
+        await resourceUsageService.addUsage(requestContext, {
+          resource: studyResourceId,
+          setName: memberAccountId,
+          item: envId,
+        });
+
+        break;
+      }
+
+      if (!maxReached(addMemberAccount(entity, memberAccountId))) {
```
jn1119, line 219: Does this take care of duplicate member account additions? Basically we shouldn't add duplicates.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -0,0 +1,553 @@
...
+        // We found an existing fs role entity and we can fit the member account in its trust doc
+
+        // Update the fs role policy in the data source account
+        await this.updateAssumeRolePolicy(entity);
```
jn1119, line 223: Does this take care of duplicate member account additions? Basically we shouldn't add duplicates.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
```diff
@@ -0,0 +1,363 @@
...
+    container.register('resourceUsageService', new ResourceUsageService());
+    container.register('roles-only/filesystemRoleService', new FilesystemRoleService());
+    await container.initServices();
+
+    service = await container.find('roles-only/filesystemRoleService');
+    appRoleService = await container.find('roles-only/applicationRoleService');
+    usageService = await container.find('resourceUsageService');
+  });
+
+  describe('when allocating a role', () => {
```
jn1119, line 130: Can we add a few test cases are following cases:

1. We found the appropriate filesystem role but couldn't add study, env, account via resourceUsageService because of network failure or something. Calling the allocateRole method should work successfully without side effects.

2. The above case when we didn't find a filesystem role and created a new one. But couldn't add the (study, env, account) via resourceUsageService

3. The account doesn't exist in fileSystemRole, so we update the policy but method fails while saving the entity so we can't call resourceUsageService as well. A retry on the method should ideally work.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
hatimkhan: Yes. The 'addMemberAccount' does  a check on the array before adding the member.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
hatimkhan: Yes. The entity has an array named 'trust' and since use the 'addMemberAccount' to add accounts, the 'addMemberAccount' won't add an account to the 'trust' array if it is already there.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
hatimkhan: Great suggestions. I am going to write more unit tests and document the different error scenarios that can happen and how to recover from them.  I am thinking of doing that once we have the whole solution in a working state.

SanketD92 (approved):
LGTM

PR #238, opened 2020-12-14
feat: install R3.6 and system packages required for research

Ludo171:
**Motivation**
We're currently trying to reproduce a biomedical analyses using RStudio workspaces.
We faced the following blockers when trying to install R libraries :
- the workspace misses system packages required by R to install packages (like 'devtools' or 'tidyverse')
- multiple broadly used R libraries require R>=3.5, whereas only R3.4 was installed through amazon-linux-extras repositories.
For security considerations, users do not have root access to the workspaces, so system packages have to be preinstalled in the AMI of the workspace.

**Description of changes:**
The changes here consist of modifying the provisioning script used to build RStudio AMIs : `main\solution\machine-images\config\infra\provisioners\provision-rstudio.sh`.
The following modifications make them meet the requirements of most analysis using R : 
- Install R3.6.3 from source (version can be changed easily by changing a variable in the script)
- Install system packages necessary to install 'tidyverse' and 'devtools' which are commonly used (deduced from debugging)
- Install additional system packages recommended by [Rstudio](https://docs.rstudio.com/rsc/post-setup-tool)

**Checklist:** 

### DIFF
--- a/main/solution/machine-images/config/infra/provisioners/provision-rstudio.sh
+++ b/main/solution/machine-images/config/infra/provisioners/provision-rstudio.sh
@@ -1,19 +1,33 @@
 #!/usr/bin/env bash
 
+# Various development packages needed to compile R
+sudo yum install -y gcc gcc-gfortran gcc-c++
+sudo yum install -y java-1.8.0-openjdk-devel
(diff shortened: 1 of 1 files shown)


SanketD92 (approved):
Looks good.

PR #241, opened 2020-12-15
feat: bring your own bucket - workflow changes and permission propagation

hatimkhan:
This PR includes two parts:

1. BYOB implementation needed when a workspace is provisioned and terminated
2. When a study permission is changed, propagating the changes to all impacted environments

Most of the code implementing item 1 can be found in the roles-only/environmentResourceService.  While most of the code implementing item 2 can be found in the studyOperationService.

Checklist: 

* [X] Have you linted your code locally prior to submission?
* [X] Have you written new tests for your core changes, as applicable?
* [X] Have you successfully ran unit tests and manual tests with your changes locally?


addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/environment-resource-service.test.js
```diff
@@ -0,0 +1,159 @@
...
+    container.register('environmentScService', new EnvironmentScService());
+    container.register('resourceUsageService', new ResourceUsageService());
+    container.register('roles-only/applicationRoleService', new ApplicationRoleService());
+    container.register('roles-only/filesystemRoleService', new FilesystemRoleService());
+    container.register('roles-only/environmentResourceService', new EnvironmentResourceService());
+    await container.initServices();
+
+    service = await container.find('roles-only/environmentResourceService');
+  });
+
```
Bingjiling, line 110: There's no test for allocate or deallocate?

addons/addon-base-raas/packages/base-raas-services/lib/environment/__tests__/environment-mount-service.test.js
```diff
@@ -45,9 +45,9 @@ describe('EnvironmentMountService', () => {
   let service = null;
   let environmentScService = null;
   let iamService = null;
-  let studyService = null;
-  let studyPermissionService = null;
-  let aws = null;
+  // let studyService = null;
```
Bingjiling, line 48: Nit: Remove the services if they're not needed anymore?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-service.js
```diff
@@ -487,6 +491,141 @@ class EnvironmentScService extends Service {
...
 
+  /**
+   * Updates the study ids for the environment sc entity. IMPORTANT: do NOT use this from an api handler, this method
+   * does NOT validate permissions to the study ids provided. If you need to update the study for an environment with
+   * permissions enforced, then use the studyOperationService.updatePermissions() method.
+   *
+   * @param requestContext The standard request context
+   * @param studyIds An array of study ids
+   */
+  async updateStudyIds(requestContext, id, studyIds = []) {
```
Bingjiling, line 502: This method was not called anywhere? Also, why do we need a separate method to update studyIds specifically instead of using the generic update method that already exist?

addons/addon-base-raas/packages/base-raas-services/lib/environment/__tests__/environment-mount-service.test.js
hatimkhan: Done

addons/addon-base-raas/packages/base-raas-services/lib/study/study-operation-service.js
```diff
@@ -0,0 +1,286 @@
...
+      'auditWriterService',
+      'lockService',
+      'studyService',
+      'environmentScService',
+      'pluginRegistryService',
+    ]);
+  }
+
+  /**
+   * This method not only updates the study/study permissions entities, but it also understands how to propagate the
```
Bingjiling, line 54: Nice comment!

addons/addon-base-raas/packages/base-raas-services/lib/study/study-operation-service.js
```diff
@@ -0,0 +1,286 @@
...
+    const [studyService, lockService, environmentScService] = await this.service([
+      'studyService',
+      'lockService',
+      'environmentScService',
+    ]);
+
+    // First, we need to have a lock at the study operation level
+    const result = await lockService.tryWriteLockAndRun({ id: `study-${studyId}-operation` }, async () => {
+      // Currently, propagating permission is not done via a workflow, so we have to do everything in less than
+      // 30 seconds. In an attempt to fail fast, an exception is thrown if we are trying to update more than
```
Bingjiling, line 82: Do we need a special note or spinning on the UI so user don't see it as a bug when it's taking a lot of time?

hatimkhan: The existing UI for updating study permissions has the spinner implemented, so we are good here.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-service.js
hatimkhan: Removed.
I used to call it from the study registration service, I then commented out the code in the study registration service and thought I would leave it for future use.  The main reason for not having it done in the update method is that it meant that we needed to updated the 'update schema' and allow the 'studyIds' to be updated. This meant that we needed to add validation logic that checks the permission for each study id against the creator of the environment not the request context, basically it was going to complicate the update method.  Anyway, I think removing it for now is a good thing.  Done.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/environment-resource-service.test.js
hatimkhan: Added, thanks.

SanketD92 (dismissed):
LGTM

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
```diff
@@ -0,0 +1,287 @@
...
+            fsRoleArn,
+            study,
+            environmentScEntity,
+            memberAccountId,
+          );
+
+          delete studyRoles[study.id];
+        }
+      });
+    };
```
jn1119, line 167: nit: Is it worth moving this code into a shared function? I feel its fine as it is but wondering if you thought about it?

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
```diff
@@ -0,0 +1,287 @@
...
+   * the standard StudyEntity, however, there is one additional attributes added to each of the StudyEntity. This
+   * additional attribute is called 'envPermission', it is an object with the following shape:
+   * { read: true/false, write: true/false }
+   * @param environmentScEntity the EnvironmentSc entity. The entity is expected to have an attributed named 'studyRoles'.
+   * The studyRoles is a map where the key is the study id and the value is the fs role arn
+   *
+   */
+  async provideEnvRolePolicy(requestContext, { policyDoc, studies: allStudies, environmentScEntity }) {
+    // We are only interested in studies with bucket access = 'roles'
+    const studies = _.filter(allStudies, study => study.bucketAccess === 'roles');
```
jn1119, line 193: Shouldn't this kind of filter be added in legacy version of environment-resource-service

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
```diff
@@ -0,0 +1,287 @@
...
+      s3Mounts.push(item);
+    });
+
+    return s3Mounts;
+  }
+
+  // @private
+  // Returns a map. The key is the bucket name and the value is the bucket kmsArn. IMPORTANT: not all studies buckets
+  // will be included in the map. Only Buckets for studies with kmsScope = bucket are included.
+  async getBuckets(studies) {
```
jn1119, line 248: Where is this method used? Also perhaps rename it to getKMSScopedBuckets or something

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-operation-service.test.js
```diff
@@ -0,0 +1,262 @@
...
+const StudyService = require('../study-service');
+const StudyPermissionService = require('../study-permission-service');
+const StudyOperationService = require('../study-operation-service');
+
+const createResearcherContext = ({ uid = 'uid-researcher-1' } = {}) => ({
+  principalIdentifier: { uid },
+  principal: { userRole: 'researcher', status: 'active' },
+});
+
+describe('StudyOperationService', () => {
```
jn1119, line 52: Can you add a test for the case when updating an environment fails but we go through other environments successfully.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
hatimkhan: The existing studies and studyService.create() method don't populate this property.   This property is only populated when you register a **bucket** and then you register a study in that bucket.  When you register a **bucket**, the API asks for the access strategy to use when accessing studies in the bucket.   Then, when you register the study (you specify the account and bucket), we  populate the study.bucketAccess value based on the bucket  access strategy.  The legacy studies don't have that.

The best why to tell if a study is "legacy" study is by checking if study.resources is empty/undefined.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
hatimkhan: Sorry, it was used by code that is no longer in the file.  I removed the method just now.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
hatimkhan: Didn't really think about putting that in a function. Thinking about it now, I feel it is okay the way it is for now.  Maybe something I can address if we add more logic to this exact loop.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
jn1119: Got it thanks. Basically study.resouces is not empty.

SanketD92 (approved):

jn1119 (approved):

PR #243, opened 2020-12-15
feat: Added code for add/remove permissions for internal studies which use bucket policy

jn1119:
Issue #, if available: GALI-615

Description of changes:
* Added the legacy methods for adding and removing S3 bucket and KMS policies
* Tested by bringing two workspaces and terminating them. Verified that workspaces could access the study appropriately and bucket and KMS policies were correctly reflected later when the workspaces were terminated.

Checklist: 

* [] If you had to run manual tests, have you considered automating those tests by adding them to [end-to-end tests](../main/end-to-end-tests/README.md)?

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/data-source/__tests__/environment-resource-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/data-source/__tests__/environment-resource-service.test.js
@@ -1,0 +1,666 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License").
+ *  You may not use this file except in compliance with the License.
+ *  A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+const ServicesContainer = require('@aws-ee/base-services-container/lib/services-container');
+
+// Mocked services
+jest.mock('@aws-ee/base-services/lib/audit/audit-writer-service');
+jest.mock('@aws-ee/base-services/lib/json-schema-validation-service');
+jest.mock('@aws-ee/base-services/lib/lock/lock-service');
+jest.mock('@aws-ee/base-services/lib/db-service');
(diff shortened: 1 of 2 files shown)


SanketD92 (dismissed):
LGTM

addons/addon-base-raas/packages/base-raas-services/lib/data-source/__tests__/environment-resource-service.test.js
```diff
@@ -0,0 +1,645 @@
...
+      });
+
+      AWSMock.mock('S3', 'putBucketPolicy', (params, callback) => {
+        expect(params).toMatchObject({
+          Bucket: 'study-bucket',
+          Policy: JSON.stringify(s3Policy),
+        });
+        callback(null, {});
+      });
+      await environmentResourceService.addToBucketPolicy({}, studies, 'accountId1');
```
hatimkhan, line 207: You might need to add a line at the end using expect.assertions(number) https://jestjs.io/docs/en/expect#expectassertionsnumber  otherwise if the putBucketPolicy() and getBuckPolicy() are never called, the test will still succeed.

jn1119: Thanks for bringing this up. I have added the assertions for putBucketPolicy. For getBucketPolicy it was implicit since we use the return value later.

SanketD92 (approved):

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/legacy/environment-resource-service.js
```diff
@@ -23,6 +23,42 @@ const { isOpenData } = require('../../../study/helpers/entities/study-methods');
 const settingKeys = {
   studyDataBucketName: 'studyDataBucketName',
   studyDataKmsKeyArn: 'studyDataKmsKeyArn',
+  studyDataKmsPolicyWorkspaceSid: 'studyDataKmsPolicyWorkspaceSid',
+};
+
+const listStatementParamsFn = (bucket, prefix) => {
+  return {
+    statementId: `List:${prefix}`,
```
hatimkhan, line 31: Can you check what characters are allowed to appear in the statementId?  This is because the prefix of an S3 object might also contain characters that the statement id might not allows.

jn1119: It seems like the statementId if a free-form field. I tried using all the special characters including the characters that should be avoided mentioned on S3 documentation[1] and was still able to save the policy.

https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html

PR #245, opened 2020-12-17
feat: studies list page update

Bingjiling:
Issue #, if available:

Description of changes:

* Show status for studies
* Hide readwrite permission for readonly study
* Disable selection and permission update for pending and unavailable studies 

Existing internal study and ReadOnly external study in available status: 
Screen Shot 2020-12-17 at 3 02 13 PM

External ReadWirte study in pending status:
Screen Shot 2020-12-17 at 3 01 46 PM

External ReadWrite study in error status:
Screen Shot 2020-12-17 at 3 01 32 PM

Checklist: 


SanketD92 (changes requested):
The base branch needs updating on your topic branch.

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/studies/StudyPermissionsTable.js
```diff
@@ -103,7 +107,8 @@ class StudyPermissionsTable extends React.Component {
 
   renderTable() {
     const studyPermissions = this.permissionsStore.studyPermissions;
-    const isEditable = studyPermissions.adminUsers.some(uid => uid === this.currUser.uid);
+    const isEditable =
+      studyPermissions.adminUsers.some(uid => uid === this.currUser.uid) && this.study.state.canChangePermission;
```
hatimkhan, line 111: Is it possible that you move this existing logic as a view on the permissions entity model?

something like  isAdmin(uid)  as a view on the permissions entity model?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/studies/Study.js
```diff
@@ -56,7 +92,7 @@ const Study = types
 
     getPermissionsStore() {
       if (!self.permissionsStore) {
-        self.permissionsStore = StudyPermissionsStore.create({ studyId: self.id });
+        self.permissionsStore = StudyPermissionsStore.create({ studyId: self.id, studyAccessType: self.accessType });
```
hatimkhan, line 95: Is it possible not to introduce the property 'studyAccessType' at the permissions entity level?  can you make it a derived property, instead?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/studies/StudyPermissionsTable.js
Bingjiling: Emmm, view methods can not have pass in parameter

Bingjiling: Ah, my bad. I thought view method can only be `get` something, which can't have pass in parameter. Will update it.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/studies/Study.js
```diff
@@ -15,10 +15,44 @@
 
 import { types, applySnapshot } from 'mobx-state-tree';
 
+import _ from 'lodash';
 import { StudyFilesStore } from './StudyFilesStore';
 import { StudyPermissionsStore } from './StudyPermissionsStore';
 import { categories } from './categories';
 
+// 'pending', 'error', 'default'
+// Note: When study is available, the status column is empty
```
hatimkhan, line 24: Is that true?  That is only in the database, but when we convert the dbEntity to a studyEntity in the backend we populate the status as 'reachable' if there is no status in the database.  Can you confirm? thanks

hatimkhan (approved):
return _.map(result, toStudyEntity);

SanketD92 (approved):
LGTM

PR #246, opened 2020-12-18
feat: bring your own bucket - Data Sources listing UI

hatimkhan:
This PR introduces a new page that is only available to admins.  The page lists registered accounts including the registered studies.  The admin can check the connection at the account level or at the study level.  In addition, the admin can inspect the detail of a study.

A few screenshots:
data-sources-1
data-sources-2
data-sources-3

Checklist: 


addons/addon-base-raas-ui/packages/base-raas-ui/src/models/data-sources/DataSourceAccount.js
```diff
@@ -0,0 +1,135 @@
...
+        if (value === 'studies') return; // we don't want to update the studies
+        self[key] = value;
+      });
+
+      // We want to take care of thee statusMsg because it might come as undefined
+      if (_.isUndefined(raw.statusMsg)) self.statusMsg = '';
+    },
+
+    setStudies(studies) {
+      consolidateToMap(self.studies, studies, (exiting, newItem) => {
```
Bingjiling, line 79: Nit: existing?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/data-sources/DataSourceAccountsStore.js
```diff
@@ -0,0 +1,120 @@
...
+        if (!entry) {
+          // Lazily create the store
+          self.accountStores.set(accountId, DataSourceAccountStore.create({ accountId }));
+          entry = self.accountStores.get(accountId);
+        }
+
+        return entry;
+      },
+
+      getStudyStore({ accountId, studyId }) {
```
Bingjiling, line 72: Why do we want a convenient method here to get StudyStore directly? In an idea world, component trying to use study store should already has account store right?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/data-sources/DataSourceStudy.js
```diff
@@ -0,0 +1,137 @@
...
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+import _ from 'lodash';
+import { types } from 'mobx-state-tree';
+
+const states = {
+  pending: {
```
Bingjiling, line 20: Is it a good idea define the states of DataSourceStudy and DataSourceAccount once and let the two entity share it?

hatimkhan: They might look like the same now.  But, I was thinking in terms of managed/unmanaged member accounts in the future.  The account reachability states might differ between studies and accounts.  For example, an account might be partially configured (this could happen when some app roles are working and some are not).  I felt it is save to keep them separate.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/data-sources/DataSourceAccount.js
hatimkhan: fixed

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/data-sources/DataSourceAccountsStore.js
hatimkhan: It depends on if the store is a root store.  Usually, a react component might only need an MST entity and it gets the store it needs from a root store.  A root store is any store that is accessible from the appContext.  This way we don't have to pass stores everywhere.  However,  I do see some pros and cons doing this.  Anyway,  I have a screenshot to explain what root stores are.

At the end, I decided to remove this method and pass the store to the React component per your suggestion :)

RaaS

SanketD92 (approved):
LGTM

addons/addon-base-raas/packages/base-raas-services/lib/data-source/data-source-reachability-service.js
```diff
@@ -110,20 +111,24 @@ class DataSourceReachabilityService extends Service {
...
+    });
 
     // Write audit event
     await this.audit(requestContext, {
       action: 'check-dsAccount-reachability',
       body: { id, type },
     });
 
-    return outputVal;
+    return entity;
```
SanketD92, line 131: Lets also update the expected response (from status to entity object) in the Quip doc.

SanketD92 (approved):

PR #247, opened 2020-12-22
Feat byob error handling unit tests filesystem

SanketD92:
Issue #, if available:
GALI-637

Description of changes:
Retries to allocateRole and deallocateRole methods work without issues, and are not short-circuited due to potential previous incomplete transactions.

Checklist: 


addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
```diff
@@ -235,6 +235,106 @@ describe('DataSourceBucketService', () => {
...
+            accessType: 'readwrite',
+            envPermission: { read: true, write: true },
+            folder: '/',
+            kmsArn: undefined,
+            kmsScope: 'none',
+          },
+        },
+        trust: ['3333333'],
+      };
+      const initialValue = JSON.parse(JSON.stringify(fsRoleEntity)); // Deep copy
```
hatimkhan, line 259: Can you use _.cloneDeep() instead?

hatimkhan:
I don't see any new logic added to deprovisionRole() method.  Inside that method, you might want to do a try and catch, then see if AWS returns an error code that says that the role does not exist, and if that is the case, then you don't throw the exception, otherwise you throw the exception.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
```diff
@@ -235,6 +235,106 @@ describe('DataSourceBucketService', () => {
...
+
+      // Attempt #2 (Successful)
+      service.updateAssumeRolePolicy = jest.fn();
+      service.find = jest.fn().mockResolvedValue(initialValue);
+      await service.allocateRole(requestContext, study, env, memberAcct);
+
+      // CHECK
+      expect(service.provisionRole).not.toHaveBeenCalled();
+      // Called only during successful attempt
+      expect(usageService.addUsage).toHaveBeenCalledTimes(1);
```
jn1119, line 293: Shouldn't this be called twice. One for environment usage and one for file system role usage

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -378,10 +371,8 @@ class FilesystemRoleService extends Service {
       return;
     }
 
-    if (usage.removed) {
-      // Delete the role from the data source account
-      await this.deprovisionRole(fsRoleEntity);
-    }
+    // Delete the role from the data source account
+    await this.deprovisionRole(fsRoleEntity);
```
jn1119, line 375: What is the behavior of calling this method if the policy or the role is already deleted? If it throws an error, we need to fix it.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
```diff
@@ -323,20 +323,14 @@ class FilesystemRoleService extends Service {
     // If it is not found then we have done the deallocation logic already
     if (_.isUndefined(fsRoleEntity)) return;
 
-    // If we don't have the study in the fs role entity then we have done the deallocation logic already
-    if (!hasStudy(fsRoleEntity, studyEntity)) return;
-
-    // Do we have the member account listed in the trust document?
-    if (!hasMemberAccount(fsRoleEntity, memberAccountId)) return;
-
```
jn1119, line 331: Add a test for this branch? This would including creating the fsRoleEntity such that it doesn't have study and member account associated.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
SanketD92: No, because our test provides the fsRoleEntity object with the matching arn value, but with the missing account in trust list. Therefore the logic returns the `fsRoleEntity` object on line 240 if it is not undefined.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/filesystem-role-service.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
```diff
@@ -359,5 +466,170 @@ describe('DataSourceBucketService', () => {
...
+      const env = { id: 'env-1' };
+      const memberAcct = '1234456789012';
+      const arn = 'fs-role-arn';
+      const fsRoleEntity = {
+        arn,
+        studies: {},
+        trust: [],
+      };
+      service.updateAssumeRolePolicy = jest.fn();
+      service.deprovisionRole = jest.fn();
```
jn1119, line 609: is `deprovisionRole` being unit tested anywhere?

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
jn1119: Got it, thanks

jn1119: I think there is already a test covering the new file system role scenario. So this can be resolved.

addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/__tests__/filesystem-role-service.test.js
SanketD92: Doesn't look like it, I'll add it.

jn1119 (approved):

PR #253, opened 2020-12-30
Feat: Legacy env-resource-service tests

SanketD92:
Issue #, if available:
GALI-638

Description of changes:
The change in implementation takes into consideration that updates to the bucket policy and KMS policy could have failed previously and therefore API calls to update them should succeed irrelevant of the resource usage state suggested by the DB.

Unit tests have been added to confirm the same.

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/legacy/environment-resource-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/legacy/environment-resource-service.js
@@ -111,7 +111,7 @@ class EnvironmentResourceService extends Service {
           item: environmentScEntity.id,
         });
 
-        if (usage.added && _.size(usage.items) === 1) studiesToAdd.push(study);
+        if (_.size(usage.items) >= 1) studiesToAdd.push(study);
       };
 
       // We do the usage tracking calls, 20 at a time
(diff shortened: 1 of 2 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/legacy/environment-resource-service.js
```diff
@@ -111,7 +111,7 @@ class EnvironmentResourceService extends Service {
           item: environmentScEntity.id,
         });
 
-        if (usage.added && _.size(usage.items) === 1) studiesToAdd.push(study);
```
Bingjiling, line 114: If we remove `usage.added` and `_.size(usage.items) === 1` from the if statement from here, it need to be checked in `addToBucketPolicy` that the policy was not already added, or else we'll run into the situation where the same policy was added multiple times, right? Same goes for other places where if statement was updated, check need to be added to subsequent methods it calls to make sure we don't add things that already exist or delete things that are still being used.

SanketD92: Both addToBucketPolicy and removeFromBucketPolicy generate statements from scratch, and both rely on updateS3BucketPolicy (all 3 are idempotent on their own). This ensures multiple calls with the same params during environment provisioning/termination proceed with similar AWS service calls each time. This is therefore protecting against erroneous AWS service responses, and providing opportunity for retries.

Bingjiling: Got it! That makes sense. Can we remove the if statement altogether then? Since we just added a usage, `_.size(usage.items) >= 1` should always be true?

Bingjiling (approved):
Ship it!

PR #255, opened 2021-01-08
Fixed cross-region S3 bucket access for windows

jn1119:
Issue #, if available: GALI-645

Description of changes:

If workspace and S3 bucket were in different regions, we would see the files being listed in windows but it won't have any contents. After reading of the code, it seems like we were using the correct region centric AWS session only for file listing and not for file copying, writing or deleting. In this change I have moved the logic of finding the correct region upstream and based on that create the correct S3 session.

I didn't find a good way to unit test this change, so I will be extending our old backlog item to improve the unit testing or better add this as part of integ testing.

Checklist: 


SanketD92 (approved):
Looks good.

PR #257, opened 2021-01-09
Feat byob fixes

SanketD92:
Issue #, if available:
1. Once CfN is updated for an already available DS Account in order to add a new study, SWB should check reachability of all pending studies under this account. When the Data Source Study is registered, nothing is done currently to update its status until a user manually tests connection for the study. Current workaround is to hit “Test Connection“ on the new study.
2. When any step reports an error while registering a DS account/study the UI still permits the user to go forward with CfN stack update/create.

Description of changes:
1. Adding reachability check to existing lambda for checking and updating status of unreachable DS studies.
2. Removing Next button and adding Cancel button when any DS registration step displays an error.

Checklist: 

### DIFF
--- a/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/data-sources/register/SubmitStep.js
+++ b/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/data-sources/register/SubmitStep.js
@@ -18,7 +18,7 @@ import React from 'react';
 import { decorate, computed, action } from 'mobx';
 import { observer, inject } from 'mobx-react';
 import { withRouter } from 'react-router-dom';
-import { Segment, Button, Icon, Header, Progress } from 'semantic-ui-react';
+import { Segment, Button, Icon, Header, Progress, Message } from 'semantic-ui-react';
 
 import { swallowError } from '@aws-ee/base-ui/dist/helpers/utils';
 import { gotoFn } from '@aws-ee/base-ui/dist/helpers/routing';
@@ -107,11 +107,25 @@ class SubmitStep extends React.Component {
             </Header>
           )}
         </div>
+
         <div>
           {_.map(operations.ops, op => this.renderOperation(op))}
           {this.renderButtons()}
         </div>
       </div>
+    );
+  }
+
+  renderFailedStepsWarning() {
+    return (
+      <Message warning>
+        <Message.Header>Failures have occurred</Message.Header>
+        <p>It seems that one or more steps have failed while registration. Please fix the errors and retry.</p>
+        <p>
+          If you wish to proceed anyway with creating/updating the CloudFormation stack, resources corresponding to the
+          failed steps might not be reflected in the CloudFormation template.
+        </p>
+      </Message>
     );
   }
 
@@ -160,12 +174,13 @@ class SubmitStep extends React.Component {
     const success = this.success;
     const failure = this.failure;
     const allFailed = this.allFailed;
-    const showNext = !allFailed && success;
+    const showNext = !allFailed && (failure || success);
     const showRetry = allFailed || failure;
     const showCancel = showRetry;
 
     return (
       <div className="mt3">
+        {this.failure && this.renderFailedStepsWarning()}
         {showNext && (
           <Button
             floated="right"
--- a/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/data-sources/register/SubmitStep.js
+++ b/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/data-sources/register/SubmitStep.js
@@ -160,9 +160,9 @@ class SubmitStep extends React.Component {
     const success = this.success;
     const failure = this.failure;
     const allFailed = this.allFailed;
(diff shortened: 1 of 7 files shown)


main/solution/backend/src/lambdas/data-source-reachability/handler.js
```diff
@@ -25,9 +27,32 @@ const handler = async () => {
...
+  const studyService = await container.find('studyService');
+  const dataSourceAccountService = await container.find('dataSourceAccountService');
   const userContext = getSystemRequestContext();
+
+  // Check reachability of all Data Source Accounts ONLY in unreachable states
+  // This kicks off reachability check workflows for all its associated DS studies only if the DS Account status changes
   await dataSourceReachabilityService.attemptReach(userContext, { id: '*', status: 'pending' });
   await dataSourceReachabilityService.attemptReach(userContext, { id: '*', status: 'error' });
+
+  // Check reachability of all Data Source Studies ONLY in unreachable states
```
jn1119, line 39: If an unreachable study is becoming reachable, the opposite might also be true, right? I guess the status would eventually refresh?

main/solution/backend/src/lambdas/data-source-reachability/handler.js
```diff
@@ -25,9 +27,32 @@ const handler = async () => {
...
+      const studiesToCheck = _.filter(studies, study => study.status && _.includes(['pending', 'error'], study.status));
+
+      const processor = async study => {
+        await dataSourceReachabilityService.attemptReach(userContext, { id: study.id, type: 'study' });
+      };
+
+      // Reach out 10 at a time
+      await processInBatches(studiesToCheck, 10, processor);
+    }),
+  );
```
jn1119, line 55: Can we add a unit test for this branch?

main/solution/backend/src/lambdas/data-source-reachability/handler.js
SanketD92: When a reachable study becomes unreachable, it's status in the DB won't change until the next force check (occurs once every day via `data-source-reachability-daily` lambda).

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/data-sources/register/SubmitStep.js
```diff
@@ -160,9 +160,9 @@ class SubmitStep extends React.Component {
     const success = this.success;
     const failure = this.failure;
     const allFailed = this.allFailed;
-    const showNext = !allFailed && (failure || success);
+    const showNext = !allFailed && success;
```
hatimkhan, line 163: @SanketD92  per our discussion, you might want to undo this

main/solution/backend/src/lambdas/data-source-reachability/handler.js
jn1119: I think one good to have feature would be to add some metadata in UI which shows when was the last check performed regardless of its status. Seems misleading to call something reachable if it isn't reachable and refresh it after a day. Probably something to add in some follow up ticket?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/data-sources/register/SubmitStep.js
SanketD92: A quick recap of the discussion:

When any (but not all) of the steps during the registration have failed, there could be additional resources in the CloudFormation stack which still could be safely updated in the DS Account (for eg. a new study could be overlapping, but the parent bucket could be a newly registered valid resource). For this reason, we need to still show the Next button on the UI.

jn1119: The ultimate goal is to have a workable study which won't really happen until user takes the action of correcting whatever is it that went wrong. Updating the CloudFormation stack here doesn't really add much value since it needs to be corrected later anyways. I don't mind what we do either way but I thought it was important to call this bit out. @Sanket I will leave it on you to decide.

SanketD92: The effort to compute all steps have either failed or been skipped is a bit too much just to disable the Next button. Rolling back changes to show the Next button and adding a warning message to alert the admin that whatever has failed would not make it to the CFN template if they wished to proceed anyway.
Screen Shot 2021-01-11 at 1 17 43 PM

main/solution/backend/src/lambdas/data-source-reachability/handler.js
SanketD92: This already exists and the user can view this when expanding the study card.

main/solution/backend/src/lambdas/data-source-reachability/handler.js
SanketD92: Yup, on it.

SanketD92: Done

jn1119 (approved):

PR #260, opened 2021-01-13
refactor: Remove unused authZ provider logic

Bingjiling:
Issue #, if available:

Description of changes:

* Remove unused authZ provider logic
* Deployed and tested login, add user logic were not affected

Checklist: 

### DIFF
--- a/addons/addon-base-rest-api/packages/base-controllers/lib/authentication-provider-controller.js
+++ b/addons/addon-base-rest-api/packages/base-controllers/lib/authentication-provider-controller.js
@@ -106,16 +106,6 @@ async function configure(context) {
   );
 
   // ===============================================================
-  //  POST /configs (mounted to /api/authentication/provider)
-  // ===============================================================
-  router.post(
-    '/configs',
-    wrap(async (req, res) => {
-      await saveAuthenticationProvider(res, req, authProviderConstants.provisioningAction.create);
-    }),
-  );
-
-  // ===============================================================
   //  PUT /configs (mounted to /api/authentication/provider)
   // ===============================================================
   router.put(


addons/addon-base-ui/packages/base-ui/src/models/authentication/AuthenticationProviderConfigsStore.js
```diff
@@ -74,22 +68,6 @@ const AuthenticationProviderConfigsStore = BaseStore.named('AuthenticationProvid
       });
     },
 
-    getUpdateAuthenticationProviderConfigEditor(authenticationProviderConfigId) {
```
SanketD92, line 77: This method is being called by `addons/addon-base-ui/packages/base-ui/src/parts/authentication-providers/EditAuthenticationProvider.js` (line 95). Do we not need changes there?

addons/addon-base-ui/packages/base-ui/src/models/authentication/AuthenticationProviderConfigsStore.js
```diff
@@ -114,15 +92,6 @@ const AuthenticationProviderConfigsStore = BaseStore.named('AuthenticationProvid
       return undefined;
     },
 
-    async updateAuthenticationProvider(authenticationProviderConfig) {
```
SanketD92, line 117: Same comment as above.

addons/addon-base-ui/packages/base-ui/src/models/authentication/AuthenticationProviderConfigsStore.js
Bingjiling: Good catch! Will update.

Bingjiling:
Hi @SanketD92, thanks for the comments earlier. I did another review of the front end functionality and realized the auth provider update method is actually in use by the front end, and it does update the DDB for the auth provider information. However, the update will only stay temporarily till the next deployment is done. I've added the codes related to update auth provider back, and only removed the endpoint for create auth provider, which was not used by any front end assets. Meanwhile, I'll create another ticket to refactor the update method since we need to understand how/if customers are using this method before we do anything.

SanketD92 (approved):
LGTM

PR #269, opened 2021-01-26
feat: front end and back end cidr update changes

SanketD92:
Issue #, if available:
GALI-657

Description of changes:
Currently, the CIDR block that has access to the workspace being provisioned is set in its respective Workspace Type Configuration. The default is set to <Workspace provisioner's IP>/32 and is stored in the 'cidr' field as a string value in the EnvironmentsSc object in the <namespace>-EnvironmentsSc table. This is just for the initial assignment of the CIDR block and should be considered obsolete when workspace provisioning completes.

The solution proposed attempts to mimic the EC2 Security Group’s Ingress Rules section. The security group ID would be acquired by the CloudFormation stack created during provisioning the workspace in the member account.

The “Edit CIDRs” button is available to both the owner of the workspace and the system administrators.

CidrUpdate (3)

CidrUpdate (1) (1) (1)

For more design details, please take a look at the "Update Workspace Ingress Rules Design" Quip document.

Checklist: 


addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvironmentUpdateCidrs.js
```diff
@@ -0,0 +1,308 @@
...
+    }));
+
+    const dropdownOnChange = action((_event, data) => {
+      this.ingressRules[index].cidrBlocks = data.value;
+    });
+
+    const anyWideCidr = () => {
+      return _.some(this.ingressRules[index].cidrBlocks, cidr => {
+        const parts = cidr.split('/');
+        const n = parts[parts.length - 1];
```
rsmayda, line 245:
```suggestion
        const prefixBits = parts[parts.length - 1];
```

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvironmentUpdateCidrs.js
```diff
@@ -0,0 +1,308 @@
...
+        return n <= 16;
+      });
+    };
+
+    const anyInvalidCidr = () => {
+      return _.some(this.ingressRules[index].cidrBlocks, cidr => IsCidr(cidr) !== 4);
```
rsmayda, line 251: where is the isCidr method coming from?

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-windows-instance.cfn.yml
```diff
@@ -46,10 +46,10 @@ Parameters:
   RecurringDownloads:
     Type: String
     Description: A flag indicating whether to keep syncing studies data to local EBS volumes on recurring basis. Setting this to false will download studies data only once at the instance bootstrap time. When this flag is set to true the instance will periodically sync changes from S3 to local EBS i.e., it will download any new files added to S3, re-download any files changed in S3 (will use object ETag value to determine if file changed in S3), delete files from local EBS if they are deleted from S3.
-    Default: "true"
+    Default: 'true'
```
rsmayda, line 49: Any functional change in this file?

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/environment-sc-controller.js
```diff
@@ -68,6 +68,26 @@ async function configure(context) {
...
+  // ===============================================================
+  router.post(
+    '/:id/cidr',
+    wrap(async (req, res) => {
+      const requestContext = res.locals.requestContext;
+      const id = req.params.id;
+      const updateRequest = req.body.cidr;
+
+      const [environmentScCidrService] = await context.service(['environmentScCidrService']);
+      const result = await environmentScCidrService.update(requestContext, {
```
rsmayda, line 82: Where does AuthN/Z happen?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+
+    // Write audit event
+    await this.audit(requestContext, { action: 'update-environment-sc-cidr', body: existingEnvironment });
+
+    return existingEnvironment;
+  }
+
+  getIpPermission(updateRule, cidrList) {
+    const ipv4List = _.filter(cidrList, cidr => IsCidr.v4(cidr));
+    // const ipv6List = _.filter(cidrList, cidr => IsCidr.v6(cidr));
```
rsmayda, line 108: nit remove

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+
+  async update(requestContext, { id, updateRequest }) {
+    // Before anything, check if the payload is valid
+    this.checkRequest(updateRequest);
+
+    const [environmentScService, lockService] = await this.service(['environmentScService', 'lockService']);
+    const existingEnvironment = await environmentScService.mustFind(requestContext, { id });
+
+    // Check if user is allowed to update cidrs
+    await this.assertAuthorized(
```
rsmayda, line 74: Does this also check if the user is allowed, ie owner of the sec group?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+    if (!_.isEmpty(ipv6Format))
+      throw this.boom.badRequest(
+        `Currently only IPv4 formatted CIDR inputs are allowed. Please remove : ${ipv6Format.join(
+          ', ',
+        )} CIDR ranges from your list`,
+        true,
+      );
+  }
+
+  async update(requestContext, { id, updateRequest }) {
```
rsmayda, line 66: Add logging

rsmayda (commented):
Generally looks good - please add unit tests

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+    const ipv4List = _.filter(cidrList, cidr => IsCidr.v4(cidr));
+    // const ipv6List = _.filter(cidrList, cidr => IsCidr.v6(cidr));
+
+    const IpPermission = {
+      FromPort: updateRule.fromPort,
+      IpProtocol: updateRule.protocol,
+      IpRanges: _.map(ipv4List, cidr => ({
+        CidrIp: cidr,
+        // For future: Custom description doesn't work, since that has to match exactly what's there in the SG.
+        // This would require storing cidr field in DB
```
jn1119, line 116: Isn't this comment misleading based on what we had discussed? We shouldn't require DB addition as long as we treat one IPAddress per rule.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+    }
+  }
+
+  async authorizeSecurityGroupIngress(ec2Client, authorizeParams) {
+    try {
+      await ec2Client.authorizeSecurityGroupIngress(authorizeParams).promise();
+    } catch (err) {
+      // Skip errors that pertain to rule already exists
+      if (err.code !== 'AlreadyExists')
+        throw this.boom.internalError(`There was a problem granting access in security group. Error: ${err}`, true);
```
jn1119, line 183: Do these errors get logged in Cloudwatch?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+    const IpPermission = {
+      FromPort: updateRule.fromPort,
+      IpProtocol: updateRule.protocol,
+      IpRanges: _.map(ipv4List, cidr => ({
+        CidrIp: cidr,
+        // For future: Custom description doesn't work, since that has to match exactly what's there in the SG.
+        // This would require storing cidr field in DB
+        // Description: `Updated via SWB`,
+      })),
+      // TODO: Ipv6 calls have to be done separately and cannot habe
```
jn1119, line 119: Complete this comment?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-service.js
```diff
@@ -655,6 +671,83 @@ class EnvironmentScService extends Service {
...
+        workspaceIngressRules,
+        workspaceRule =>
+          cfnRule.FromPort === workspaceRule.FromPort &&
+          cfnRule.ToPort === workspaceRule.ToPort &&
+          cfnRule.IpProtocol === workspaceRule.IpProtocol,
+      );
+      const currentCidrRanges = matchingRule
+        ? _.concat(
+            _.map(matchingRule.IpRanges, ipRange => ipRange.CidrIp),
+            _.map(matchingRule.Ipv6Ranges, ipRange => ipRange.CidrIp),
```
jn1119, line 736: So for Ipv6Ranges what is the flow like? We display them on the UI but if the updates contain one, we don't allow those requests?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+    });
+
+    // Get params ready
+    const revokeParams = { GroupId: securityGroupId, IpPermissions: revokeIpPermissions };
+    const authorizeParams = { GroupId: securityGroupId, IpPermissions: addIpPermissions };
+
+    // Start making security group updates
+    const ec2Client = await this.getEc2Client(requestContext, existingEnvironment.id);
+    if (!_.isEmpty(revokeIpPermissions)) await this.revokeSecurityGroupIngress(ec2Client, revokeParams);
+    if (!_.isEmpty(addIpPermissions)) await this.authorizeSecurityGroupIngress(ec2Client, authorizeParams);
```
jn1119, line 162: Since there are multiple networking calls involved, it is possible that we have a partial update scenario and throw an exception half way through the update. How does UI handle that. Will it display the partial update correctly? Would calling update once simplify things for us?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvironmentUpdateCidrs.js
SanketD92: https://www.npmjs.com/package/is-cidr

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-windows-instance.cfn.yml
SanketD92: No, just lint fixes.

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/environment-sc-controller.js
SanketD92: Inside the `environmentScCidrService.update` method

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
SanketD92: Yes, this check passes only when the user is the owner of the workspace or an application admin.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
SanketD92: Line 101.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/forms/CreateCidrFormModel.js
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  const extra = {};
+
+  const rules = {
+    'cidr': 'required',
+    'cidr[].fromPort': 'required|integer',
+    'cidr[].toPort': 'required|integer',
+    'cidr[].protocol': 'required',
+    'cidr[].cidrBlocks': 'array',
```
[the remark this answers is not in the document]
SanketD92: Yes, this is the model of the separated properties form that gets created on the UI.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+
+  checkRequest(updateRequest) {
+    if (_.isUndefined(updateRequest) || _.isEmpty(updateRequest))
+      throw this.boom.badRequest('The request made had an empty body. Please check your payload', true);
+
+    const erroneousInputs = [];
+    const ipv6Format = [];
+    _.map(updateRequest, rule => {
+      _.forEach(rule.cidrBlocks, cidrBlock => {
+        if (IsCidr(cidrBlock) === 0) erroneousInputs.push(cidrBlock);
```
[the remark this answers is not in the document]
SanketD92: No, this returns 0 for invalid CIDR. https://www.npmjs.com/package/is-cidr

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+    const addIpPermissions = [];
+    const revokeIpPermissions = [];
+    const newCidrList = [];
+
+    _.forEach(currentIngressRules, existingRule => {
+      const matchingRule = _.find(updateRequest, {
+        fromPort: existingRule.fromPort,
+        toPort: existingRule.toPort,
+        protocol: existingRule.protocol,
+      });
```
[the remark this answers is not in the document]
SanketD92: Not from the UI. All the CIDR blocks are grouped together in their respective `fromPort`, `toPort`, and `protocol` combinations. 

But since such grouping cannot be guaranteed by someone using the API directly, adding a check makes sense.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  async revokeSecurityGroupIngress(ec2Client, revokeParams) {
+    try {
+      await ec2Client.revokeSecurityGroupIngress(revokeParams).promise();
+    } catch (err) {
+      // Skip errors that pertain to rule not found
+      if (err.code !== 'EntityNotFound')
+        throw this.boom.internalError(`There was a problem revoking access in security group. Error: ${err}`, true);
+    }
```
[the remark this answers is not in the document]
SanketD92: According to EC2 documentation, security group ingress rule changes follow the eventual consistency method. This special condition is to cover unlikely scenarios when a newly added CIDR block hasn't been updated yet on the SG and is attempted to revoke.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
SanketD92: Yes. Had added that comment before our discussion. I'll update it.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
SanketD92: Yes

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-service.js
SanketD92: Correct. User cannot continue if Ipv6 is included in any of the rules.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/forms/CreateCidrFormModel.js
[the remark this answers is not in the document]
jn1119: Oops sorry, the comment was meant for line 40 for cidrBlocks. I wanted to know why they are not required?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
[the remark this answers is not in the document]
jn1119: In case we get an EntityNotFound errror, are we sure that our request went through? Is there a way to confirm that with a test or is that documented?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
SanketD92: There are at most only two calls made for any given update request, revoke and authorize. If there is any error during these calls, the UI throws an error and closes the Edit CIDRs component. If the user wishes to try again, they have to click on this component which fetches the latest state (partially updated) of the ingress rules.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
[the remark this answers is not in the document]
SanketD92: The request will fail whenever it encounters these errors. We're not ignoring these exceptions, we still throw them (I see now the comment is misleading, I'll change that). The exception catching is just to provide a security group-specific message, instead of showing "Something went wrong", which isn't useful at all.

After a failed/partial CIDR block update, the user would still be able to see the error message and the up to date ingress rules on that workspace.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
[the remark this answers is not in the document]
SanketD92: This only updates descriptions. For updating the actual CIDR blocks, we have to split them up into what needs to get added (authorize) and what needs to get removed (revoke).

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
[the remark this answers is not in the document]
jn1119: So we are catching the exception here and for any error except EntityNotFound we are throwing the error but for EntityNotFound we seem to be swallowing the exception and moving on since we are in catch block. Am I reading it wrong?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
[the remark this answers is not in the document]
jn1119: Nice, did I accidentally find how to update description without splitting the rules apart?

jn1119: ok, so the user will see the latest state during partial updates? Have you tested the experience by deliberately making say the authorize call fail?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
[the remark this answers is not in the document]
SanketD92: You're right. Removing exception swallowing part.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/forms/CreateCidrFormModel.js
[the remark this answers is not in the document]
SanketD92: Yeah, it should be required as well. I'll test and push that change.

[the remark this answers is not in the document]
SanketD92: If we make it required then users would see an error message when trying to remove all CIDR blocks from an ingress rule

Screen Shot 2021-01-27 at 5 29 28 PM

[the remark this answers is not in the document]
jn1119: Interesting! I guess I assumed some cidr range should be allowed otherwise what is the point of having a workspace around. I guess there might be some extreme case when you do this temporarily!

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
SanketD92: Yes. In case of partial updates, the user sees an error dialogue. The last updated Ingress Rule state is displayed when the UI Store (holding the value of the CIDR blocks) is updated (once every 30 sec), or when the page refreshes, whichever happens first.

main/solution/backend/config/infra/cloudformation.yml
```diff
@@ -801,9 +801,20 @@ Resources:
                 - ec2:ModifyImageAttribute
                 - ec2:DescribeImages
                 - ec2:DescribeInstances
+                - ec2:DescribeSecurityGroups
+                - ec2:RevokeSecurityGroupIngress
+                - ec2:AuthorizeSecurityGroupIngress
```
jn1119, line 806: Do we need to update our documentation to call out that member account roles need to add these permissions?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-cidr-service.test.js
```diff
@@ -0,0 +1,322 @@
...
+      // OPERATE
+      try {
+        await service.update(requestContext, params);
+        expect.hasAssertions();
+      } catch (err) {
+        expect(err.message).toEqual('User is not authorized');
+      }
+    });
+
+    it('should use the correct CIDR blocks to make revoke and authorize calls', async () => {
```
jn1119, line 228: I think there are a few more cases which can be added:

1. Update request and current ingress rules have some overlapping IPs and we have some IPs to revoke and some to register
2. Empty update request (just revoke)
3. Current ingress rules contain the same protocol, fromPort and toPort
4. Current ingress rules don't contain the protocol, fromPort and toPort in the request
5. Update request has IPs but current ingress rules have nothing (just authorize)
6. And adding failure scenarios like any revoke or authorize failure isn't swallowed and exception is thrown

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
```diff
@@ -820,4 +820,142 @@ describe('EnvironmentSCService', () => {
       );
     });
   });
+
+  describe('getSecurityGroupDetails function', () => {
```
jn1119, line 824: Probably add a scenario on how we handle cases where we don't find a securityGroupIngress. Sounds like an Error case instead of getting empty, right?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-cidr-service.test.js
```diff
@@ -0,0 +1,322 @@
...
+            protocol: 'tcp',
+            fromPort: 123,
+            toPort: 123,
+            cidrBlocks: ['123.123.123.123/32'],
+          },
+        ],
+      };
+      service.assertAuthorized.mockImplementationOnce(() => {
+        throw new Error('User is not authorized');
+      });
```
jn1119, line 217: This here looks good to me. Is it possible to add a couple of unit tests in auth module as well?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/forms/CreateCidrFormModel.js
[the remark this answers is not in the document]
SanketD92: Yes, it will always retain the originally configured allowed protocol-port combinations even if there are no CIDR blocks currently in them.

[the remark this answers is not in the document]
jn1119: Cool thanks for verifying. Resolving this thread.

main/solution/backend/config/infra/cloudformation.yml
SanketD92: I'll be updating this in the ChangeLog doc.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
SanketD92: Yes, I'll add that check with a test case. Thanks.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-cidr-service.test.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
SanketD92: Done

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-cidr-service.test.js
SanketD92: Working on it..

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-cidr-service.test.js
```diff
@@ -283,6 +331,264 @@ describe('EnvironmentScCidrService', () => {
...
+        GroupId: 'samplesecurityGroupId',
+        IpPermissions: [{ FromPort: 123, IpProtocol: 'tcp', IpRanges: [{ CidrIp: '4.4.4.4/32' }], ToPort: 123 }],
+      };
+      const authorizeParams = {
+        GroupId: 'samplesecurityGroupId',
+        IpPermissions: [
+          {
+            FromPort: 123,
+            IpProtocol: 'tcp',
+            IpRanges: [{ CidrIp: '2.2.2.2/32' }, { CidrIp: '3.3.3.3/32' }],
```
jn1119, line 565: 1.1.1.1/32 doesn't need to be included since it is already there?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-cidr-service.test.js
```diff
@@ -283,6 +331,264 @@ describe('EnvironmentScCidrService', () => {
       expect(service.authorizeSecurityGroupIngress).toHaveBeenCalledWith({}, authorizeParams);
     });
 
+    it('should not call anything since request has different protocol-port combinations', async () => {
```
jn1119, line 334: How does this not throw an exception/error. The updateRequest basically didn't work/go through. No?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-cidr-service.test.js
SanketD92: Correct.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-cidr-service.test.js
SanketD92: Done. Added those pertaining to the CIDR update switch-case.

jn1119 (approved):

Issue #123, opened 2020-08-12
cfn template "emr-cluster.cfn.yml" and related resource should be modified.


ghost:
deployment region - ap-northeast-2

~~1. In the template, Service Role for EMR does not have KMS policy. Thus SecurityConfiguration makes below problem.~~

~~The EMR Service Role must have the kms:GenerateDataKey* and kms:ReEncrypt* permission for the KMS key configuration when you enabled EBS encryption by default. You can retrieve that KMS key's ID by using the ec2:GetEbsDefaultKmsKeyId API.~~

~~KMS policy should be added to the Service Role.~~
_Identified as my member account setting problem_

2. Fail to bootstrap "us-east-1.elasticmapreduce/bootstrap-actions/run-if" on non us-east-1 region.
EMRFS could not download bootstrap file on other region. Thus it makes below problem.

Master instance failed attempting to download bootstrap action 1 file from S3

run-if script should be copied to the same region bucket and cfn template should be modified.

~~3. Additionally, cross account role also should have KMS policy.~~
_Identified as my member account setting problem_

rsmayda:
Hey are you still facing any issues I see that 1 & 3 seem to be account setting issues. Is 2 still applicable?

inhwan-aws:
Could you please make EMR cluster on the other region except us-east-1?

rsmayda:
Thanks for confirming issue 2 is still present -- we will keep this issue up to date with our progress

SanketD92:
@inhwan-aws Thanks for bringing this up. We have identified the potential root cause for this bug. The bootstrap action script path and IAM policy resource reference the us-east-1 region in the emr templates (built-in, service-catalog and external), which would need the following changes:

```
From
s3://us-east-1.elasticmapreduce/bootstrap-actions/run-if
to
!Sub s3://${AWS::Region}.elasticmapreduce/bootstrap-actions/run-if
```

```
From
'arn:aws:s3:::us-east-1.elasticmapreduce/bootstrap-actions/run-if'
to
!Sub 'arn:aws:s3:::${AWS::Region}.elasticmapreduce/bootstrap-actions/run-if'
```

We will be making and testing these changes soon. Stay tuned.

inhwan-aws:
Thank you, Sanket!

virgo81net:
@SanketD92 

In the process of combining the two accounts, the existing account (@inhwan-aws) was deleted, 
so please understand that previous posts and comments appear as ghost :)

[...]

PR #278, opened 2021-01-29
fix: Update EMR configuration and add run-if file to deploy any region

kwpark-amzn:
Issue #, if available:

Description of changes: 
- Current SWB can be just deployed us-east-1 region because of hard coding in addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/emr-cluster.cfn.yml.
- To  deploy any region  EMR cfn configuration update and run-if file to add into main/solution/post-deployment/config/environment-files/ is needed.

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/emr-cluster.cfn.yml
+++ b/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/emr-cluster.cfn.yml
@@ -151,7 +151,7 @@ Resources:
               - Effect: 'Allow'
                 Action: 's3:GetObject'
                 Resource:
-                  - 'arn:aws:s3:::us-east-1.elasticmapreduce/bootstrap-actions/run-if'
+                  - !Sub 'arn:aws:s3:::${EnvironmentInstanceFiles}/run-if'
                   - !Sub
                     - 'arn:aws:s3:::${S3Location}/*'
                     # Remove "s3://" prefix from EnvironmentInstanceFiles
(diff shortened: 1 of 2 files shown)


jn1119 (commented):
1. Can you update your commit message and prefix it with 'fix: '
2. What is the error when you tried deploying in another region besides `us-east-1`. Why does downloading the file cross region not work?

kwpark-amzn:
This is related with #123. 

Fail to bootstrap "us-east-1.elasticmapreduce/bootstrap-actions/run-if" on non us-east-1 region.
EMRFS could not download bootstrap file on other region. Thus it makes below problem.
"Master instance failed attempting to download bootstrap action 1 file from S3."

If you try to deploy SWB on Seoul region, you could know what it is exaclty.

jn1119 (commented):
Can we download the local `run-if` file only if the original file fails to download? Or can we use the local file only in regions where we see the error. My concern is that this `run-if` file is copied from EMR's repository and it can go through change and it would be hard to maintain that. We need to root cause this better that why the file cannot be downloaded cross-region.

kwpark-amzn:
I did not succeed to adop 1 you recommended even though I tried to edit cfn template in various ways. 
Could you let me know how I can adopt it?

SanketD92 (changes requested):
Could you update this PR to be against the develop branch? 

According to our updated processes, once your changes are merged to the develop branch and all tests pass on our workflow environment, they will be automatically merged into mainline.

jn1119 (commented):
I was able to verify that EMR provisioning worked and didn't fail due to bootstrap errors after adding region to S3 arn and url. 

https://github.com/awslabs/service-workbench-on-aws/pull/394

I am going to go ahead and close this PR once you verify that the new change works for you. Thanks for reporting the issue.

kwpark-amzn:
Great to hear that. Thanks for your effort to fix it. I agree to close this PR even though I do not test.

jn1119:
Fixed in PR https://github.com/awslabs/service-workbench-on-aws/pull/394

PR #280, opened 2021-01-29
docs: Add user and authN APIs in api docs

Bingjiling:
Issue #, if available:

Description of changes:

* Add OpenAPI docs for authZ and user APIs

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -18,6 +18,8 @@ tags:
     description: Operations available to guest users
   - name: guests
     description: Operations available to guest users
+  - name: public
+    description: Operations available to public
   - name: BYOB
     description: Operations for bring your own bucket configuration
   - name: SC-EnvType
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -50,7 +52,81 @@ tags:
...
+      tags:
+        - admin
+        - authN
+      summary: Get authentication provider configs detail
+      operationId: GetAuthNConfigsDetail
+      description: |
+        Get authentication provider configs detail
+      responses:
+        '200':
+          description: Array of authN provider configs detail
```
nguyen102, line 69: I noticed in the rest of the API docs, the description has the word "return" in the beginning of the description. Do we want to continue that pattern? For example in this case it would be: 
`Return array of authN provider configs detail`

nguyen102 (commented):
Overall, looks good. One question about format of response description.

jn1119 (approved):

openapi.yaml
Bingjiling: Good question! Just spot checked the doc, looks like we already have both, with or without `Return`. I'll leave it as is for now.

PR #290, opened 2021-02-04
feat: Integration Test setup

SanketD92:
Issue #, if available:
GALI-623

Description of changes:
Added a test mechanism that leverages the API endpoint of any SWB deployment to perform basic integration tests.

Checklist: 


main/integration-tests/helpers/api-param-generator.js
```diff
@@ -0,0 +1,345 @@
...
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+// **************** IMPORTANT  ****************
+// Please update this document if any SWB API is not found here
+// ********************************************
+
+const apiEndpoint = process.env.API_ENDPOINT;
+// const apiEndpoint = 'http://localhost:4000'; // When debugging locally
```
Bingjiling, line 21: Maybe this can be an input for the bash script? So tests can run against local deployment without code change.

main/integration-tests/utils/auth-tokens.js
```diff
@@ -0,0 +1,42 @@
...
+  throw new Error('Currently only internal auth provider is accepted in the integration test suite');
+}
+
+async function getInternalUserToken(username, password) {
+  const params = getAuthIdTokenParams({
+    username,
+    password,
+    authenticationProviderId: 'internal',
+  });
+  const response = await axios.post(params.api, params.payload);
```
Bingjiling, line 34: It'll be helpful to add a comment here suggesting the actual api path ``${apiEndpoint}/api/authentication/id-tokens`?

main/integration-tests/helpers/base-fixture.js
```diff
@@ -0,0 +1,70 @@
...
+      if (projectVerified && testAdminVerified) BaseFixture.ready = true;
+    } catch (err) {
+      throw new Error(`There was a problem verifying base fixture resources: ${err}`);
+    }
+  }
+
+  // Check if TestAdmin is actually admin
+  async verifyTestAdmin() {
+    const bearerToken = await getTestAdminToken(this.testConfig);
+    const allUsers = await listUsers(bearerToken);
```
Bingjiling, line 54: The `/api/user` GET endpoint should return the current user information.

main/integration-tests/helpers/api-param-generator.js
```diff
@@ -0,0 +1,345 @@
...
+
+function getWindowsRpInfoParams(envId, connectionId) {
+  return { api: `${apiEndpoint}/api/workspaces/service-catalog/${envId}/connections/${connectionId}/windows-rdp-info` };
+}
+
+module.exports = {
+  getAuthIdTokenParams,
+  addIndexParams,
+  listUsersParams,
+  addUsersParams,
```
Bingjiling, line 286: Two more user API in this controller - https://github.com/awslabs/service-workbench-on-aws/blob/master/addons/addon-base-rest-api/packages/base-controllers/lib/user-controller.js

main/integration-tests/utils/common.js
```diff
@@ -0,0 +1,34 @@
...
+ *  permissions and limitations under the License.
+ */
+
+const _ = require('lodash');
+
+// Resources created via integration tests should have this description, wherever possible.
+// This would make it easier in the future for audit/cleanup
+const RESOURCE_DESCRIPTION = 'Resource automatically created by SWB integration test';
+
+function validResponse(response, statusCode = 200, statusText = 'OK') {
```
Bingjiling, line 22: Nit: update function name to isValidResponse or validateResponse?

main/integration-tests/__test__/api-tests/studies/__fixtures__/update-study-fixture.js
```diff
@@ -0,0 +1,71 @@
...
+
+// Test Fixture
+/**
+ * This class is designed to pre-generate & help generate (using methods) resources
+ * that would be required by a SPECIFIC integration test suite
+ *
+ * The verification of these components are performed once every respective test suite cycle
+ * For example, whenever update-study-api test suite is triggered
+ */
+class UpdateStudyFixture extends BaseFixture {
```
Bingjiling, line 29: Nit: Can we remove the word update from the class name and file name for update-study-fixture.js and update-study-api.test.js? Once we start filling in integ tests, other study API tests will be added in here right?

Bingjiling (commented):
Looks good! A few nit picks.

main/integration-tests/helpers/api-param-generator.js
carvantes: There's no need to treat env variables as arguments for the bash script. You can just set the env variables before running integ tests.
i.e. I can run fwoa integ tests locally with:
```
API_URL=https://xxxxxx.execute-api.us-west-2.amazonaws.com/dev COGNITO_USERNAME=workshopuser COGNITO_PASSWORD=xxx COGNITO_CLIENT_ID=xxx yarn run int-test
```

main/integration-tests/__test__/api-tests/studies/update-study-api.test.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+    if (!UpdateStudyFixture.ready) {
+      await testFixture.setupPreRequisites();
+    }
+  });
+
+  describe('Update Study API', () => {
+    it('should fail while trying to update Open Data studies', async () => {
+      // BUILD
+      const adminBearerToken = await getTestAdminToken(testFixture.testConfig);
```
carvantes, line 40: There's a lot of passing bearer token around as argument to different functions. You can create an axios instance with fixed headers.

See: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/integration-tests/utils.ts#L43

I suggest having a function like:
```
const getSWBClient = (role) => {//returns axios instance with headers for user with that role}
```

main/integration-tests/__test__/api-tests/studies/update-study-api.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+      const updateRequest = { id: studyId, description: 'Sample desc change' };
+      const expectedErr = new Error('Request failed with status code 404');
+
+      const userA = await testFixture.createNonAdminUser(adminBearerToken, projectId);
+      const headers = { 'Authorization': userA.token, 'Content-Type': 'application/json' };
+      const params = updateStudyParams(studyId, updateRequest);
+
+      // EXECUTE & CHECK
+      await expect(axios.put(params.api, params.body, { headers })).rejects.toThrow(expectedErr);
```
carvantes, line 57: check the numeric status code instead

https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/integration-tests/search.test.ts#L85

carvantes (changes requested):

main/integration-tests/utils/studies.js
```diff
@@ -0,0 +1,70 @@
...
+
+// ************************ API calls ************************
+
+async function getStudy(bearerToken, studyId) {
+  const params = getStudyParams(studyId);
+  const headers = { 'Authorization': bearerToken, 'Content-Type': 'application/json' };
+  const response = await axios.get(params.api, { headers });
+
+  if (validResponse(response)) return response.data;
+  throw new Error('getStudy response was different than expected');
```
carvantes, line 44: I think we should remove this. 

Assertions should ideally be written on the actual test. Adding assertions here makes this method useless for testing negative scenarios like "should return 404 when getting a non existing study"

Also, throwing this error with a fixed message is bad for debugging since it hides the actual error. It's better to just throw the original errors or use `jest.expect`

carvantes (changes requested):

main/integration-tests/utils/common.js
carvantes, line 22: axios already throws errors for responses that are not 2xx. This function may not be needed

https://www.npmjs.com/package/axios#handling-errors

main/integration-tests/__test__/api-tests/studies/__fixtures__/update-study-fixture.js
SanketD92: Well, not necessarily. Depending on how many more tests are needed for the update study suite, and what the requirements for other study scenarios are, we might need to create test suites accordingly. 
Having said that, these test suites should be broad enough to not cause confusion (eg. ONLY one test suite for ALL study update scenarios).

The idea is that `main/integration-tests/__test__/api-tests/studies` will house all study-related test suites. Test fixtures need to be kept as independent as possible.

main/integration-tests/helpers/base-fixture.js
SanketD92: Unfortunately it does not provide userRole information which is needed here.

main/integration-tests/utils/common.js
SanketD92: Agreed. Removing.

main/integration-tests/utils/studies.js
SanketD92: Agreed. Removing.

main/integration-tests/__test__/api-tests/studies/update-study-api.test.js
SanketD92: Done

main/integration-tests/utils/common.js
SanketD92: Removing method altogether. (axios does this check for us)

main/integration-tests/utils/auth-tokens.js
SanketD92: Sure, but this shouldn't eventually become a requirement when using APIs from the `api-param-generator.js` since it would defeat the purpose of having all APIs abstracted away in a single module.

main/integration-tests/__test__/api-tests/studies/update-study-api.test.js
SanketD92: Done

main/integration-tests/helpers/api-param-generator.js
SanketD92: Making it configurable in the test config file.

main/integration-tests/helpers/api-param-generator.js
SanketD92: Added. Thanks

main/integration-tests/config/settings/example.yml
```diff
@@ -0,0 +1,14 @@
+# Enter the Test Administrator username and password below. (Please make sure this user has admin privileges)
+username: ""
+password: ""
+
+# This is a Project ID in your SWB deployment safe to be used by integration tests
```
nguyen102, line 5: Does this comment need to be rephrased? It looks like it's a run on sentence.

main/integration-tests/__test__/api-tests/studies/__fixtures__/update-study-fixture.js
```diff
@@ -0,0 +1,67 @@
...
+  // Create and add pre-requisite resources here as needed
+  async setupPreRequisites() {
+    // We currently only need helper methods to assist tests in the update-study test suite
+    // Since we don't have any pre-requisite resources needed, set this to true
+    UpdateStudyFixture.ready = true;
+  }
+
+  async createNonAdminUser(axiosClient, projectId) {
+    const testName = 'UpdateStudy';
+    const nonAdminUserJson = createUserJson({ projId: projectId, testName });
```
nguyen102, line 52: nitpick: When building the JSON and not making a request, maybe we can use the term "build". In this case it would be `buildUserJson`

main/integration-tests/__test__/api-tests/studies/__fixtures__/update-study-fixture.js
```diff
@@ -0,0 +1,67 @@
...
+    const response = await createUser(axiosClient, nonAdminUserJson);
+    const newUserClient = await getInternalUserClient(nonAdminUserJson.username, nonAdminUserJson.password);
+
+    return { ...response, password: nonAdminUserJson.password, axiosClient: newUserClient };
+  }
+
+  async createMyStudy(axiosClient, projectId) {
+    const testName = 'UpdateStudy';
+    const studyToCreate = createStudyJson({ projectId, testName });
```
nguyen102, line 61: buildStudyJson

main/integration-tests/__test__/api-tests/studies/update-study-api.test.js
```diff
@@ -0,0 +1,72 @@
...
+      // but let the test admin confirm that anyway
+      const studyId = '1000-genomes';
+      const projectId = testFixture.testConfig.projectId;
+      const openDataStudies = await listStudies(adminClient, 'Open Data');
+      const studyOfInterest = _.find(openDataStudies, study => study.id === studyId);
+      expect(studyOfInterest.category).toBe('Open Data');
+
+      const updateRequest = { id: studyId, description: 'Sample desc change' };
+
+      const userA = await testFixture.createNonAdminUser(adminClient, projectId);
```
nguyen102, line 45: `userA` can be renamed to `nonAdminUser`

main/integration-tests/__test__/api-tests/studies/update-study-api.test.js
```diff
@@ -0,0 +1,72 @@
...
+  beforeAll(async () => {
+    testFixture = new UpdateStudyFixture();
+    if (!UpdateStudyFixture.baseReady || _.isUndefined(UpdateStudyFixture.baseReady)) await testFixture.setupParent();
+    if (!UpdateStudyFixture.ready || _.isUndefined(UpdateStudyFixture.ready)) await testFixture.setupPreRequisites();
+  });
+
+  describe('Update Study API', () => {
+    it('should fail while trying to update Open Data studies', async () => {
+      // BUILD
+      const adminClient = await getTestAdminClient(testFixture.testConfig);
```
nguyen102, line 34: Instead of creating an `adminClient` and then passing `adminClient` to each function, what are your thoughts on creating a userClient, and passing the testConfig to the userClient
```
// userClient
class UserClient {
    let client;
    constructor(testConfig) {
       // initialize client
    }
    
    async listStudies(param) {
        //TODO: implement method
    }

}
```

So basically the idea is you pass the config into a `UserClient`. The client uses those credentials to make future API calls. You no longer hav to pass the `client` to each method.

nguyen102 (commented):
Overall looks good. I would suggest refactoring to encapsulate the axios client as part of an object, instead of passing the client to different functions. 

So for example, having a UserClient that can `listStudies`, `getStudies`, etc...

The test code would be along the lines of
```
const adminUserClient = new UserClient(adminAxiosClient);
await expect(adminUserClient.createStudy(study)).resolves.toEqual({
    //expected values
});


```

main/integration-tests/__test__/api-tests/studies/__fixtures__/update-study-fixture.js
```diff
@@ -0,0 +1,67 @@
...
+    UpdateStudyFixture.baseReady = BaseFixture.ready;
+  }
+
+  // IMPORTANT: These will be created afresh every new integration test cycle
+  // Create and add pre-requisite resources here as needed
+  async setupPreRequisites() {
+    // We currently only need helper methods to assist tests in the update-study test suite
+    // Since we don't have any pre-requisite resources needed, set this to true
+    UpdateStudyFixture.ready = true;
+  }
```
hatimkhan, line 48: I think you can eliminate all of these  methods and the static variables, buy using 'super'.
For example, you can have a method named async setup()  in the BaseFxiture, do the setup logic for the base in that method.  Then, in UpdateStudyFixture, you can have async setup() and inside the method you can call  await super.setup();  then do the setup logic for the update study.

main/integration-tests/helpers/base-fixture.js
```diff
@@ -0,0 +1,69 @@
...
+
+  async verifyTestProject() {
+    const params = getProjectParams(this.testConfig.projectId);
+    const axiosClient = await getTestAdminClient(this.testConfig);
+    const response = await axiosClient.get(params.api);
+    return !(_.isUndefined(response) || _.isEmpty(response));
+  }
+}
+
+module.exports = BaseFixture;
```
hatimkhan, line 69: Please see my comments in the update study fixture.  I am pasting it here:

> I think you can eliminate all of these methods and the static variables, by using 'super'.
For example, you can have a method named async setup() in the BaseFxiture, do the setup logic for the base in that method. Then, in UpdateStudyFixture, you can have async setup() and inside the method you can call await super.setup(); then do the setup logic for the update study.

main/integration-tests/config/settings/example.yml
SanketD92: Fixed. Coming in the next commit

main/integration-tests/__test__/api-tests/studies/__fixtures__/update-study-fixture.js
SanketD92: Fixed. Coming in the next commit

main/integration-tests/__test__/api-tests/studies/__fixtures__/update-study-fixture.js
SanketD92: Fixed. Coming in the next commit

main/integration-tests/__test__/api-tests/studies/update-study-api.test.js
SanketD92: Fixed. Coming in the next commit

main/integration-tests/__test__/api-tests/studies/update-study-api.test.js
SanketD92: So one to consider here.. Future tests might depend on a different number of utility methods, therefore their fixtures, which would be responsible for creating a UserClient shown above, would need to populate them accordingly for each test suite. Passing the axoisClient helps keep this flexible. 

Let me know if I'm missing something.

SanketD92: As discussed, adding a method in the test fixture for `getAdminUser` (similar to `createNonAdminUser`) which will return the user object along with the `axiosClient`

main/integration-tests/__test__/api-tests/studies/__fixtures__/update-study-fixture.js
SanketD92: As discussed, doing async setup() and not including the static variables approach.

main/integration-tests/helpers/base-fixture.js
SanketD92: As discussed, doing async setup() and not including the static variables approach.

main/integration-tests/__test__/api-tests/studies/__fixtures__/update-study-fixture.js
```diff
@@ -0,0 +1,61 @@
...
+class UpdateStudyFixture extends BaseFixture {
+  // IMPORTANT: These will be created afresh every new integration test cycle
+  async setup() {
+    // Create and add pre-requisite resources here as needed
+    // We currently only need helper methods to assist tests in the update-study test suite
+    // Since we don't have any pre-requisite resources needed, set this to true
+    await super.setup();
+  }
+
+  async getAdminUser() {
```
nguyen102, line 38: Should this be in the BaseFixture?

nguyen102 (approved):

carvantes (approved):

carvantes (approved):

PR #291, opened 2021-02-05
fix: Made external researcher used UserOnboarding template less permissive

jn1119:
Issue #, if available: GALI-20

Description of changes:
* Note that the change is only applicable for external-researchers, so in order to test the change I had to enable built-in and external-researcher config
* Verified that I was able to create/connect/terminate SageMaker instance
* Verified that I was able to create/terminate EMR
* RStudio/Windows/Linux EC2 are not supported for external-researchers, and they don't show up as an option for provisioning

Out of scope but broken
* Stopping SageMaker and connecting with EMR is broken currently, and I will create a separate issue about this
* Also note that we have permissions to createRole with permissions boundary. That is also overly permissive but the change would be more involved. I will create a separate issue in our backlog for this too.

Checklist: 

* [ ] Have you successfully ran unit tests and manual tests with your changes locally?


Bingjiling (approved):
LGTM

PR #292, opened 2021-02-06
fix: Added usernameInIdp property to update user schema

jn1119:
Issue #, if available:

Description of changes:

This change fixes authentication of users who are added by admin from ServiceWorkbench UI directly

With userId PR https://github.com/awslabs/service-workbench-on-aws/pull/148 we had broken the above use case since the field `usernameInIdp` was removed from update schema. With users added by admin, this property is tried to be added back to user table[1][2] during authentication but the update fails since the field was removed from schema. Note that when user does self registration via idp the field `usernameInIdp` is already present from idp authentication and since create-user schema supports it, the field is added successfully. I figured that keeping the field around is useful so added it to update schema as well. It seems like removing the field entirely would have worked too.

[1]https://github.com/awslabs/service-workbench-on-aws/blob/37878582af54d2741a1eeb47b4673bdcc295d1b5/addons/addon-base-rest-api/packages/services/lib/authentication-providers/built-in-providers/cogito-user-pool/user-attributes-mapper-service.js
[2]https://github.com/awslabs/service-workbench-on-aws/blob/master/addons/addon-base-rest-api/packages/services/lib/authentication-providers/built-in-providers/cogito-user-pool/provider-service.js#L134-L138

Tested the change under following scenarios:
1. Existing internal login works
2. New internal login works
3. Existing federated login works
4. New federated login registered by admin works
5. Self federated registration also works

Checklist: 

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/update-user.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/update-user.json
@@ -10,6 +10,12 @@
       "type": "string",
       "pattern": "^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\\.[a-zA-Z0-9-]+)*$"
     },
+    "usernameInIdp": {
+      "type": "string",
+      "minLength": 3,
+      "maxLength": 300,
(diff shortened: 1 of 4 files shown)


SanketD92 (approved):
LGTM

PR #301, opened 2021-02-11
Add study bug-fix and additional integration tests

SanketD92:
Issue #, if available:
GALI-700: Creating a study shows error popup

Description of changes:
Study-controller was incorrectly setting study.access to a string when it should've been a list.

Checklist: 


main/integration-tests/__test__/api-tests/studies/study.test.js
```diff
@@ -0,0 +1,132 @@
...
+    it('should pass while trying to get own studies', async () => {
+      // BUILD
+      const nonAdminUser = await testFixture.createNonAdminUser(testAdmin.axiosClient, testProjectId);
+      const studyA = await testFixture.createMyStudy(nonAdminUser.axiosClient, testProjectId);
+      const params = getStudyParams(studyA.id);
+
+      // EXECUTE & CHECK
+      const response = await nonAdminUser.axiosClient.get(params.api);
+      expect(response.data).toHaveProperty('projectId');
+      expect(response.data).toHaveProperty('rev');
```
carvantes, line 98: We should be checking that the fields in the response have the same values as `studyA`, not just checking their existence.

Also, it'd be better to use `.toMatchObject` instead of one assertion per field.

https://jestjs.io/docs/en/expect#tomatchobjectobject

main/integration-tests/__test__/api-tests/studies/study.test.js
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  describe('Get Study List API: /api/studies/', () => {
+    it('should pass while trying to get studies list', async () => {
+      // BUILD
+      const newUser = await testFixture.createNonAdminUser(testAdmin.axiosClient, testProjectId);
+
+      // EXECUTE & CHECK
+      const response = await listStudies(newUser.axiosClient, 'Open Data');
+      expect(response).not.toBeUndefined();
```
carvantes, line 120: We should check for something more specific. At the very least check that it responds with status 200.

Is the idea here to only test authz? are we adding later another test that actually tests that this api is listing studies? It'd be nice to have assertions that check that some known 'Open data' studies appear in the response.

carvantes (commented):
Why are integ tests not related to the bugfix part of this same PR?

The bug fix seems straightforward. The only comments I left are related to integ tests. If the PR was only about the bug fix you'd be able to merge it sooner.

main/integration-tests/__test__/api-tests/studies/study.test.js
SanketD92: Yeah, I did consider using this. The response object has some random string attached to it, so it might need some changes. I'll work on using toMatchObject

main/integration-tests/__test__/api-tests/studies/study.test.js
SanketD92: This is still a work in progress. I'll remove the tests that are not pertaining to this bug fix for now.

main/integration-tests/__test__/api-tests/studies/study.test.js
SanketD92: I've removed this test for now, still a work in progress.

carvantes (approved):

carvantes:
Fixes https://github.com/awslabs/service-workbench-on-aws/issues/303

PR #319, opened 2021-02-18
test: integration tests for user controller

Bingjiling:
Issue #, if available:

Description of changes:
* Add integration tests for user controller
* Add clean up method for users resource
* Add setup method to return username and session

Checklist: 


main/integration-tests/__test__/api-tests/user/update-user.test.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    it('should fail if non-admin user update restrictive fields', async () => {
+      const researcherSession = await setup.createResearcherSession();
+      await expect(
+        researcherSession.resources.currentUser.update({ rev: 0, isSamlAuthenticatedUser: false }),
+      ).rejects.toMatchObject({ code: errorCode.http.code.forbidden });
+    });
+
+    it('should not allow alleviate privilege', async () => {
```
SanketD92: You probably mean "elevate" here.

main/integration-tests/__test__/api-tests/user/update-user.test.js
```diff
@@ -0,0 +1,103 @@
...
+        },
+      );
+      // Admin should not alleviate to root
+      const admin2Session = await setup.createAdminSession();
+      await expect(admin2Session.resources.currentUser.update({ rev: 0, userRole: 'root' })).rejects.toMatchObject({
+        code: errorCode.http.code.notFound,
+      });
+    });
+
+    it('should not allow inactive user to become active', async () => {
```
SanketD92, line 65: Suggesting a change in description to something like: "'should not allow inactive user to make themselves active". Also add a test to ensure only admins can perform this operation successfully.

main/integration-tests/__test__/api-tests/user/update-user.test.js
```diff
@@ -0,0 +1,103 @@
...
+      const researcherSession = await setup.createResearcherSession();
+      await researcherSession.resources.currentUser.update({ rev: 0, status: 'inactive' });
+      await expect(researcherSession.resources.currentUser.update({ rev: 1, status: 'active' })).rejects.toMatchObject({
+        code: errorCode.http.code.forbidden,
+      });
+    });
+
+    // Note: This use-case is for federated user self-registration
+    it('should allow inactive user to become pending', async () => {
```
SanketD92, line 74: Same comment as above

main/integration-tests/__test__/api-tests/user/get-user.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+  afterAll(async () => {
+    await setup.cleanup();
+  });
+
+  describe('Getting current user', () => {
+    it('should fail for anonymous user', async () => {
+      const anonymousSession = await setup.createAnonymousSession();
+      await expect(anonymousSession.resources.currentUser.get()).rejects.toMatchObject({
+        code: errorCode.http.code.badImplementation,
```
hatimkhan, line 34: It is interesting that the API is returning this response, maybe we can add a comment to the first tab in the spreadsheet to see if it possible to return a more appropriate response code

main/integration-tests/support/setup.js
```diff
@@ -106,7 +111,7 @@ class Setup {
     const session = await getClientSession({ idToken, setup: this });
     this.sessions.push(session);
 
-    return session;
+    return { session, username };
```
SanketD92: Do we need this? I though `<session>.resources.currentUser.get()` also provides the username.

SanketD92: Nevermind, I see how you're using it.

main/integration-tests/__test__/api-tests/user/get-user.test.js
```diff
@@ -0,0 +1,56 @@
...
+      await expect(researcher1Session.resources.currentUser.get()).resolves.toEqual(
+        expect.objectContaining({ username }),
+      );
+    });
+
+    it('should return current user information for inactive and pending user', async () => {
+      const researcher2Session = await setup.createResearcherSession();
+      await expect(researcher2Session.resources.currentUser.get()).resolves.toMatchObject({ status: 'active' });
+
+      await researcher2Session.resources.currentUser.update({ status: 'inactive', rev: 0 });
```
hatimkhan: This is interesting.  I didn't know that the API allowed non admin users to change their status to inactive.

main/integration-tests/__test__/api-tests/user/get-user.test.js
```diff
@@ -0,0 +1,56 @@
...
+    });
+
+    it('should return current user information', async () => {
+      const { session: researcher1Session, username } = await setup.createResearcherSessionAndUserName();
+      await expect(researcher1Session.resources.currentUser.get()).resolves.toEqual(
+        expect.objectContaining({ username }),
+      );
+    });
+
+    it('should return current user information for inactive and pending user', async () => {
```
hatimkhan: If you want you can use it.each() and pass the different statuses, this way your test can be run against all the status that you want.  (Just a thought)

main/integration-tests/__test__/api-tests/user/update-user.test.js
```diff
@@ -0,0 +1,103 @@
...
+      const researcher1Session = await setup.createResearcherSession();
+      const researcher2Session = await setup.createResearcherSession();
+      const researcher2Info = await researcher2Session.resources.currentUser.get();
+      const researcher1Info = await researcher1Session.resources.currentUser.get();
+      await expect(
+        researcher1Session.resources.currentUser.update({ uid: researcher2Info.uid, rev: 0, status: 'pending' }),
+      ).resolves.toEqual(expect.objectContaining({ uid: researcher1Info.uid }));
+    });
+
+    it('should fail if non-admin user update restrictive fields', async () => {
```
hatimkhan: Maybe we can use it.each() this way you can provide a list of all the restricted fields and the test will run for each on of them.

main/integration-tests/support/setup.js
```diff
@@ -91,6 +91,11 @@ class Setup {
   }
 
   async createResearcherSession() {
+    const sessionInfo = await this.createResearcherSessionAndUserName();
```
hatimkhan: Per our discussion.  We can have a different approach to this.

main/integration-tests/support/resources/users/users.js
```diff
@@ -54,6 +54,10 @@ class Users extends CollectionResource {
     };
   }
 
+  async cleanup(user) {
```
hatimkhan, line 57: I am thinking of adding this as the default implementation in Resource. So, we might be able to remove it from here.

main/integration-tests/__test__/api-tests/user/get-user.test.js
Bingjiling: +1 I'll add it to the spreadsheet 

Sending a wrong/expired token does get an Unauthorized error, but when the token is missing a 500 is thrown

main/integration-tests/__test__/api-tests/user/get-user.test.js
Bingjiling: Nice! I didn't know it.each() before. Will update it.

main/integration-tests/__test__/api-tests/user/get-user.test.js
Bingjiling: Yes, this API allows user in Active state to change to any other state.

main/integration-tests/__test__/api-tests/user/update-user.test.js
Bingjiling: Good catch! My poor english T_T

main/integration-tests/__test__/api-tests/user/update-user.test.js
Bingjiling: The `/user` path can only be used to update current user, Admin update other user's status will be covered as part of the `/users` path.

main/integration-tests/__test__/api-tests/user/update-user.test.js
Bingjiling: Same answer as above :D

main/integration-tests/support/resources/users/users.js
Bingjiling: I'm leaning towards keeping it here as a temporary before we have the default implementation. I had 100+ users created in my environment yesterday when running integ tests, having it here saves a lot of trouble.

SanketD92 (approved):
LGTM

PR #333, opened 2021-02-19
test: Implemented remaining workspace-types apis test except delete

ealemayehu:
Delete test will be implemented once the Resource class supports it.


main/integration-tests/__test__/api-tests/workspace-types/get-workspace-types.test.js
```diff
@@ -0,0 +1,93 @@
...
+      });
+    });
+
+    it('should return approved if user is researcher', async () => {
+      const researcherSession = await setup.createResearcherSession();
+
+      await expect(researcherSession.resources.workspaceTypes.getApproved()).resolves.toEqual(expect.any(Array));
+    });
+
+    it('should return no not-approved if user is researcher', async () => {
```
Bingjiling: Nit: The description is a bit confusing, should it return, or should not?

main/integration-tests/__test__/api-tests/workspace-types/get-workspace-types.test.js
```diff
@@ -0,0 +1,93 @@
...
+      const researcherSession = await setup.createResearcherSession();
+      const workspaceTypeId = setup.gen.string({ prefix: 'workspace-test' });
+
+      await adminSession.resources.workspaceTypes.create({
+        id: workspaceTypeId,
+        status: 'not-approved',
+      });
+
+      await expect(researcherSession.resources.workspaceTypes.getNotApproved()).resolves.toEqual([]);
```
Bingjiling: Can we add not-approved workspace type so the return is not empty?

Bingjiling (commented):
Looks good! A few nit picks.

main/integration-tests/__test__/api-tests/workspace-types/get-workspace-types.test.js
ealemayehu: Thanks, I have updated this test case as well as other positive test cases to create the appropriate workspace type in case one does not already exist in the database. Note, I have also updated the logic to check the array.  If I expect the array to be empty, I call "toHaveLength(0)". If I expect the array to not be empty, I call "not.toHaveLength(0)"

main/integration-tests/__test__/api-tests/workspace-types/get-workspace-types.test.js
ealemayehu: Thanks, updated this and other descriptions

Bingjiling (approved):
LGTM

PR #335, opened 2021-02-19
feat: Framework update - default clean up logic implementation and AWS APIs supprting classes

hatimkhan:
This PR includes the following:

- Retrieve the test admin password from parameter store, eliminating the need to store it in yaml file
- No need to run the get-info.sh bash script (it was used inside the run-integration-test.sh)
- Default implementation of the clean up logic at the resource node level. Custom cleanup logic can override the default implementation.
- All the necessary supporting/foundational classes to allow easy access to AWS APIs

An example of accessing dynamoDB (only for the cases where we don't have an API to clean up the resource)

```javascript
const db = await setup.aws.services.dynamoDb();
const users = await db.tables.users.scanner().scan();
```

Checklist:

- [X] Have you successfully deployed to an AWS account with your changes?


main/integration-tests/support/aws/services/dynamodb.js
```diff
@@ -0,0 +1,59 @@
...
+    // the dynamo update operation fails with
+    // "ExpressionAttributeValues contains invalid value: One or more parameter values were invalid: An AttributeValue may not contain an empty string for key :desc" error
+    // See https://github.com/aws/aws-sdk-js/issues/833 for details
+    this.client = new this.aws.sdk.DynamoDB.DocumentClient({
+      convertEmptyValues: true,
+    });
+
+    this.helpers = {
+      unmarshal,
+      scanner: () => new Scanner(this.log, this.client),
```
SanketD92: Where is `this.log` being initialized? If it's undefined right now, do we plan on creating a `log` variable in the future?

hatimkhan: Inside the scanner, the default is console.  I see how this might be confusing, let me fix it.

main/integration-tests/support/complex/delete-study.js
```diff
@@ -0,0 +1,150 @@
...
+  // - Delete the study entry from the study permissions table
+  // - We delete the study from the studies table
+  // - We delete the study s3 folder in the default data bucket (if the path contains runId)
+  //   We need to revisit this logic once the BYOB branch is merged
+
+  const runId = aws.settings.get('runId');
+
+  if (!id.includes(runId)) {
+    console.log(
+      `Study id "${id}" does not contain the runId "${runId}", skipping the deletion of this study as a measurement of caution`,
```
SanketD92, line 45: Could we look for `-${id}-` instead, this would avoid unlikely deletion mistakes for a study named like 'StudyFor`runId`Research'. (Also, currently we have all studies creating by the test having the string -test- in them as well. Maybe we can leverage that too here?)

SanketD92: Could we have the studyIds populated in an array, and then have them passed to this cleanup logic instead? That way we're not trying to match test-created studyIds with those already in the DB.

SanketD92, line 45: Also check if runId has prefix and suffix as a `-`. (eg. `add-study-${runId}-randomString`)

hatimkhan: Good idea.  Will add it.

main/integration-tests/support/complex/delete-user.js
```diff
@@ -0,0 +1,106 @@
...
+
+  const runId = aws.settings.get('runId');
+  const db = await aws.services.dynamoDb();
+  const user = await db.tables.users
+    .getter()
+    .key({ uid: id })
+    .get();
+
+  if (_.isEmpty(user)) {
+    console.log(`User with id "${id}" does not exist, skipping the deletion of this user`);
```
SanketD92, line 46: Same comment as above (`-${id}-`)

main/integration-tests/support/resources/base/collection-resource.js
```diff
@@ -70,8 +79,11 @@ class CollectionResource {
     return this.doCall(async () => this.axiosClient.get(api, { params }));
   }
 
-  // TODO - delete
-  // async delete
+  // In general, most of SWB APIs on the server side should not support the ability to delete a collection
+  // resource. However, it might be desireable that we tes against this. Therefore, this method exists.
```
SanketD92: nit: "test" typo

main/integration-tests/support/resources/projects/project.js
```diff
@@ -29,6 +29,11 @@ class Project extends Resource {
     if (_.isEmpty(parent)) throw Error('A parent resource was not provided to resource type [project]');
   }
 
+  async cleanup() {
+    if (this.id === this.setup.gen.defaultProjectId()) return;
```
SanketD92, line 33: Why is user resource cleanup is in its CollectionResource class, while project resource cleanup needs to be in the Resource class? (I'm assuming this succeeds in deleting the project)

SanketD92 (approved):
LGTM

PR #336, opened 2021-02-22
Feat index API tests

SanketD92:
Issue #, if available:
GALI-714

Description of changes:
1. Integration tests for Indexes APIs.
2. Improved study test resource naming

Checklist:

AS review ticket id:

### DIFF
--- a/main/integration-tests/support/utils/generators.js
+++ b/main/integration-tests/support/utils/generators.js
@@ -34,7 +34,7 @@ async function getGenerators({ setup }) {
     lastName: () => `TestUser${chance.last({ nationality: 'en' })}`,
     description: () => `Resource automatically created by SWB integration test - ${runId}`,
     defaultProjectId: () => setup.settings.get('projectId'),
-    defaultIndexId: async () => setup.getDefaultIndexId(),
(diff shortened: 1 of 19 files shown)


main/integration-tests/support/utils/generators.js
```diff
@@ -34,6 +34,7 @@ async function getGenerators({ setup }) {
     lastName: () => `TestUser${chance.last({ nationality: 'en' })}`,
     description: () => `Resource automatically created by SWB integration test - ${runId}`,
     defaultProjectId: () => setup.settings.get('projectId'),
+    defaultIndexId: async () => setup.getDefaultIndexId(),
```
hatimkhan: Hmm, this might be an issue. Because all generators are not async, except this one.  What you can do is look up the value in the init method of setup and then use it, instead of looking it up every time we call a generator method.

Originally, I was thinking that defaultProjectId() does not belong to generator, since we are not actually generating this.  But, it was convenient making it available via the generator.

SanketD92: Made it async.

hatimkhan (approved):
LGTM

PR #342, opened 2021-02-23
Feat accounts integ tests

SanketD92:
Issue #, if available:
GALI-716

Description of changes:
1. New integration tests for Accounts APIs
2. Added validation for AwsAccountId check in indexes-service
3. Some restructuring for default ID handling in integration test setup

Checklist:

AS review ticket id: N/A

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/indexes/indexes-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/indexes/indexes-service.js
@@ -143,7 +143,7 @@ class IndexesService extends Service {
       try {
         await awsAccountsService.mustFind(requestContext, { id: rawData.awsAccountId });
       } catch (err) {
-        throw this.boom.badRequest(`Incorrect AWS Account ID provided`, true).cause(err);
+        throw this.boom.badRequest('Incorrect AWS Account ID provided', true).cause(err);
       }
     }
 
--- a/addons/addon-base-raas/packages/base-raas-services/lib/indexes/indexes-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/indexes/indexes-service.js
@@ -92,7 +92,7 @@ class IndexesService extends Service {
     try {
       await awsAccountsService.mustFind(requestContext, { id: rawData.awsAccountId });
     } catch (err) {
-      throw this.boom.badRequest(`Incorrect AWS Account ID provided`, true).cause(err);
+      throw this.boom.badRequest('Incorrect AWS Account ID provided', true).cause(err);
     }
 
     // For now, we assume that 'createdBy' and 'updatedBy' are always users and not groups
--- a/addons/addon-base-raas/packages/base-raas-services/lib/indexes/indexes-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/indexes/indexes-service.js
@@ -29,7 +29,13 @@ const settingKeys = {
 class IndexesService extends Service {
   constructor() {
     super();
-    this.dependency(['jsonSchemaValidationService', 'authorizationService', 'dbService', 'auditWriterService']);
+    this.dependency([
+      'jsonSchemaValidationService',
+      'authorizationService',
+      'dbService',
+      'auditWriterService',
+      'awsAccountsService',
+    ]);
   }
 
   async init() {
@@ -76,8 +82,18 @@ class IndexesService extends Service {
     );
 
     // Validate input
-    const [validationService] = await this.service(['jsonSchemaValidationService']);
+    const [validationService, awsAccountsService] = await this.service([
+      'jsonSchemaValidationService',
+      'awsAccountsService',
+    ]);
     await validationService.ensureValid(rawData, createSchema);
+
+    // Make sure the AWS Account actually exists
+    try {
+      await awsAccountsService.mustFind(requestContext, { id: rawData.awsAccountId });
+    } catch (err) {
+      throw this.boom.badRequest(`Incorrect AWS Account ID provided`, true).cause(err);
+    }
 
     // For now, we assume that 'createdBy' and 'updatedBy' are always users and not groups
     const by = _.get(requestContext, 'principalIdentifier.uid');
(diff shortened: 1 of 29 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/indexes/indexes-service.js
```diff
@@ -76,9 +82,19 @@ class IndexesService extends Service {
...
+      'jsonSchemaValidationService',
+      'awsAccountsService',
+    ]);
     await validationService.ensureValid(rawData, createSchema);
 
+    // Make sure the AWS Account actually exists
+    try {
+      await awsAccountsService.mustFind(requestContext, { id: rawData.awsAccountId });
+    } catch (err) {
+      throw this.boom.badRequest(`Incorrect AWS Account ID provided`, true).cause(err);
```
hatimkhan: Since you are not using any string expression, you might want to use single quote instead.  Otherwise, you might get an eslint issue.

main/integration-tests/support/setup.js
```diff
@@ -60,11 +60,12 @@ class Setup {
...
-    this.defaultIndexId = await this.getDefaultIndexId();
-
-    // AWS account ID linked to default test project provided
-    this.defaultAwsAccountId = await this.getDefaultAwsAccountId();
+    // An object to abstract out the default setup IDs (eg. test project ID)
+    this.defaults = {};
+    this.defaults.projectId = this.settings.get('projectId');
+    this.defaults.indexId = await this.getDefaultIndexId();
+    this.defaults.awsAccountId = await this.getDefaultAwsAccountId();
+    this.defaults.stepTemplateId = 'st-obtain-write-lock';
```
hatimkhan: For future enhancement,  we can capture this in a different file similar to how we did the getGenerators() and getServices().

main/integration-tests/support/setup.js
```diff
@@ -82,13 +83,13 @@ class Setup {
 
   async getDefaultIndexId() {
     const adminSession = await this.defaultAdminSession();
-    const defaultProject = await adminSession.resources.projects.project(this.gen.defaultProjectId()).get();
+    const defaultProject = await adminSession.resources.projects.project(this.defaults.projectId).get();
```
hatimkhan: Doing this.defaults.projectId (inside this function) does require that getDefaultIndexId() be called at a specific sequence when building the defaults. I would suggest that you make the function as follows:
`async getDefaultIndexId(projectId) { ... }`
This will eliminate the need to use this.defaults when we are still initializing it.

main/integration-tests/support/setup.js
```diff
@@ -82,13 +83,13 @@ class Setup {
...
     return defaultProject.indexId;
   }
 
   async getDefaultAwsAccountId() {
     const adminSession = await this.defaultAdminSession();
-    const defaultIndex = await adminSession.resources.indexes.index(this.defaultIndexId).get();
+    const defaultIndex = await adminSession.resources.indexes.index(this.defaults.indexId).get();
```
hatimkhan: Doing this.defaults.indexId (inside this function) does require that getDefaultAwsAccountId() be called at a specific sequence when building the defaults. I would suggest that you make the function as follows:
`async getDefaultAwsAccountId(indexId) { ... }`
This will eliminate the need to use this.defaults when we are still initializing it.

main/integration-tests/support/setup.js
SanketD92: Added as a comment, for now.

main/integration-tests/support/setup.js
SanketD92: Combined in one method as discussed

main/integration-tests/support/setup.js
SanketD92: Combined in one method as discussed

addons/addon-base-raas/packages/base-raas-services/lib/indexes/indexes-service.js
```diff
@@ -116,9 +132,21 @@ class IndexesService extends Service {
...
+      'awsAccountsService',
+    ]);
     await validationService.ensureValid(rawData, updateSchema);
 
+    // Make sure the AWS Account actually exists
+    if (!_.isUndefined(rawData.awsAccountId)) {
+      try {
+        await awsAccountsService.mustFind(requestContext, { id: rawData.awsAccountId });
+      } catch (err) {
+        throw this.boom.badRequest(`Incorrect AWS Account ID provided`, true).cause(err);
```
hatimkhan: Same ` vs '   if this is not giving you linting issues, then we can leave it as is

main/integration-tests/__test__/api-tests/accounts/get-account.test.js
```diff
@@ -0,0 +1,77 @@
...
+    await setup.cleanup();
+  });
+
+  describe('Getting an Account', () => {
+    it('should fail if user is inactive', async () => {
+      const admin2Session = await setup.createAdminSession();
+      await adminSession.resources.users.deactivateUser(admin2Session.user);
+      const defaultAwsAccount = await adminSession.resources.awsAccounts.mustFindByAwsAccountId(
+        setup.defaults.awsAccount.id,
+      );
```
hatimkhan: Is there a reason why you need to load the account again?
Can you just do
`const defaultAwsAccount = setup.default.awsAccount`?

main/integration-tests/support/setup.js
```diff
@@ -80,16 +77,23 @@ class Setup {
...
+    const indexId = project.indexId;
+    const index = await adminSession.resources.indexes.index(indexId).get();
+
+    const awsAccountId = index.awsAccountId;
+    const awsAccount = await adminSession.resources.awsAccounts.awsAccount(awsAccountId).get();
+
+    const stepTemplate = await adminSession.resources.stepTemplates
+      .versions('st-obtain-write-lock')
+      .version(1)
+      .get();
```
hatimkhan, line 94: I am wondering if we need this extra call.  None of the step template tests need any other property of the step template object, just the id.

hatimkhan (dismissed):

hatimkhan (dismissed):

main/integration-tests/support/setup.js
SanketD92: Just to keep it consistent like other returned resources.

hatimkhan (dismissed):

hatimkhan (approved):

PR #351, opened 2021-03-02
chore: add schema checking to the DataSource APIs

rsmayda:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/register-data-source-account.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/register-data-source-account.json
@@ -12,7 +12,7 @@
     "description": {
       "type": "string",
       "maxLength": 2048,
-      "pattern": "^([A-Za-z0-9-_ .,!?/#$]+)$"
+      "pattern": "^([A-Za-z0-9-_ .,!?/#$;]+)$"
     },
     "contactInfo": {
       "type": "string",
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/register-data-source-account.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/register-data-source-account.json
@@ -6,14 +6,17 @@
     "name": {
       "type": "string",
       "minLength": 1,
-      "maxLength": 300
+      "maxLength": 300,
+      "pattern": "^([A-Za-z0-9-_ ]+)$"
     },
     "description": {
       "type": "string",
-      "maxLength": 2048
+      "maxLength": 2048,
+      "pattern": "^([A-Za-z0-9-_ .,!?/#$]+)$"
     },
     "contactInfo": {
       "type": "string",
+      "format": "email",
       "maxLength": 2048
     },
     "id": {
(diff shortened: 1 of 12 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/schema/register-data-source-account.json
```diff
@@ -6,14 +6,17 @@
...
     },
     "description": {
       "type": "string",
-      "maxLength": 2048
+      "maxLength": 2048,
+      "pattern": "^([A-Za-z0-9-_ .,!?/#$;]+)$"
     },
     "contactInfo": {
       "type": "string",
+      "format": "email",
```
hatimkhan: Can we use a pattern instead?  The contact info is supposed to be a free form, the user can input a combination of  phone numbers, names, addresses  and emails.

addons/addon-base-raas/packages/base-raas-services/lib/schema/register-study.json
```diff
@@ -11,15 +11,17 @@
     },
     "name": {
       "type": "string",
-      "maxLength": 2048
+      "maxLength": 300,
+      "pattern": "^([A-Za-z0-9-_ ]+)$"
```
hatimkhan: Do we want to handle cases where study names can contain umlauts and possibly other characters.

addons/addon-base-raas/packages/base-raas-services/lib/schema/register-study.json
```diff
@@ -30,11 +32,13 @@
     "folder": {
       "type": "string",
       "minLength": 1,
-      "maxLength": 1000
+      "maxLength": 1000,
+      "pattern": "^([A-Za-z0-9-_ ]+)$"
```
hatimkhan: We might want to account for more characters and umlauts.  including '@'.  In SWB default buckets, some folders contain '@'.

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-data-source-account.json
```diff
@@ -8,19 +8,23 @@
...
     },
     "description": {
       "type": "string",
-      "maxLength": 2048
+      "maxLength": 2048,
+      "pattern": "^([A-Za-z0-9-_ .,!?/#$;]+)$"
     },
     "contactInfo": {
       "type": "string",
+      "format": "email",
```
hatimkhan: Can we use a pattern instead?  The contact info is supposed to be a free form, the user can input a combination of  phone numbers, names, addresses  and emails.

hatimkhan:
Thank you Robert for updating the schema. In terms of Cross Site Scripting (XSS), in our design we don't depend on trying to prevent the user from entering characters like < or > or strings like <script>alert()</script>.  I agree it is a good idea to add one additional layer, we just need to make sure that we accounted for international study names.

In our current design we address XSS in two additional layers that we have in place:

- When we build Dynamo expressions and send data to the database, we use the AWS SDK document client which handles strings correctly.
- When we display anything in the browser, we use React which (unless you explicitly tell it to not do so) will always escape characters to avoid XSS.

addons/addon-base-raas/packages/base-raas-services/lib/schema/register-data-source-account.json
rsmayda: Why give so many options it seems confusing? Do we have a use-case that the contactInfo couldn't be an email?

hatimkhan (approved):

SanketD92 (approved):
LGTM

PR #369, opened 2021-03-05
test: Added delays to POST/PUT/DELETE APIs to reduce flakey tests

ealemayehu:


main/integration-tests/support/complex/default-integration-test-product.js
```diff
@@ -29,6 +31,9 @@ async function createDefaultServiceCatalogProduct(setup) {
 
   const templateS3Path = `s3://${bucket}/${key}`;
 
+  // Large delay needed to ensure the above service catalog changes are persisted.
+  await sleep(30000);
+
```
jn1119: Do we need a delay right after `createProduct` as well? Or `associateProductWithPortfolio` `createConstraint` are ServiceCatalog APIs, so they work well with eventual consistency of createProduct API?

jn1119 (commented):
LGTM overall, had one minor callout

main/integration-tests/support/complex/default-integration-test-product.js
ealemayehu: I tried what you recommended here and it did not make a difference. The issue of the error we are encountering is with associateProductWithPortfolio. The error is:

The product prod-4ziaxr6zdvk2a is not shared with the arn:aws:iam::370253196967:role/eyor-va-sw-EnvMgmt role

The role is the role of the portfolio.

jn1119 (approved):

PR #381, opened 2021-03-11
fix: remove the POST workflow status query endpoint

hatimkhan:
The current POST /api/workflows/instances/status/{status}  API is not used and the underlying database index is missing from the beginning.  No services or other APIs depend on this API.

Checklist:

- [X] Have you successfully deployed to an AWS account with your changes?
- [X] Have you successfully tested with your changes locally?


jn1119 (approved):

jn1119 (changes requested):
Seems like we need to remove this reference as well?

https://github.com/awslabs/service-workbench-on-aws/blob/e800cea5f30aa2208e11962207f6c2e181ddbde6/addons/addon-base-workflow-ui/packages/base-workflow-ui/src/helpers/api.js#L93-L95

jn1119 (approved):

PR #387, opened 2021-03-12
test: Delays to fix flakes caused when getting ID tokens

ealemayehu:


main/integration-tests/support/setup.js
```diff
@@ -131,6 +136,10 @@ class Setup {
       userRole: 'researcher',
       projectId,
     });
+
+    // Delay to prevent eventual consistency issues from affecting the next call.
+    await sleep(1000);
+
```
hatimkhan: Doesn't the 'create'  already introduce a delay?

main/integration-tests/support/setup.js
```diff
@@ -151,6 +160,10 @@ class Setup {
       userRole,
       projectId,
     });
+
+    // Delay to prevent eventual consistency issues from affecting the next call.
+    await sleep(1000);
```
hatimkhan: Doesn't the 'create' already introduce a delay?

main/integration-tests/support/setup.js
ealemayehu: Thanks for catching this! I have updated the default deflake delay from 1 to 2 seconds. The 1 second delay is what the normal maximum time dynamodb needs to persist eventually consistent changes.

main/integration-tests/support/setup.js
ealemayehu: Fixed as explained above

hatimkhan (approved):
LGTM

PR #423, opened 2021-04-05
chore: Add custom user agent

nguyen102:
Issue #, if available:

Description of changes:
chore: Add custom user agent

This was added to all lambdas in the `backend` solution and also the `post-deployment` solution.

Confirmation that the change worked, base on the logs in Cloud Trail. `AwsLabs/SO0144/2.1.2` is in the `userAgent`
```
{
    ...
    "eventTime": "2021-04-05T16:57:21Z",
    "eventSource": "s3.amazonaws.com",
    "eventName": "ListObjects",
    "awsRegion": "us-east-2",
    "sourceIPAddress": "18.217.252.15",
    "userAgent": "[aws-sdk-nodejs/2.804.0 linux/v12.21.0 exec-env/AWS_Lambda_nodejs12.x AwsLabs/SO0144/2.1.2 promise]",
    ...
}
```
Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/budgets/__tests__/budgets-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/budgets/__tests__/budgets-service.test.js
@@ -18,6 +18,7 @@ const sinon = require('sinon');
 const ServicesContainer = require('@aws-ee/base-services-container/lib/services-container');
 const Aws = require('@aws-ee/base-services/lib/aws/aws-service');
 const JsonSchemaValidationService = require('@aws-ee/base-services/lib/json-schema-validation-service');
+const SettingsService = require('@aws-ee/base-services/lib/settings/env-settings-service');
 const AWSMock = require('aws-sdk-mock');
 
 // Mocked dependencies
@@ -52,6 +53,15 @@ describe('BudgetsService', () => {
   beforeAll(async () => {
     // Initialize services container and register dependencies
     const container = new ServicesContainer();
+    const settingsService = new SettingsService();
+    settingsService.get = jest.fn(key => {
+      if (key === 'customUserAgent') {
+        return 'AwsLabs/SO0144/X.Y.Z';
+      }
+      throw new Error('Unexpected key');
+    });
+    container.register('settings', settingsService);
+
     container.register('awsAccountsService', new AWSAccountsServiceMock());
     container.register('aws', new Aws());
     container.register('jsonSchemaValidationService', new JsonSchemaValidationService());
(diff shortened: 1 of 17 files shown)


main/config/settings/.defaults.yml
```diff
@@ -170,3 +170,6 @@ auditQldbName: ${self:custom.settings.namespace}-audit
 # Enable this to create a Service Catalog portfolio and populate workspace products
 # Override and disable this setting if you wish to create your portfolio manually
 createServiceCatalogPortfolio: true
+
+# Version number of current release
+customUserAgent: AwsLabs/SO0144/2.1.2
```
rsmayda: A new patch is about to be released FYI
```suggestion
customUserAgent: AwsLabs/SO0144/2.1.3
```

jn1119: Does this need to change with every release? I also wonder if this is the best place to put this and make it configurable given that this file is our environment specific settings and we don't want customers to even change this. What do you think?

nguyen102: Yep, this value will need to be updated with new releases. This will ensure the AWS SDK requests have the correct version number. 

Since this is the `.default` config file, I don't expect customers to actually edit the values in this file. They'll create their own environment specific files that will inherit from the `.default` config file. I placed the value here as this seems to be the only spot in which we can pass a value to all SDCs. I'm not as familiar with SWB, so please let me know if there are other places in which I can place a value and have it propagate to other SDCs.

jn1119: I actually had something complex in mind- something along the lines of pulling version from change log and adding it to an SSM parameter as part of a deployment script. Later the parameter can be used for various purposes like displaying the version in UI, using it for user agent or even giving warning to users if they are doing a major version release change in the environment. But it seems like we didn't scope the automation as part of the issue you were working on.

addons/addon-base/packages/services/lib/aws/aws-service.js
```diff
@@ -39,6 +39,10 @@ class AwsService extends Service {
       this._sdk = AWSXRay.captureAWS(require('aws-sdk'));
     }
 
+    this._sdk.config.update({
+      customUserAgent: this.settings.get('customUserAgent'),
+    });
```
jn1119: what is the behavior if this field is missing? And how does local API testing work? I am guessing we don't want the SDK to have this enabled when service is deployed locally?

jn1119 (dismissed):

jn1119 (approved):

addons/addon-base-raas/packages/base-raas-services/lib/budgets/__tests__/budgets-service.test.js
```diff
@@ -52,6 +53,15 @@ describe('BudgetsService', () => {
   beforeAll(async () => {
     // Initialize services container and register dependencies
     const container = new ServicesContainer();
+    const settingsService = new SettingsService();
+    settingsService.get = jest.fn(key => {
+      if (key === 'customUserAgent') {
```
Bingjiling, line 58: Good catch!

Bingjiling (approved):
LGTM

PR #435, opened 2021-04-08
fix: Fix cost dashboard bugs

jn1119:
Bug 1: Fix cost per index when index has more than one environments that have non-zero cost
Bug 2: Collect cost by environment Id so that identical environment names don't overwrite costs

Issue #, if available: GALI-689, GALI-781

Description of changes:

1. Bug 1: Any time an index had multiple non-zero costs associated with it, we would not track the costs correctly, and the result was that the whatever was the last environment seen for that index, we would just use its cost.  Mainly I noticed that `_.get(projNameToUserTotalCost, `${projName}.${user}`, 0);` always returned a zero since user contains a '.' since it is an email address so I have changed it to `_.get(projNameToUserTotalCost[projName], user, 0);`. This fixed the cost reporting on index and total cost. Note that this resulted in both- inaccurate costs or costs not showing up at all since some environment might have had no costs associated with it and that might have been the final environment in the index that the code saw and assumed that index's cost is zero USD

Before snapshot where there were multiple environments in the index but cost of just one environment was showing up in total
image

After the fix, the index costs are getting reflected correctly
image

2. Bug 2:  Before this change we were treating environment names as unique when they aren’t necessarily unique. This results in the code using and showing costs for one of the environment and messes up with the total cost reflected for projects/indexes. As an example: if you have two workspaces with the same name- cost of one of them will replace the other when we display costs in our dashboard and only one of the costs would be used when aggregating the cost for projects. This can lead to accuracy issue and also result in costs not being reflected at all (assuming an environment with zero cost replaces an environment with non-zero cost)
 
To distinguish the environments with same name in the UI we only have environmentId as the identifier where uniqueness is guaranteed. Since environment name isn’t unique, the change here will display the name multiple times but add environmentId along with its name to indicate that there are multiple environments that share the same name.

As an example have a look at the following screenshot where environmentId is appended since environment name wasn't unique. Note that environmentId is only appended when we have duplicate names

image

Note: This issue doesn't fix the scaling issue where we make calls for all the environments together and the fact that our cost explorer UI doesn't handle pagination. This only fixes the accuracy issues which can lead to incorrect cost data or cost dashboard not showing up at all. We have a separate epic for fixing such scaling issues.

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
+++ b/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
@@ -36,7 +36,9 @@ class Dashboard extends React.Component {
       totalCost: 0,
       projNameToTotalCost: {},
       projNameToUserTotalCost: {},
-      envNameToCostInfo: {},
+      envIdToCostInfo: {},
+      envIdToEnvMetadata: {},
+      duplicateEnvNames: new Set(),
       isLoading: true,
     };
   }
@@ -44,12 +46,21 @@ class Dashboard extends React.Component {
   async componentDidMount() {
     window.scrollTo(0, 0);
     try {
-      const { totalCost, projNameToTotalCost, projNameToUserTotalCost, envNameToCostInfo } = await this.getCosts();
+      const {
+        totalCost,
+        projNameToTotalCost,
+        projNameToUserTotalCost,
+        envIdToCostInfo,
+        envIdToEnvMetadata,
+        duplicateEnvNames,
+      } = await this.getCosts();
       this.setState({
         totalCost,
         projNameToTotalCost,
         projNameToUserTotalCost,
-        envNameToCostInfo,
+        envIdToCostInfo,
+        envIdToEnvMetadata,
+        duplicateEnvNames,
         isLoading: false,
       });
     } catch (error) {
@@ -115,18 +126,17 @@ class Dashboard extends React.Component {
   }
 
   async getCosts() {
-    const { envNameToCostInfo, envNameToIndex } = await this.getAccumulatedEnvCost();
+    const { envIdToCostInfo, envIdToEnvMetadata, duplicateEnvNames } = await this.getAccumulatedEnvCost();
 
     const projNameToUserTotalCost = {};
-    Object.keys(envNameToCostInfo).forEach(envName => {
-      const projName = envNameToIndex[envName];
+    Object.keys(envIdToCostInfo).forEach(envId => {
+      const projName = envIdToEnvMetadata[envId].index;
       if (projNameToUserTotalCost[projName] === undefined) {
         projNameToUserTotalCost[projName] = {};
       }
-      Object.keys(envNameToCostInfo[envName].pastMonthCostByUser).forEach(user => {
-        const currentUserCost = _.get(projNameToUserTotalCost, `${projName}.${user}`, 0);
-        projNameToUserTotalCost[projName][user] =
-          currentUserCost + envNameToCostInfo[envName].pastMonthCostByUser[user];
+      Object.keys(envIdToCostInfo[envId].pastMonthCostByUser).forEach(user => {
+        const currentUserCost = _.get(projNameToUserTotalCost[projName], user, 0);
+        projNameToUserTotalCost[projName][user] = currentUserCost + envIdToCostInfo[envId].pastMonthCostByUser[user];
       });
     });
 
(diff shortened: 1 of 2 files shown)


SanketD92 (commented):
Could we add unit tests for these changes?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
```diff
@@ -115,18 +126,17 @@ class Dashboard extends React.Component {
...
 
   async getCosts() {
-    const { envNameToCostInfo, envNameToIndex } = await this.getAccumulatedEnvCost();
+    const { envIdToCostInfo, envIdToEnvMetadata, duplicateEnvNames } = await this.getAccumulatedEnvCost();
 
     const projNameToUserTotalCost = {};
-    Object.keys(envNameToCostInfo).forEach(envName => {
-      const projName = envNameToIndex[envName];
+    Object.keys(envIdToCostInfo).forEach(envId => {
+      const projName = envIdToEnvMetadata[envId].index;
```
Bingjiling: ProjectName is stored under key index?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
```diff
@@ -115,18 +126,17 @@ class Dashboard extends React.Component {
...
       if (projNameToUserTotalCost[projName] === undefined) {
         projNameToUserTotalCost[projName] = {};
       }
-      Object.keys(envNameToCostInfo[envName].pastMonthCostByUser).forEach(user => {
-        const currentUserCost = _.get(projNameToUserTotalCost, `${projName}.${user}`, 0);
-        projNameToUserTotalCost[projName][user] =
-          currentUserCost + envNameToCostInfo[envName].pastMonthCostByUser[user];
+      Object.keys(envIdToCostInfo[envId].pastMonthCostByUser).forEach(user => {
+        const currentUserCost = _.get(projNameToUserTotalCost[projName], user, 0);
```
Bingjiling: Emmmm I can't figure out why the previous way `_.get(projNameToUserTotalCost, `${projName}.${user}`, 0)` was not working, do you know?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
```diff
@@ -218,20 +243,26 @@ class Dashboard extends React.Component {
...
       return <ProgressPlaceHolder />;
     }
 
     const pastMonthCostTotalArray = [];
-    Object.keys(this.state.envNameToCostInfo).forEach(envName => {
+    Object.keys(this.state.envIdToCostInfo).forEach(envId => {
       let total = 0;
-      Object.keys(this.state.envNameToCostInfo[envName].pastMonthCostByUser).forEach(user => {
-        total += this.state.envNameToCostInfo[envName].pastMonthCostByUser[user];
+      Object.keys(this.state.envIdToCostInfo[envId].pastMonthCostByUser).forEach(user => {
```
Bingjiling: Would `_.sum(_.values(this.state.envIdToCostInfo[envId].pastMonthCostByUser))` work?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
```diff
@@ -242,13 +273,19 @@ class Dashboard extends React.Component {
...
   renderYesterdayCostPerEnv() {
-    if (_.isEmpty(this.state.envNameToCostInfo)) {
+    if (_.isEmpty(this.state.envIdToCostInfo)) {
       return <ProgressPlaceHolder />;
     }
     const title = "Yesterday's Env Cost";
-    const labels = Object.keys(this.state.envNameToCostInfo);
-    const dataPoints = Object.keys(this.state.envNameToCostInfo).map(envName => {
-      return this.state.envNameToCostInfo[envName].yesterdayCost;
+    const labels = Object.keys(this.state.envIdToCostInfo).map(envId => {
```
Bingjiling: This block looks the same as line 259 to 265, is there a way to write it only once?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
jn1119: I think it is a bad variable name. The entire file is using project name at all places, I can rename the entire file (around 48 usages)

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
jn1119: Not sure, I didn't dive deeper and assumed the syntax was off.

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
jn1119: Yeah, that works too. Will replace it

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
jn1119: Good point, I moved it to a different function and added a unit test for it too

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
jn1119: Actually I just realized it is because user contains a '.' since it is an email address

SanketD92 (approved):
LGTM

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/dashboard/Dashboard.js
Bingjiling: Ah! That makes sense!

Bingjiling (approved):
LGTM

PR #442, opened 2021-04-14
fix: Open data handler fix

Bingjiling:
Issue #, if available:

Description of changes:

Fix the following bugs:

Data validation bugs:
* Sha: Regex does not have * at the end, only matching one digit 
* Description: description for open data exceed the length requirement and have () <> and [] characters 

Study update bugs: 
* Update study is not updating DB with newly pulled open data, rather it’s using record just pulled from the DB by the find study method 
* The find study method has been updated as part of BYOB to add a status for a study, it’s not part of the update study schema and is failing the data validation 

Time out bug: 
* Timeout for Lambda was not set, it’s default to 6 seconds, the function can take up to 10 seconds sometime 

Permission bug: 
* Open data scrape handler role does not have permission for studyPermissions table, which is needed for update study 

WIP - Unit tests 

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
@@ -20,9 +20,7 @@
       "enum": ["My Studies", "Organization", "Open Data"]
     },
     "description": {
-      "type": "string",
-      "maxLength": 2048,
-      "pattern": "^([^<>{}]*)$"
+      "type": "string"
     },
     "projectId": {
       "type": "string",
@@ -36,7 +34,8 @@
     "sha": {
       "type": "string",
       "maxLength": 64,
-      "pattern": "^([A-Fa-f0-9])$"
+      "pattern": "^([A-Fa-f0-9]*)$",
+      "description": "A unique identifier for Open Data in MD5 hash, hexadecimal"
     },
     "resources": {
       "type": "array",
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
@@ -20,7 +20,8 @@
       "enum": ["My Studies", "Organization", "Open Data"]
     },
     "description": {
-      "type": "string"
+      "type": "string",
+      "description": "Leaving length and pattern blank to accommodate open data"
     },
     "projectId": {
       "type": "string",
@@ -34,7 +35,7 @@
     "sha": {
       "type": "string",
       "maxLength": 64,
-      "pattern": "^([A-Fa-f0-9]*)$",
+      "pattern": "^([A-Fa-f0-9]{40})$",
       "description": "A unique identifier for Open Data in MD5 hash, hexadecimal"
     },
     "resources": {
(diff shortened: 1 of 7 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
```diff
@@ -36,7 +34,8 @@
     "sha": {
       "type": "string",
       "maxLength": 64,
-      "pattern": "^([A-Fa-f0-9])$"
+      "pattern": "^([A-Fa-f0-9]*)$",
```
jn1119: Should we use `{40}` to lock down its length? I

main/solution/backend/src/lambdas/open-data-scrape-handler/handler-impl.js
```diff
@@ -183,24 +187,26 @@ const newHandler = async ({ studyService, log = consoleLogger } = {}) => {
...
+          const existingStudy = await studyService.find(userContext, study.id);
+          if (!existingStudy) {
+            await studyService.create(userContext, study);
+          } else {
+            await studyService.update(userContext, { rev: existingStudy.rev, ..._.omit(study, 'category') });
+          }
+          // Catch the err here so other open data update could continue
+        } catch (err) {
+          log.error(err);
+          log.error(study);
```
jn1119: Add some messaging here perhaps? `Study ${study} failed with error`...

main/solution/backend/src/lambdas/open-data-scrape-handler/handler-impl.js
```diff
@@ -183,24 +187,26 @@ const newHandler = async ({ studyService, log = consoleLogger } = {}) => {
...
-          await studyService.update(userContext, studyToUpdate);
+        try {
+          const existingStudy = await studyService.find(userContext, study.id);
+          if (!existingStudy) {
+            await studyService.create(userContext, study);
+          } else {
+            await studyService.update(userContext, { rev: existingStudy.rev, ..._.omit(study, 'category') });
+          }
+          // Catch the err here so other open data update could continue
+        } catch (err) {
```
jn1119: Can we add a unit test here which checks that Promise.all succeeds and doesn't throw an Exception if one of the underlying calls fail.

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
jn1119: Can we add a unit test for this?

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
```diff
@@ -17,6 +17,7 @@ const _ = require('lodash');
 const ServicesContainer = require('@aws-ee/base-services-container/lib/services-container');
 const JsonSchemaValidationService = require('@aws-ee/base-services/lib/json-schema-validation-service');
 const AwsService = require('@aws-ee/base-services/lib/aws/aws-service');
+const { getSystemRequestContext } = require('@aws-ee/base-services/lib/helpers/system-context');
```
SanketD92, line 20: Nit: Is system context really necessary in a unit test? Can't we just mock the authorization logic and proceed with the test?

Bingjiling: For open data create/update, study-service uses `isSystem` to check authorization, which was an imported function. It's not quite straightforward to mock compared to a in class `isAuthorized` function.

SanketD92 (approved):
LGTM

PR #470, opened 2021-05-03
chore: Lock environment installation file versions

nguyen102:
Issue #, if available:

Description of changes:
chore: Lock environment installtion file versions

I tested the new `bootstrap.sh` script on Sagemaker, EMR, EC2-linux, and RStudio. All four workspace types could read and write from mounted studies.

I built a new AMI with `provision-rstudio.sh` file and the RStudio instance was able to read and write from mounted studies.

Checklist:

AS review ticket id:

### DIFF
--- a/.github/workflows/unit-test-code-analysis.yml
+++ b/.github/workflows/unit-test-code-analysis.yml
@@ -16,6 +16,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
--- a/main/solution/machine-images/config/infra/provisioners/provision-rstudio.sh
+++ b/main/solution/machine-images/config/infra/provisioners/provision-rstudio.sh
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
 
 # Various development packages needed to compile R
-sudo yum install -y gcc gcc-gfortran gcc-c++
-sudo yum install -y java-1.8.0-openjdk-devel
-sudo yum install -y readline-devel zlib-devel bzip2-devel xz-devel pcre-devel
-sudo yum install -y libcurl-devel libpng-devel cairo-devel pango-devel
-sudo yum install -y xorg-x11-server-devel libX11-devel libXt-devel
+sudo yum install -y gcc-7.3.* gcc-gfortran-7.3.* gcc-c++-7.3.*
+sudo yum install -y java-1.8.0-openjdk-devel-1.8.0.*
+sudo yum install -y readline-devel-6.2 zlib-devel-1.2.* bzip2-devel-1.0.* xz-devel-5.2.* pcre-devel-8.32
+sudo yum install -y libcurl-devel-7.61.* libpng-devel-1.5.* cairo-devel-1.15.* pango-devel-1.42.*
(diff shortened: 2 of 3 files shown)


jn1119 (commented):
LGTM overall. Couple of questions

1. Have you verified the vendor guidance for the specific versions added here? Are they safe to use security/bug wise and are actively been supported?
2. According to codecov there is a branch coverage drop by more than one percent. Do you know why the check still passes? My assumption was that any drop more than [0.1 percent](https://github.com/awslabs/service-workbench-on-aws/blob/mainline/codecov.yml#L10) will fail the check.

.github/workflows/unit-test-code-analysis.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
```
nguyen102, line 19: This was added to fix an error when running CodeCov in GH Action.
Error is here
```
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS brotli GSS-API HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP UnixSockets
==> GitHub Actions detected.
    Env vars used:
      -> GITHUB_ACTIONS:    true
      -> GITHUB_HEAD_REF:   lock-bootstrap-version
      -> GITHUB_REF:        refs/pull/470/merge
      -> GITHUB_REPOSITORY: awslabs/service-workbench-on-aws
      -> GITHUB_RUN_ID:     807977121
      -> GITHUB_SHA:        f41ac82e088df80b6aea33b828b5bf8629ee34f7
      -> GITHUB_WORKFLOW:   Unit Tests & Code Analysis
->  Issue detecting commit SHA. Please run actions/checkout with fetch-depth > 1 or set to 0
```
https://github.com/awslabs/service-workbench-on-aws/pull/470/checks?check_run_id=2495242442

Recommended fix that was followed
https://github.com/codecov/codecov-action/issues/190#issuecomment-790729633

nguyen102:
@jn1119 

> Have you verified the vendor guidance for the specific versions added here? Are they safe to use security/bug wise and are actively been supported?

I looked into finding the security release notes of the software, and was unable to find security and vendor guidance for all the software. What I can confirm is that all of the software we're using and locking down, is the latest version that is available on YUM for our linux server. Since they're the latest version that is available, I think it would be safe to assume that all bug/security related issues that are known and solvable has been resolved.


> According to codecov there is a branch coverage drop by more than one percent. Do you know why the check still passes? My assumption was that any drop more than 0.1 percent will fail the check.

Looks like the check is now failing as expected. I think CodeCov is not analyzing this PR correctly. It might be because this PR only has changes in scripts, which isn't testable. The drop in coverage is an error and is probably a result of this
https://github.com/awslabs/service-workbench-on-aws/pull/470/checks?check_run_id=2520195837
```
 project root: .
    Yaml found at: codecov.yml
    -> Found 1 reports
==> Detecting git/mercurial file structure
==> Reading reports
    - file not found at ./coverage/coverage-final.json
--> No coverage data found.
```
The drop is a false alarm because CodeCov wasn't able to find `coverage-final.json`

jn1119 (approved):

PR #498, opened 2021-05-18
chore: Add GH action to check for dependency vulnerabilities 

nguyen102:
Issue #, if available:

Description of changes:
This adds a check during PRs that checks for any security vulnerabilities in our dependencies. The threshold for when the check would fail is set at greater than 2 for  "High severity" vulnerabilities or greater than 0 for "Critical severity" vulnerabilities. 

The limit is set at 2 because I was unable to bring the total lower than 2 as a part of this task.

This PR also include update to our dependencies that brought the number of High vulnerabilities from 7 to 2. 

**Testing Done**
* Successfully launched and access these workspaces
    * linux
    * emr
    * rstudio
    * sagemaker
    * windows
* Also able to mount study to workspace and edit the study
* Integration test passes

You can see an example of the check here
https://github.com/awslabs/service-workbench-on-aws/pull/498/checks?check_run_id=2621811483

Click into the section "Check dependencies for vulnerabilities"

Checklist:

AS review ticket id:


scripts/check-dependency-vulnerabilities.sh
```diff
@@ -0,0 +1,22 @@
...
+pnpm audit --json --audit-level high --prod > audit-results.json
+high_vul=$(cat audit-results.json | jq '.metadata.vulnerabilities.high')
+critical_vul=$(cat audit-results.json | jq '.metadata.vulnerabilities.critical')
+
+high_vul_threshold=2
+critical_vul_threshold=0
+printf "\nHigh vulnerabilities threshold set at $high_vul_threshold\n"
+echo "Critical vulnerabilities threshold set at $critical_vul_threshold"
+
+if (($high_vul > $high_vul_threshold)) || (($critical_vul > critical_vul_threshold)); then
```
SanketD92, line 15: PR description mentions "_fail is set at 2 or greater "High severity" vulnerabilities_". 
From the code, looks like it would only fail for greater than 2. Is that right?

nguyen102: Yep, you're right. It's greater than 2. I'll update that in the PR description.

SanketD92 (approved):
LGTM

PR #502, opened 2021-05-21
fix: add termination status for non-found workspaces

rsmayda:
Description of changes:

- In the situation that a instance from the AWS console is no longer able to be found from SWB but within SWB it is in a non-terminal state -> we want to put that workspace into "FAILED" status

Checklist:


addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-service.js
```diff
@@ -210,26 +222,16 @@ class EnvironmentScService extends Service {
...
-          sagemakerUpdated[key] = {
-            ddbID: existingEnvRecord.id,
-            currentStatus: expectedDDBStatus,
-            staleStatus: existingEnvRecord.status,
-          };
-        } catch (e) {
-          this.log.error(`Error updating record ${existingEnvRecord.id}`);
-          this.log.error(e);
-        }
+        updateStatusResult = await this.updateDDBStatus(requestContext, existingEnvRecord, expectedDDBStatus);
```
nguyen102: Looks like there is an opportunity to DRY the code between here and the `pollAndSyncEc2Status` method.

nguyen102 (approved):

PR #503, opened 2021-05-25
fix: Do not allow users to change root password

nguyen102:
Issue #, if available:

Description of changes:
* Do not allow users to change root password
* Allow post deployment lambda to change root user password by having the lambda use the function `saveRootPassword`

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-post-deployment/packages/base-post-deployment/lib/steps/create-root-user-service.js
+++ b/addons/addon-base-post-deployment/packages/base-post-deployment/lib/steps/create-root-user-service.js
@@ -60,9 +60,8 @@ class CreateRootUserService extends Service {
       });
       this.log.info('Created root user in the data lake');
 
-      await dbPasswordService.savePassword(getSystemRequestContext(), {
+      await dbPasswordService.saveRootPassword(getSystemRequestContext(), {
         uid: createdUser.uid,
-        username: rootUserName,
         password: rootUserPassword,
       });
       this.log.info("Created root user's password");
--- a/addons/addon-base/packages/services/lib/db-password/db-password-service.js
+++ b/addons/addon-base/packages/services/lib/db-password/db-password-service.js
@@ -19,6 +19,7 @@ const uuid = require('uuid/v4');
 const Service = require('@aws-ee/base-services-container/lib/service');
 const { ensureCurrentUserOrAdmin } = require('../authorization/assertions');
 const { runAndCatch } = require('../helpers/utils');
+const { isSystem } = require('../authorization/authorization-utils');
 
 const settingKeys = {
   tableName: 'dbPasswords',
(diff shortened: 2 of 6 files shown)


addons/addon-base-post-deployment/packages/base-post-deployment/lib/steps/create-root-user-service.js
```diff
@@ -60,9 +60,8 @@ class CreateRootUserService extends Service {
       });
       this.log.info('Created root user in the data lake');
 
-      await dbPasswordService.savePassword(getSystemRequestContext(), {
+      await dbPasswordService.saveRootPassword(getSystemRequestContext(), {
         uid: createdUser.uid,
-        username: rootUserName,
```
jn1119, line 65: It seems odd that we need to pass `root` username when the method itself is called `saveRootPassword`

addons/addon-base/packages/services/lib/db-password/db-password-service.js
```diff
@@ -45,10 +46,27 @@ class DbPasswordService extends Service {
     }
   }
 
+  async saveRootPassword(requestContext, { password, uid }) {
+    const username = this.settings.get('rootUserName');
+    if (!isSystem(requestContext)) {
+      throw this.boom.badRequest("'root' password can not be changed", true);
```
jn1119: `cannot be changed by non-system user`?

addons/addon-base-post-deployment/packages/base-post-deployment/lib/steps/create-root-user-service.js
nguyen102: We're not passing the `root` username right? The `root` username is being deleted. We're just passing in the `rootUserPassword`

main/solution/backend/config/infra/functions.yml
```diff
@@ -156,6 +156,7 @@ apiHandler:
     APP_ENV_TYPE_CONFIGS_BUCKET_NAME: ${self:custom.settings.envTypeConfigsBucketName}
     APP_ENV_BOOTSTRAP_BUCKET_NAME: ${self:custom.settings.environmentsBootstrapBucketName}
     APP_CUSTOM_USER_AGENT: ${self:custom.settings.customUserAgent}
+    APP_ROOT_USER_NAME: ${self:custom.settings.rootUserName}
```
jn1119, line 159: Where do we use `APP_ROOT_USER_NAME `?

jn1119 (commented):
LGTM overall (just some minor comments)

addons/addon-base-post-deployment/packages/base-post-deployment/lib/steps/create-root-user-service.js
jn1119: oops sorry, makes sense :)

addons/addon-base/packages/services/lib/db-password/db-password-service.js
nguyen102: For now it's only being used by the post deployment lambda, which is a system user. Therefore let's restrict it to only `systems` and reevaluate if we need to expand the scope in the future.

main/solution/backend/config/infra/functions.yml
nguyen102: We use it [here](https://github.com/awslabs/service-workbench-on-aws/blob/27523aee8b40670d169154d75bcc681aa079d29d/addons/addon-base/packages/services/lib/db-password/db-password-service.js#L50) to grab the root username. We also use it on line 62 in the same file.

addons/addon-base/packages/services/lib/db-password/db-password-service.js
jn1119: no, I mean the error/exception message can call out explicitly the error :)

main/solution/backend/config/infra/functions.yml
jn1119: Got it, I keep forgetting that the convention changes to camel case without APP

hatimkhan: Hi all, I happened to see this and wanted to give you a little bit of context around why we have APP_
These are unix environment variables that are passed to the lambda node.js unix process.  We add APP_ so that our environment variables don't collide with any ones provided by lambda.

jn1119 (approved):

PR #504, opened 2021-05-27
[UpdateUserPermission Feature] AWS Accounts UI Update

ahl27:
Issue #, if available: [GALI-929](https://sim.amazon.com/issues/GALI-929)

Description of changes:

_This feature is still a WIP_

Implements the UI redesign for the auto-update user permissions feature ([GALI-424](https://sim.amazon.com/issues/b7644135-9856-45f5-b99d-152a8ed41bb2)). The following items have been changed:

- AWS Accounts page has been updated to list accounts as cards (consistent with how workspaces and studies are listed). Account details are shown in a collapsible table. Previous Budget Detail functionality is preserved. Accounts are sorted alphabetically.
- Each Account card shows an indicator on if its permissions are up-to-date or need to be updated. New accounts that need to be onboarded are shown as 'New'. This indicator is pulled from the backend AwsAccounts DDB (see next bullet)
- AwsAccounts DDB now holds an additional `needsPermissionUpdate` value, which is a `maybe(types.Boolean)`. New accounts are created without this variable, and the to-be-implemented `checkPermissions` feature will update the value with `true/false` appropriately. Account card reads `undefined` as a new account, `true` as needing to be updated, and `false` as being up-to-date.
- AWS Accounts page can filter based on permissions status. New `filter` method added to the AwsAccountsStore, and unit test added for this method (identical functionality to the same method implemented for workspaces).
- AwsAccounts had an `update` method that was already implemented, but it wasn't hooked up to an API call. New API PUT call was added to update an AwsAccount, and the schema for the call was updated. New entry for `update-account` was added to `openapi.yaml`. This API call isn't used in the code currently, but will be utilized in the future to update `needsPermissionUpdate` value as well as to implement an Edit Account Details button.
- New button to check permissions status was added. No functionality yet.
- For accounts that or new or out of date, buttons show to update permissions or onboard the account. No functionality yet.
- Quick bash script added to redeploy the UI in one step. Happy to delete this prior to merging into the main develop branch.

Work to be completed in subsequent PRs:
- Check Permissions button has no functionality. New API GET call for this will be added in the future, and `update-account` API call will be used to update the backend.
- Logic to select accounts on the accounts page is implemented, but the checkbox has been temporarily removed. Box will be added back when the Edit Account and Delete Account buttons and functionality are added.
- E2E tests will be added after functionality for checking permissions is added.
- Integration tests will be added when more API calls are added in the future. `update-account` API call was implemented but hasn't been used for anything yet, so I can't write an integ test for it yet. This will also rectify the areas codecov has flagged as having decreased coverage.
- Update Permissions/Onboard Account buttons have no functionality yet.
- AwsAccountsStore has a `checkPermissions` method that currently just flips the value of `needsPermissionUpdate`. Once the backend API call to check permissions is implemented, this method will be adjusted. This method is currently inaccessible via the UI or the API. I have a unit test for the current functionality, and when I fix it later with the new API call I'll update the unit test accordingly.

Checklist:

AS review ticket id:


ahl27:
https://user-images.githubusercontent.com/30053966/119838532-7b3d7680-bed1-11eb-9d2d-85519aff2312.mov

Video showing the new UI design

addons/addon-base-raas-ui/packages/base-raas-ui/src/helpers/api.js
```diff
@@ -60,6 +60,11 @@ function addIndex(index) {
   return httpApiPost('api/indexes', { data: index });
 }
 
+// Note the accountUUID used here is the 'id' column in dbAwsAccounts table and 'id' attribute in AwsAccount.js, not AWS account id
+function updateAwsAccount(accountUUID, data) {
```
Bingjiling, line 60: Move this method to be together with other AwsAccount methods?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
```diff
@@ -53,6 +54,13 @@ const AwsAccount = types
       self.updatedAt = rawAwsAccounts.updatedAt || self.updatedAt;
       self.createdBy = rawAwsAccounts.createdBy || self.createdBy;
       self.updatedBy = rawAwsAccounts.updatedBy || self.updatedBy;
+      self.needsPermissionUpdate =
```
Bingjiling: Is the second check `typeof self.needsPermissionUpdate === 'boolean'` needed? If `self.needsPermissionUpdate` is not a boolean, is it `undefined`?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
```diff
@@ -13,14 +13,30 @@
...
+import { getAwsAccounts, addAwsAccount, createAwsAccount, updateAwsAccount } from '../../helpers/api';
 import { AwsAccount } from './AwsAccount';
 import { BudgetStore } from './BudgetStore';
 import Budget from './Budget';
 
+const filterNames = {
+  ALL: 'All',
+  CURRENT: 'Up-to-Date',
+  UPDATEME: 'Needs Update',
+  NEW: 'New',
```
Bingjiling: Does NEW mean PENDING here?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
```diff
@@ -70,6 +86,21 @@ const AwsAccountsStore = BaseStore.named('AwsAccountsStore')
...
+      checkPermissions: async () => {
+        // This is a placeholder function that just switches the needsPermissionUpdate to its opposite value
+        // Will be implemented later
+        const awsAccounts = (await getAwsAccounts()) || [];
+        self.runInAction(() => {
+          awsAccounts.forEach(account => {
+            const newStatus = account.needsPermissionUpdate === undefined ? false : !account.needsPermissionUpdate;
+            if (account.needsPermissionUpdate !== newStatus) {
+              const newAccount = { ...account, needsPermissionUpdate: newStatus };
+              updateAwsAccount(newAccount.id, newAccount);
```
Bingjiling: Emmm I'm not getting what this method is doing. `newStatus` is derived from `account.needsPermissionUpdate`, and then we're comparing `newStatus` against `account.needsPermissionUpdate` afterwards?

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-aws-accounts.json
```diff
@@ -49,6 +49,21 @@
     "encryptionKeyArn": {
       "type": "string",
       "pattern": "^arn:aws:kms:.*$"
+    },
+    "needsPermissionUpdate": {
+      "type": "boolean"
+    },
+    "createdAt": {
```
Bingjiling: These four fields are auto generated by DDB, it should not be passed in from the call. (e.g. the createdBy field could be used for audit, and we don't want ppl to change that value as they wish) For update API method, only the fields to be updated and the `required` fields in the schema need to be passed in.

openapi.yaml
```diff
@@ -2921,6 +2940,29 @@ components:
...
           example: 'My test account'
+    UpdateAwsAccount:
+      required:
+        - name
+        - accountId
+        - roleArn
+        - externalId
+      type: object
+      properties:
+        name:
```
Bingjiling, line 2951: Properties not required need to be included in the schema as well.

scripts/redeploy-ui.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
Bingjiling: Is this for redeploying UI locally? You can run `pnpm run build:watch` in `addons/addon-base-raas-ui/packages/base-raas-ui` and keep the terminal up to get local changes reflected in local deployment in real time.

Bingjiling (commented):
Great work!!

addons/addon-base-raas-ui/packages/base-raas-ui/src/helpers/api.js
ahl27: sure thing

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
ahl27: It does not, NEW means an account that has been added to SWB but that hasn't been onboarded yet. PENDING is a label I should probably add, I was just thinking about that today.

ahl27: For this PR I'm not sure if it's necessary, since none of the functionality makes use of a PENDING tag, but when the check permissions button is implemented I think a pending tag and functionality would be useful.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
ahl27: 😅 This method isn't doing anything, it's a placeholder method for testing that's going to be removed in the next update (adding logic behind check permissions button). It's not linked up to anything and has a comment indicating it, but if you think it's better I can remove it and replace it with just a `return undefined`.

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-aws-accounts.json
ahl27: gotcha, thank you! I'll remove those.

scripts/redeploy-ui.sh
ahl27: It's actually for redeploying the UI non-locally, since at the time I wasn't using the local environment. I can remove this file, I had planned to remove it prior to merge to mainline anyway.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
ahl27: Yes, I was envisioning the flow to be:
- user is added to SWB, `needsPermissionUpdate === undefined` since it hasn't been onboarded yet
- user is onboarded, `needsPermissionUpdate` is set to either `true` or `false`
- once onboarded, `needsPermissionUpdate` should only ever be a boolean

ahl27: I wasn't sure if there was a better way to implement this...
The other maybe values used a double check for permissions and I was emulating those, but it may be simpler to skip the second check and just default to `self.needsPermissionUpdate`, since if it's missing it'll be equal to `undefined` anyway. I'll make that change.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
ahl27: the `const newStatus = ...` line is implemented that way to simulate the permissions status changing. When the backend for permissions status checking is added, that line will be replaced with something that calls the `getAwsPermissionsStatus` API and stores it in `newStatus`.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
Bingjiling: Yes, the flow makes sense. BYOB account go through the same flow as well, might be worth checking how it's handled there. Normally we want to avoid using `undefined` or `null` for an actual meaningful status, since it's confusing for readers and it could be the result of an unintended event(exception or something else).

openapi.yaml
ahl27: 👍🏻

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
ahl27: gotcha. Maybe it would be better to just use a string arg with predefined entries

ahl27: I looked at BYOB and made edits based on how theirs is implemented--`needsPermissionUpdate` is now strictly a boolean, and checking for onboarding will look at the `cfnStackName` parameter, which will be created and added when the account is onboarded.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
Bingjiling: Got it. Yea, we can further define the status when the functionality is ready. NEW is a bit confusing to be used as a status, maybe we can use something like `ONBOARDME : 'Needs Onboard'`?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
Bingjiling: Yes! Return undefined would be great! I saw the comment and assumed you implemented the method afterwards and forgot to remove the comment 😅

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
ahl27: Yep, that makes sense. I made that change just now

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
ahl27: haha okay, just fixed that and modified the unit test

ahl27:
removing the half-implemented function resulted in codecov going down by 0.01%--most of the testing won't be implementable until later functionality is as well, but expect that codecov will have a net coverage of at least +0.02% with the next PR (and a net positive for the final merge)

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-aws-accounts.json
```diff
@@ -26,6 +26,13 @@
       "type": "string",
       "pattern": "^[0-9]{12}$"
     },
+    "cfnStackName": {
+      "type": "string",
+      "maxLength": 255
+    },
+    "mainRegion:": {
```
Bingjiling, line 33: Do we allow the member account region to be different from main account? Not sure if region customization is supported right now...

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
```diff
@@ -27,6 +27,9 @@ const AwsAccount = types
     description: '',
     accountId: '',
     externalId: '',
+    needsPermissionUpdate: true,
```
Bingjiling, line 30: Is there a `types.boolean` ?

Bingjiling (commented):
Looks good! Two minor questions:

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-aws-accounts.json
ahl27: thats a good question that I hadn't thought of, I can find out

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
ahl27: there is, although setting it to `true` here has the same effect with a default value of `true`.

Bingjiling (approved):
LGTM! We can get the region removed and update the account status to a string as discussed in the next PR.

PR #515, opened 2021-06-07
fix: Allow sagemaker to have the proper IAM permission to autostop itself

nguyen102:
Issue #, if available:

Description of changes:

* Update permission boundary for Sagemaker instance to allow stopping a Sagemaker instance
* Change Sagemaker to check whether idle limit has passed, on a minute by minute cadence instead of an hourly cadence
* Update `envStatusPollHandler` to run once every minute instead of once every 3 minutes. (More accurate, since users can choose to have Sagemaker stop itself after 1 minute of inactivity)

I tested that Sagemaker was able to stop itself after inactivity. I saw the Sagemaker instance was stopped on the AWS Console and within SWB UI.

Checklist:

- [x] Is this change also required on the AWS Solution version?

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/sagemaker-notebook-instance.cfn.yml
+++ b/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/sagemaker-notebook-instance.cfn.yml
@@ -180,7 +180,7 @@ Resources:
                 chmod a+x /usr/local/bin/autostop.py
                 IDLE_TIME=`expr ${AutoStopIdleTimeInMinutes} \* 60`
                 echo "Starting the SageMaker autostop script in cron"
-                (crontab -l 2>/dev/null; echo "5 * * * * /usr/bin/python /usr/local/bin/autostop.py --time $IDLE_TIME --ignore-connections > /var/log/autostop.log") | crontab -
+                (crontab -l 2>/dev/null; echo "*/5 * * * * /usr/bin/python /usr/local/bin/autostop.py --time $IDLE_TIME --ignore-connections >> /var/log/autostop.log") | crontab -
               fi
 
 Outputs:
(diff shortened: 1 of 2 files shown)


jn1119 (approved):

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/sagemaker-notebook-instance.cfn.yml
```diff
@@ -73,6 +73,11 @@ Resources:
             Action:
               - 'sts:AssumeRole'
             Resource: 'arn:aws:iam::*:role/swb-*'
+          - Effect: Allow
+            Action:
+              - sagemaker:DescribeNotebookInstance
+              - sagemaker:StopNotebookInstance
+            Resource: '*'
```
jn1119, line 80: Can we restrict the resource here?

jn1119: Can the reference be `- !Ref BasicNotebookInstance`?

nguyen102: I don't think we can reference it because the notebook is created after this policy is created. More info [here](https://github.com/awslabs/service-workbench-on-aws/blob/efcb8d80a44f30ed7af249e36428a0de4379a4b6/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/sagemaker-notebook-instance.cfn.yml#L136). Also since this is a permission boundary I think it's ok to have a broader resource section here. In the actual Sagemaker IAM role, we do scope the permission down to just the notebook.

jn1119: Yeah, that's what I was trying to understand. If the IAM policy can reference the instance then why can't permission boundary policy do the same.

>>I think it's ok to have a broader resource section here. In the actual Sagemaker IAM role, we do scope the permission down to just the notebook.

Yeah, that's why I approved the change. But I am still curious if it could have been done just like we did with IAM policy

PR #523, opened 2021-06-09
feat: App Stream vpc

nguyen102:
Issue #, if available:

Description of changes:

Updated onboard CFN to allow for setting up AppStream.

A parameter is provided to the customer called `EnableAppStream`. Based on their response, the template either spins up the default resources or the AppStream resources (private subnets, vpc endpoints, gateway endpoints).

**Testing**
Setup: I spun up an AppStream instance in the private AppStream subnet. I spun up a Sagemaker instance and an Ec2 instance in the private Workspace Subnet.

When updating the `onboard` CF stack from it's original CF stack to the `AppStream` stack, the Public subnet gets deleted and the private subnets get added. Below is a screenshot of what the Changeset looks like.

Screen Shot 2021-06-10 at 10 39 54 AM

* Ec2 instance can't be accessed from internet
    * Ec2 instance can access s3 bucket
* App Stream i
    * App Stream can't access internet
    * App Stream can access ec2 instance
* Sagemaker instance is accessible from AppStream after presigned URL is created

Checklist:

AS review ticket id:


nguyen102:
Test: When an Old VPC with an EC2 instance is being updated to a new VPC with private subnets, the update fails. This is expected, and will prevent users from updating to AppStream private subnets if they still have resources in the old subnet. 

Screen Shot 2021-06-10 at 3 47 42 PM

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
```diff
@@ -408,6 +430,111 @@ Resources:
...
+    Properties:
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal: '*'
+            Action:
+              - 's3:GetObject'
+              - 's3:ListBucket'
+              - 's3:ListAllMyBuckets'
```
jn1119: We would also need write access along with some other actions. The entire list should be here: https://github.com/awslabs/service-workbench-on-aws/blob/mainline/addons/addon-base-raas/packages/base-raas-services/lib/helpers/iam/study-policy.js#L98-L113

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
```diff
@@ -28,12 +33,18 @@ Parameters:
     Type: String
     Description: The arn of workflowRunner role
 
-  # Generous subnet allocation of 8192 addresses (ie room for a few concurrent EMR clusters)
-  # ending at 10.0.31.255
-  VpcPublicSubnet1Cidr:
```
jn1119, line 33: I think we shouldn't change anything about the account if we don't enable AppStream. Would this try deleting the old subnet or modifying it? Can you double check that existing account remains unchanged.

nguyen102: Good point. Updated to keep the CIDR the same for public subnet. 
I checked that the new template keeps the existing account unchanged if AppStream is disabled, and it does.

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
nguyen102: updated

jn1119 (approved):

PR #528, opened 2021-06-11
[UpdateUserPermissions] Automatic Permissions Check

ahl27:
Issue #, if available: GALI-930

Description of changes:

Adds functionality to check permissions for user accounts.

- permissions are checked automatically whenever account information is pulled (in the AwsAccountsStore Heartbeat), also automatically updates the backend DB if necessary
- Adds new API call to check CFN template for each account against expected CFN permissions template
- Slight modifications to UI (more filters added, check permissions button removed, placeholder button to update CFN name)
- Adds new entries to `openapi.yml`
- schema for `update-aws-account` updated
- Adds unit tests
  - I didn't end up using the `AwsMock` because I had already finished these tests by the time we figured out the bugfix, but I can change them if necessary. Future tests will be implemented with `AwsMock`.
- Adds new permissions to the user CFN template
  - `getTemplate` and `describeStacks` cloudformation permissions added for stacks matching `SWB-*`
  - the `SWB-*` pattern is mainly a placeholder for now, I'm going to revisit this when I implement later tasks in this feature (input CFN name, onboard account). I wanted to avoid adding universal permissions to all stacks on the account if possible.
- `permissionsStatus` changed to a string with more possible values

I kept in the same comparison as before (removing whitespace and comments, then hashing). Users that are adding services to SWB will have to change the template anyway, and given that the difference in implementation is relatively minor to support users being able to modify the template I thought this approach was best. If we hash the entire file, we introduce an issue where users could change the template without changing any permissions, which makes it impossible to update the stack (since CFN will say there are no changes and not allow updates). All external libraries that diff CFN stacks generate ChangeSets, which isn't ideal for a regular check on the CFN stack (and requires additional permissions).

Checklist:

- [x ] Have you successfully deployed to an AWS account with your changes?
- [x ] Have you written new tests for your core changes, as applicable?
- [x ] Have you successfully tested with your changes locally?
- [x ] Have you updated openapi.yaml if you made updates to API definition (including add, delete or update parameter and request data schema)?

AS review ticket id:


addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
```diff
@@ -48,12 +47,8 @@ const AwsAccount = types
       self.description = rawAwsAccounts.description || self.description;
       self.accountId = rawAwsAccounts.accountId || rawAwsAccounts.accountId;
       self.externalId = rawAwsAccounts.externalId || self.externalId;
-      self.needsPermissionUpdate =
-        rawAwsAccounts.needsPermissionUpdate !== undefined
-          ? rawAwsAccounts.needsPermissionUpdate
-          : self.needsPermissionUpdate;
+      self.permissionStatus = rawAwsAccounts.permissionStatus || self.permissionStatus || 'ERROR';
```
Bingjiling, line 102: Would this default newly added account to 'ERROR'?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
```diff
@@ -26,16 +32,18 @@ const filterNames = {
...
 };
 
 // A map, with the key being the filter name and the value being the function that will be used to filter the workspace
 // cfnStackName is an empty string if the account hasn't been onboarded yet
 const filters = {
   [filterNames.ALL]: () => true,
-  [filterNames.CURRENT]: account => account.needsPermissionUpdate === false && account.cfnStackName !== '',
-  [filterNames.UPDATEME]: account => account.needsPermissionUpdate === true && account.cfnStackName !== '',
-  [filterNames.ONBOARDME]: account => account.cfnStackName === '',
+  [filterNames.CURRENT]: account => account.cfnStackName !== '' && account.permissionStatus === 'CURRENT',
```
Bingjiling, line 43: What's the difference between account.cfnStackName === '' and account.permissionStatus === 'NOSTACKNAME'?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
```diff
@@ -26,16 +32,18 @@ const filterNames = {
...
+  [filterNames.UPDATEME]: account => account.cfnStackName !== '' && account.permissionStatus === 'NEEDSUPDATE',
+  [filterNames.NEW]: account => account.permissionStatus === 'NEEDSONBOARD',
```
Bingjiling, line 45: How does the backend differentiate between 'NEEDSONBOARD' and 'NOSTACKNAME'?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/__tests__/AwsAccountsStore.test.js
```diff
@@ -88,20 +91,15 @@ describe('AwsAccountsStore', () => {
...
 
-  describe('checkPermissions', () => {
-    it('should try to check the permissions status (WIP)', async () => {
-      // This implementation will change in the future, this is just a stopgap unit test
-      // It'll be changed when the function is fully implemented
-
-      // BUILD
-      getAwsAccounts.mockResolvedValue([newAwsAccount]);
+  describe('updateAccount', () => {
+    it('should fail to update account due to validation errors', async () => {
```
Bingjiling: Emmm I'm not sure how the 'should fail' part is tested here....

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
```diff
@@ -27,9 +27,8 @@ const AwsAccount = types
     description: '',
     accountId: '',
     externalId: '',
-    needsPermissionUpdate: true,
+    permissionStatus: '',
```
Bingjiling, line 82: Possible status for AwsAccount should be defined in this class, see example here -> https://github.com/awslabs/service-workbench-on-aws/blob/mainline/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/data-sources/DataSourceAccount.js

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
```diff
@@ -199,20 +210,16 @@ class AccountCard extends React.Component {
...
 
   renderUpdatePermsButton() {
-    const needsUpdate = this.needsUpdate;
+    const permissionStatus = this.permissionStatus;
     const buttonArgs =
-      needsUpdate === true
-        ? { message: 'Update Permissions', color: 'orange' }
-        : { message: 'Onboard Account', color: 'purple' };
-    // This button is only displayed if needsUpdate is either True or undefined
+      permissionStatus === 'NEEDSUPDATE'
```
Bingjiling: I would suggest using if/else or switch if there's two layers of ? : for better readability

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
```diff
@@ -287,6 +287,11 @@ Resources:
               - ssm:GetParameter
             Resource:
               - !Sub 'arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/*'
+          - Effect: Allow
+            Action:
+              - cloudformation:DescribeStacks
+              - cloudformation:GetTemplate
+            Resource: !Sub 'arn:aws:cloudformation:${AWS::Region}:${AWS::AccountId}:stack/SWB-*'
```
Bingjiling, line 294: Is this naming convention backward compatible with existing onboard stacks?

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/aws-accounts-controller.js
```diff
@@ -36,6 +37,19 @@ async function configure(context) {
...
 
+  // ===============================================================
+  //  GET /permissions (mounted to /api/aws-accounts)
+  // ===============================================================
+  router.get(
+    '/permissions',
+    wrap(async (req, res) => {
+      const requestContext = res.locals.requestContext;
+      const accountsList = await awsAccountsService.list();
+      const result = await awsCfnService.batchCheckAccountPermissions(requestContext, accountsList);
```
Bingjiling: The call to awsAccountsService.list() should be part of awsCfnService instead of the router, to maintain separation of responsibility.

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/aws-accounts-controller.js
```diff
@@ -96,6 +110,23 @@ async function configure(context) {
...
+  // ===============================================================
+  //  GET /:id/permissions (mounted to /api/aws-accounts)
+  // ===============================================================
+  router.get(
+    '/:id/permissions',
+    wrap(async (req, res) => {
+      const requestContext = res.locals.requestContext;
+      const accountId = req.params.id;
+
+      const accountDetails = await awsAccountsService.mustFind(requestContext, { id: accountId });
```
Bingjiling: Same as above.

Bingjiling (commented):
Just finished looking at UI part, will continue on the backend part. Try creating smaller PRs will leads to quicker and better reviews 👍

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-cfn-service.test.js
```diff
@@ -0,0 +1,231 @@
...
+const AwsService = require('@aws-ee/base-services/lib/aws/aws-service');
+
+jest.mock('../aws-accounts-service');
+const AwsAccountsServiceMock = require('../aws-accounts-service');
+
+jest.mock('../../cfn-templates/cfn-template-service');
+const CfnTemplateMock = require('../../cfn-templates/cfn-template-service');
+const AwsCfnService = require('../aws-cfn-service');
+
+describe('AwsAccountService', () => {
```
Bingjiling, line 51: 👍

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -0,0 +1,203 @@
...
+   * { stackId, templateId, templateVer, at }
+   *
+   * An exception is thrown if an error occurs while trying to describe the stack. This could happen if the stack
+   * is not created yet or is not provisioned in the correct account and region or was provisioned but did not
+   * use the correct stack name.
+   *
+   * @param requestContext
+   * @param accountEntity
+   */
+  async getStackTemplate(requestContext, accountEntity) {
```
Bingjiling, line 60: Is this a private method as well?

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -0,0 +1,203 @@
...
+    const stack = _.find(_.get(stacks, 'Stacks', []), item => item.StackName === cfnStackName);
+
+    if (_.isEmpty(stack)) {
+      throw this.boom.notFound(`Stack '${cfnStackName}' not found`, true);
+    }
+
+    const permissionsTemplateRaw = await cfnApi.getTemplate(params).promise();
+
+    return {
+      permissionsTemplateStr: permissionsTemplateRaw.TemplateBody,
```
Bingjiling: Is there a reason for wrapping the return in an object?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
ahl27: I guess as it's implemented now, yes...this part just means if we pass an account to the store that doesn't have an entry for permissionStatus, it'll default to 'ERROR'. Ideally in the account creation/onboarding step we'll default to 'NEEDSONBOARD', so the situation where we would ever reach the default value in this code block shouldn't be reached through normal use.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
ahl27: Accounts will be initialized with no stack name and account.permissionStatus = 'NEEDSONBOARD'. Once we have onboarded the account, if we end up with an empty stack name then the permissionStatus value is set to 'NOSTACKNAME'. Essentially the account.cfnStackName=== '' covers both the 'NEEDSONBOARD' and 'NOSTACKNAME' cases.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
ahl27: The backend doesn't really, since no account that exists in the DB should be being set to 'NEEDSONBOARD' from another value. My plan for implementation is to have accounts added to SWB be initialized to 'NEEDSONBOARD', but aside from that we shouldn't have an onboarded account that is suddenly becoming not onboarded.

If the stackname is missing, then the backend reads it as 'NEEDSONBOARD' if that's what the permissionsStatus is already set to, and 'NOSTACKNAME' otherwise.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/__tests__/AwsAccountsStore.test.js
ahl27: whoops...I'm not sure how I missed this...this is testing something haha, just not the thing I wrote down. I'll fix this.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
ahl27: 👍🏻

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
ahl27: will do!

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
ahl27: Not yet, I haven't worked out what the naming convention will be, I'm pretty sure that the existing stacks are prefixed with 'swb-', but I'm not completely sure. This part was mainly a placeholder for testing purposes for now, it's on my list of things to fix prior to merge to develop but I can take a look at it as part of this PR too if you'd rather it not wait. I need to take a deeper look at the naming of the CFN stacks as part of implementing the onboarding process itself.

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/aws-accounts-controller.js
ahl27: okay, that makes sense. I was basing it off the /api/aws-accounts call, but now that I'm looking at it it's a bit of a different scenario. I'll make that change with the next commit.

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/aws-accounts-controller.js
ahl27: 👍🏻 sounds good

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: uh...I think so, let me take a look.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: I guess not presently, my thought was that if we needed to query additional information later on it could be better to have it wrapped in an object because we could just add additional entries. Now that I think about it, I think if we're pulling other information that should probably be its own method, so I'll remove the object wrapping here and have it just remove the string.

ahl27 (commented):
i made this review by accident and i dont know how to get rid of it.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
ahl27: This is probably a larger design question, I'll follow up with you on it separately.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
```diff
@@ -66,6 +118,13 @@ const AwsAccount = types
   // eslint-disable-next-line no-unused-vars
   .views(self => ({
     // add view methods here
+    get permissionStatusDetail() {
+      console.log('IAMHERE');
```
Bingjiling: Nit: Remove it or change to .debug()

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
```diff
@@ -13,9 +13,61 @@
  *  permissions and limitations under the License.
  */
 
+import _ from 'lodash';
 import { types } from 'mobx-state-tree';
 import Budget from './Budget';
 
+const states = [
```
Bingjiling, line 20: Nice!

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -77,12 +76,13 @@ class AwsCfnService extends Service {
...
 
-    return {
-      permissionsTemplateStr: permissionsTemplateRaw.TemplateBody,
-    };
+    return permissionsTemplateRaw.TemplateBody;
   }
 
-  async checkAccountPermissions(requestContext, accountEntity) {
+  async checkAccountPermissions(requestContext, accountId) {
+    const awsAccountsService = await this.service('awsAccountsService');
```
Bingjiling: Consider moving the assertAuthorized call to the beginning of checkAccountPermissions and batchCheckAccountPermissions methods? Adding authorization to the service method call from the router makes it clearer to reader where/how authorization is done.

Bingjiling (dismissed):
Looks good! Some nit picks:

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
ahl27: ah crap...this isn't supposed to be in here at all. Good catch, thank you--I thought I had removed all of these but I guess one slipped through the cracks

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: Sure, that makes sense!

Bingjiling (approved):
LGTM

PR #539, opened 2021-06-17
feat: Extend account onboarding template to include AppStream resources

jn1119:
Issue #, if available: GALI-914

Description of changes:

* Extended onboarding template to include AppStream Security Group, Fleet, and Stack.
* Also updated SageMaker endpoint to include a specific Security Group instead of using the default one. Note that this is required since we will only need SageMaker access for AppStream so same set of rules cannot apply across all the interface endpoints since all other endpoints will be accessed via workspaces.
* Added Security Group rules to ensure that AppStream instances can connect to SageMaker

Testing

* Verified the template with and without AppStream enabled in a member account
* Also verified that AppStream instances were able to connect via SageMaker. Note that I didn't verify EC2 access since that was already verified in the VPC setup PR

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
+++ b/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
@@ -40,6 +40,17 @@ Parameters:
     Type: String
     Default: 10.0.0.0/19
 
+  LaunchConstraintRolePrefix:
+    Description: Role name prefix to use when creating a launch constraint role in the on-boarded account
+    Type: String
+    Default: '*'
+
+  LaunchConstraintPolicyPrefix:
+    Description: Customer managed policy name prefix to use when creating a launch constraint role in the on-boarded account
+    Type: String
+    Default: '*'
+
+  #------------AppStream Parameters Below-------
   # Range from 10.0.32.0 to 10.0.63.255
   AppStreamSubnetCidr:
     Description: Please enter the IP range (CIDR notation) for the the AppStream subnet. This value is only used if AppStream is enabled.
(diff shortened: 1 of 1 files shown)


addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
```diff
@@ -552,7 +586,100 @@ Resources:
...
       VpcId: !Ref VPC
+      SecurityGroupEgress:
+        - CidrIp: 127.0.0.1/32
+          IpProtocol: '-1'
+
+  AppStreamSecurityGroupEgress:
+    Type: AWS::EC2::SecurityGroupEgress
+    Condition: isAppStream
+    Properties:
+      GroupId: !Ref AppStreamSecurityGroup
```
nguyen102, line 619: How does `GroupId` and `DestinationSecurityGroupId` work? I notice we reference different value for these two fields.

nguyen102 (commented):
Overall looks good. Btw, why do we only set up SG for Sagemaker? Do we need SG for ec2, emr, and RStudio as well? 

The workspace we're launching are in the private VPCs that we created. Do we actually need SGs for them, since no resource outside the VPC can access these workspace anyways. Unless the reason for the SG is to prevent workspaces from talking to each other.

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
jn1119: We are setting egress from `GroupId` to `DestinationSecurityGroupId`. Note that this is a different `Type` than standard SecurityGroup. We are allowing AppStream to connect with SageMaker

jn1119:
> Overall looks good. Btw, why do we only set up SG for Sagemaker? Do we need SG for ec2, emr, and RStudio as well?
> 
> The workspace we're launching are in the private VPCs that we created. Do we actually need SGs for them, since no resource outside the VPC can access these workspace anyways. Unless the reason for the SG is to prevent workspaces from talking to each other.

I still need to verify the path for EMR and RStudio but each EC2 instance would have its own security group and the connectivity from AppStream instances won't be via interface endpoint- it would be via SSH. At that point EC2 instance security group and AppStream security group would need to go through similar egress/ingress setup. 

We don't want our workspaces unnecessary access to SageMaker interface endpoint. That can potentially mean that they have outside access to some arbitrary SageMaker notebook that we don't know about. Therefore, we have a different SG for SageMaker so that we can only give AppStream instances the access.

nguyen102 (approved):

PR #547, opened 2021-06-24
chore: upgrade dependencies

SanketD92:
Issue #, if available:
N/A

Description of changes:
Dependabot suggested upgrade.

Checklist:

AS review ticket id:

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -23,7 +23,8 @@
     "browserslist": "^4.16.5",
     "ws": "^7.4.6",
     "normalize-url": "^4.5.1",
-    "glob-parent": "^5.1.2"
+    "glob-parent": "^5.1.2",
+    "color-string": "^1.5.5"
   },
   "browserslist": {
     "production": [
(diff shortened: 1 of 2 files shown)


carvantes (commented):
The dependabot PR was slightly better https://github.com/awslabs/service-workbench-on-aws/pull/546.

Resolutions should only be used as a last resort. In this case it's not strictly needed since `1.5.5` is compatible with the existing requirement of `color-string@^1.5.4`

Was there a reason to not simply accept the dependabot PR?

btw, you can upgrade a transitive dependency by deleting it from yarn.lock and reinstalling it https://medium.com/@ayushya/upgrading-javascript-packages-deep-dependencies-using-yarn-8b5983d5fb6b

`yarn upgrade` doesn't work for transitives: https://github.com/yarnpkg/yarn/issues/4986

SanketD92:
> The dependabot PR was slightly better #546.
> 
> Resolutions should only be used as a last resort. In this case it's not strictly needed since `1.5.5` is compatible with the existing requirement of `color-string@^1.5.4`
> 
> Was there a reason to not simply accept the dependabot PR?
> 
> btw, you can upgrade a transitive dependency by deleting it from yarn.lock and reinstalling it https://medium.com/@ayushya/upgrading-javascript-packages-deep-dependencies-using-yarn-8b5983d5fb6b
> 
> `yarn upgrade` doesn't work for transitives: [yarnpkg/yarn#4986](https://github.com/yarnpkg/yarn/issues/4986)

The dependabot PR was made against mainline. Instead of changing base and pushing an empty commit (for semantic check to work), I just created a new PR altogether. I can reopen that PR and close this out.

PR #550, opened 2021-06-24
test: add provisioning env w/ studies and adv study perm int tests

ray7yu:
Issue #, if available: https://sim.amazon.com/issues/GALI-996

Description of changes:
Add support for env type configurations and external Data source study in the integration test `example.yml` file
Add test case for the provisioning of a workspace with mounted studies.
Add advanced tests that verify the propagation of study permission updates on provisioned EC2Linux instances with studies:

These tests make updates to the study's permissions, switching a user between the Admin, Readwrite, Readonly, and None permission levels. A test connects to the instance through an SSH connection and then runs shell commands in order to confirm if the permissions have changed. The tester will have to supply their own EC2Linux `envTypeId` and `envConfigId` from their own deployment on the `{stage}.yml` file, and also supply a provisioned Data Source `externalStudy` name if they wish to run the BYOB test.

Checklist:

AS review ticket id:


main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
```diff
@@ -166,4 +179,90 @@ describe('Create workspace-service-catalog scenarios', () => {
...
+        }),
+      );
+
+      const workflows = await admin1Session.resources.workflows
+        .versions('wf-provision-environment-sc')
+        .version(1)
+        .instances()
+        .get();
+      let result;
+      for (const w of workflows) {
```
ahl27: Do we have a helper method to do this already? Should be able to find the workflow without iterating over all of them. If not, I think lodash has a _.get() method (or maybe its _.match()? I'm not an expert on lodash but I know there's something) that can do this as well in a slightly cleaner way.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
```diff
@@ -166,4 +179,90 @@ describe('Create workspace-service-catalog scenarios', () => {
...
+        if (
+          w.input.envId === env.id &&
+          w.input.envTypeId === env.envTypeId &&
+          w.input.envTypeConfigId === env.envTypeConfigId
+        ) {
+          result = w;
+        }
+      }
+      while (result.wfStatus !== 'done' && result.wfStatus !== 'error') {
```
ahl27: Check out other [tests with workflows](https://github.com/awslabs/service-workbench-on-aws/tree/mainline/main/integration-tests/__test__/api-tests/workflows), specifically `get-workflow-instances.test.js`. Most of these kinds of things we already have helper methods for, in this case I'm pretty sure `.triggerAndWait()` does the same thing as this while loop but slightly cleaner.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
```diff
@@ -166,4 +179,90 @@ describe('Create workspace-service-catalog scenarios', () => {
       });
     });
   });
+  describe('Creating a workspace service catalog environment with studies', () => {
+    it('for EC2Linux', async () => {
```
ahl27: usually for the `it()` part it should kinda read like a sentence that clearly describes what the test itself does--`describe()` says broadly what that test suite is for. So like for these, you could have like:

`describe('Workspace SC Env with studies, () => {
    it('should correctly create a workspace with studies on EC2Linux', ...`

something like that, or whatever your test is designed to do

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
```diff
@@ -166,4 +179,90 @@ describe('Create workspace-service-catalog scenarios', () => {
...
+      const env = await admin1Session.resources.workspaceServiceCatalogs.create({
+        name: workspaceName,
+        envTypeId: setup.defaults.envTypes.ec2Linux.envTypeId,
+        envTypeConfigId: setup.defaults.envTypes.ec2Linux.envTypeConfigId,
+        studyIds,
+        description: 'assignment',
+        projectId: setup.defaults.project.id,
+        cidr: '54.240.196.186/32',
+      });
+      expect(env).toEqual(
```
ahl27: you can use `expect(env).toMatchObject({...args})` if you want here, it checks that `{...args}` is a subset of the object passed in (`env` in this case).

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
```diff
@@ -166,4 +179,90 @@ describe('Create workspace-service-catalog scenarios', () => {
...
+        await new Promise(r => setTimeout(r, 1000 * 30));
+        const getPromises = await Promise.all([
+          admin1Session.resources.workflows
+            .versions('wf-provision-environment-sc')
+            .version(1)
+            .instances()
+            .instance(result.id)
+            .get(),
+        ]);
+        result = getPromises[0];
```
ahl27: Are you missing an expect statement here? These lines don't seem to be doing anything for the test.

ahl27 (commented):
good job! just some minor comments to streamline. This test also seems fairly large, although it might be unavoidable if you're trying to test bigger things. If possible though it's nice to have each test just be testing a specific component of the flow, I don't know if that's at all relevant to this test 😅

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ray7yu: I added the _.filter method from lodash to simplify this piece of code.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ray7yu: `triggerAndWait()` can't be used for my test, because the workflow is triggered when the workspace is provisioned. I looked into implementing the waiting aspect for that function, but I still obtain the no-await-in-loop error. Do you happen to know why `triggerAndWait()` doesn't run into this issue itself?

ahl27: I'm not super familiar with the method, but if I had to guess it's probably because `triggerAndWait()` doesn't pause execution, it just returns a `Promise()` for the value that's unresolved until the function completes (without requiring the loop to wait).

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ray7yu: I made this change in the code for the studies and workspace.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ray7yu: I initially had the array access operator right after the async call. This led to issues, so I just assigned it in a different statement, line 264.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ray7yu: I see, I made a change accordingly to the test blocks.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
hatimkhan: > `triggerAndWait()` can't be used for my test, because the workflow is triggered when the workspace is provisioned. I looked into implementing the waiting aspect for that function, but I still obtain the no-await-in-loop error. Do you happen to know why `triggerAndWait()` doesn't run into this issue itself?

Are you getting linting warnings?  if so, you can suppress them, we do where ever it is appropriate.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
hatimkhan: Any reason why you want to do Promise.all?  I think you can write a simple while loop with await delay and await admin1Session.resource.workflows...get()

hatimkhan: I am also thinking that you can move this to a helper method, maybe we already have a helper method that does that?

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ray7yu: Yes, I started suppressing them. A new issue is that the code will seem to run multiple times if I use lodash or a while-loop based polling technique. For example, one test run should provision one workspace and complete one workflow accordingly, but it will provision several workspaces and workflows until it halts. I believe this issue is from the async/await being in a loop. Is there a better technique for polling the workflow's status?

hatimkhan: Some javascript constructs support the usage of async (depending on node.js version)  for example 'for  of'  and 'while' support async/await.  lodash _.forEach() does not support it and array.forEach() does not support async.  It is possible that you might have used a combination that does not support async/await.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ahl27: also in response to the first comment, i saw the assignment but there's nothing that checks the actual value of the result, so it's not actually being tested.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ray7yu: It turns out that the multiple workspaces bug was the Jest retry logic, it doesn't print out errors until the last retry fails. I also changed the retry count to 1 because workspace provisioning has a large time overhead.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ray7yu: I created utility functions for polling and moved it to `workflow-version.js`, and modified triggerAndWait to reduce redundancy.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
```diff
@@ -166,4 +167,69 @@ describe('Create workspace-service-catalog scenarios', () => {
...
+      const workspaceName = setup.gen.string({ prefix: 'workspace-sc-test' });
+
+      const env = await admin1Session.resources.workspaceServiceCatalogs.create({
+        name: workspaceName,
+        envTypeId: setup.defaults.envTypes.ec2Linux.envTypeId,
+        envTypeConfigId: setup.defaults.envTypes.ec2Linux.envTypeConfigId,
+        studyIds,
+        description: 'assignment',
+        projectId: setup.defaults.project.id,
+        cidr: '54.240.196.186/32',
```
SanketD92, line 212: This CIDR block will not be applicable when running the test on a build server (like GitHub actions)

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
```diff
@@ -166,4 +167,69 @@ describe('Create workspace-service-catalog scenarios', () => {
...
+
+      // Poll until workspace is provisioned
+      await admin1Session.resources.workflows
+        .versions('wf-provision-environment-sc')
+        .version(1)
+        .findAndPollWorkflow(env.id, 10000, 48);
+
+      // Terminate instance
+      await admin1Session.resources.workspaceServiceCatalogs.workspaceServiceCatalog(env.id).delete();
+      await admin1Session.resources.workflows
```
SanketD92: Is there no delay observed between an environment delete API response and the workflow instance actually getting created? Do you think we need a delay added here before we start polling?

main/integration-tests/support/resources/workflows/workflow-version.js
```diff
@@ -56,23 +55,35 @@ class WorkflowVersion extends Resource {
...
-    let counter = 0;
-    let status;
     const triggerInfo = await this.trigger(body);
     const instanceId = _.get(triggerInfo, 'instance.id');
+    await this.waitUntilComplete(instanceId, 1000, maxSecondsCount);
+    return triggerInfo;
+  }
 
+  // Polls a workflow at an intermittent interval until it has completed, failed, or it has reached the maxIntervalCount.
+  async waitUntilComplete(wfInstanceId, interval = 1000, maxIntervalCount = 300) {
```
SanketD92, line 65: Do you think 5 minutes is enough for the workflow instance to complete? I've seen workspace provisioning take up to 8 minutes.

ray7yu: That's the default params, in line 225 of create-workspace-service-catalog.js I set the interval as 10 seconds and the interval count to 48 -> 8 minutes total.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
ray7yu: Ok, is it similar to the deflake delay in collection-resource.js and resource.js? I'll make it 2 seconds as well.

SanketD92: Yes, that should be sufficient for now.

main/integration-tests/__test__/api-tests/workspace-service-catalogs/create-workspace-service-catalog.test.js
```diff
@@ -166,4 +168,71 @@ describe('Create workspace-service-catalog scenarios', () => {
...
+
+      // Poll until workspace is provisioned
+      sleep(2000);
+      await admin1Session.resources.workflows
+        .versions('wf-provision-environment-sc')
+        .version(1)
+        .findAndPollWorkflow(env.id, 10000, 48);
+
+      // Terminate instance
+      await admin1Session.resources.workspaceServiceCatalogs.workspaceServiceCatalog(env.id).delete();
```
SanketD92: 1. Could you move the terminate logic to workspaceServiceCatalog resource's cleanup? 
2. What happens if the terminate workflow fails? Does the test fail?

ray7yu: 1. Okay, I'll move it there.
2. Not at the moment, should I add logic to make it fail in that case?

ray7yu: The workspaceServiceCatalog cleanup itself contains similar logic to what I implemented in my test, it not running must've been a jest rerunning problem.

SanketD92 (dismissed):
LGTM

SanketD92 (dismissed):

main/integration-tests/config/settings/example.yml
```diff
@@ -58,9 +58,5 @@ localApiEndpoint: http://localhost:4000
 # Provide the id of a configuration for an imported EMR environment
 # emrLinuxConfigId: ""
 
-# Provide the id of the AWS account for Data Source studies
-# accountId: ""
-# Provide the name of the external bucket for Data Source studies
-# bucketName: ""
 # Provide the name of the external Data Source study
 # externalStudy: ""
```
SanketD92: Since "external" means something else in SWB, let's call this study "byobStudy"

SanketD92 (dismissed):

SanketD92 (approved):

PR #551, opened 2021-06-24
feat: GALI-885 Backend SC4 - Egress store termination

MBtea:
Issue #, if available:

Description of changes:

1. Add egress store termination functions
2. Add egress store termination tests
3. Updated openapi.yml

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/data-egress/__test__/data-egress-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/data-egress/__test__/data-egress-service.test.js
@@ -32,7 +32,12 @@ const LockService = require('@aws-ee/base-services/lib/lock/lock-service');
 const AWSMock = require('aws-sdk-mock');
 const EnvironmentScService = require('../../environment/service-catalog/environment-sc-service');
 const DataEgressService = require('../data-egress-service');
+const { updateS3BucketPolicy } = require('../../helpers/utils');
 
+jest.mock('../../helpers/utils', () => ({
+  ...jest.requireActual('../../helpers/utils'),
+  updateS3BucketPolicy: jest.fn(),
+}));
 describe('DataEgressService', () => {
   let container;
   let dataEgressService;
@@ -46,16 +51,16 @@ describe('DataEgressService', () => {
     return {
       Statement: [
         {
-          Sid: 'Get:test-id',
+          Sid: 'Get:test-id/',
           Effect: 'Allow',
-          Principal: { AWS: ['arn:aws:iam::test-accountId:root'] },
+          Principal: { AWS: 'arn:aws:iam::test-accountId:root' },
           Action: ['s3:GetObject'],
-          Resource: ['arn:aws:s3:::test-egressStoreBucketNametest-id*'],
+          Resource: ['arn:aws:s3:::test-egressStoreBucketName/test-id/*'],
         },
         {
-          Sid: 'Put:test-id',
+          Sid: 'Put:test-id/',
           Effect: 'Allow',
-          Principal: { AWS: ['arn:aws:iam::test-accountId:root'] },
+          Principal: { AWS: 'arn:aws:iam::test-accountId:root' },
           Action: [
             's3:AbortMultipartUpload',
             's3:ListMultipartUploadParts',
@@ -63,15 +68,15 @@ describe('DataEgressService', () => {
             's3:PutObjectAcl',
             's3:DeleteObject',
           ],
-          Resource: ['arn:aws:s3:::test-egressStoreBucketNametest-id*'],
+          Resource: ['arn:aws:s3:::test-egressStoreBucketName/test-id/*'],
         },
         {
-          Sid: 'List:test-id',
+          Sid: 'List:test-id/',
           Effect: 'Allow',
-          Principal: { AWS: ['arn:aws:iam::test-accountId:root'] },
+          Principal: { AWS: 'arn:aws:iam::test-accountId:root' },
           Action: ['s3:ListBucket'],
           Resource: 'arn:aws:s3:::test-egressStoreBucketName',
-          Condition: { StringLike: { 's3:prefix': ['test-id*'] } },
+          Condition: { StringLike: { 's3:prefix': ['test-id/*'] } },
         },
       ],
     };
@@ -100,6 +105,7 @@ describe('DataEgressService', () => {
     s3Service = await container.find('s3Service');
     lockService = await container.find('lockService');
     dbService = await dataEgressService.service('dbService');
+    dataEgressService.audit = jest.fn();
   });
 
   afterEach(() => {
@@ -300,7 +306,6 @@ describe('DataEgressService', () => {
       AWSMock.mock('S3', 'putBucketPolicy', putBucketPolicyMock);
 
       const result = await dataEgressService.createEgressStore(requestContext, rawEnvironment);
-      expect(putBucketPolicyMock).toHaveBeenCalledTimes(1);
       expect(result).toStrictEqual(mockEgressStoreInfo);
     });
   });
@@ -315,7 +320,7 @@ describe('DataEgressService', () => {
         },
       };
       const requestContext = {};
-      await expect(dataEgressService.terminateEgressStore(requestContext, 'test-id')).rejects.toThrow(
+      await expect(dataEgressService.terminateEgressStore(requestContext, 'test-id/')).rejects.toThrow(
         // It is better to check using boom.code instead of just the actual string, unless
         // there are a few errors with the exact same boom code but different messages.
         // Note: if you encounter a case where a service is throwing exceptions with the
@@ -396,6 +401,7 @@ describe('DataEgressService', () => {
     });
 
     it('should success delete egress store while egress store is not in PROCESSING status', async () => {
+      dataEgressService.removeEgressStoreBucketPolicy = jest.fn();
       const s3Policy = testS3PolicyFn();
       dataEgressService._settings = {
         get: settingName => {
@@ -444,7 +450,85 @@ describe('DataEgressService', () => {
       AWSMock.mock('S3', 'putBucketPolicy', putBucketPolicyMock);
 
       await dataEgressService.terminateEgressStore(requestContext, envId);
-      expect(putBucketPolicyMock).toHaveBeenCalledTimes(1);
+      expect(dataEgressService.removeEgressStoreBucketPolicy).toHaveBeenCalledTimes(1);
+    });
+
+    it('should remove bucket policy', async () => {
+      dataEgressService.getS3BucketAndPolicy = jest.fn().mockResolvedValueOnce({
+        s3BucketName: 'test-egressStoreBucketName',
+        s3Policy: testS3PolicyFn(),
+      });
+      const mockEgressStoreInfo = {
+        id: `egress-store-test-id`,
+        readable: true,
+        writeable: true,
+        kmsArn: 'test-arn',
+        bucket: 'test-egressStoreBucketName',
+        prefix: 'test-id/',
+        envPermission: {
+          read: true,
+          write: true,
+        },
+        status: 'reachable',
+        createdBy: 'test-createdBy',
+        workspaceId: 'test-id',
+        projectId: 'test-projectId',
+        resources: [
+          {
+            arn: ` arn:aws:s3:::test-egressStoreBucketName/test-id/*`,
+          },
+        ],
+      };
+      const mockRevisedStatements = [
+        {
+          Action: ['s3:GetObject'],
+          Effect: 'Allow',
+          Principal: {
+            AWS: [],
+          },
+          Resource: ['arn:aws:s3:::test-egressStoreBucketName/test-id/*'],
+          Sid: 'Get:test-id/',
+        },
+        {
+          Action: [
+            's3:AbortMultipartUpload',
+            's3:ListMultipartUploadParts',
+            's3:PutObject',
+            's3:PutObjectAcl',
+            's3:DeleteObject',
+          ],
+          Effect: 'Allow',
+          Principal: {
+            AWS: [],
+          },
+          Resource: ['arn:aws:s3:::test-egressStoreBucketName/test-id/*'],
+          Sid: 'Put:test-id/',
+        },
+        {
+          Action: ['s3:ListBucket'],
+          Condition: {
+            StringLike: {
+              's3:prefix': ['test-id/*'],
+            },
+          },
+          Effect: 'Allow',
+          Principal: {
+            AWS: [],
+          },
+          Resource: 'arn:aws:s3:::test-egressStoreBucketName',
+          Sid: 'List:test-id/',
+        },
+      ];
+
+      await dataEgressService.removeEgressStoreBucketPolicy({}, mockEgressStoreInfo, 'test-accountId');
+
+      expect(updateS3BucketPolicy).toHaveBeenCalledWith(
+        expect.anything(),
+        'test-egressStoreBucketName',
+        expect.anything(),
+        mockRevisedStatements,
+      );
+      expect(dataEgressService.audit).toHaveBeenCalledTimes(1);
     });
   });
 });
--- a/addons/addon-base-raas/packages/base-raas-services/lib/data-egress/__test__/data-egress-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/data-egress/__test__/data-egress-service.test.js
@@ -25,7 +25,7 @@ jest.mock('@aws-ee/base-services/lib/settings/env-settings-service');
 const SettingsServiceMock = require('@aws-ee/base-services/lib/settings/env-settings-service');
 const AwsService = require('@aws-ee/base-services/lib/aws/aws-service');
 const JsonSchemaValidationService = require('@aws-ee/base-services/lib/json-schema-validation-service');
-const DBService = require('@aws-ee/base-services/lib/db-service');
+const DbServiceMock = require('@aws-ee/base-services/lib/db-service');
 const AuditWriterService = require('@aws-ee/base-services/lib/audit/audit-writer-service');
 const S3Service = require('@aws-ee/base-services/lib/s3-service');
 const LockService = require('@aws-ee/base-services/lib/lock/lock-service');
@@ -39,6 +39,8 @@ describe('DataEgressService', () => {
   let aws;
   let environmentScService;
   let s3Service;
+  let lockService;
+  let dbService;
 
   const testS3PolicyFn = () => {
     return {
@@ -80,7 +82,7 @@ describe('DataEgressService', () => {
     container = new ServicesContainer();
     container.register('aws', new AwsService());
     container.register('jsonSchemaValidationService', new JsonSchemaValidationService());
-    container.register('dbService', new DBService());
+    container.register('dbService', new DbServiceMock());
     container.register('auditWriterService', new AuditWriterService());
     container.register('settings', new SettingsServiceMock());
     container.register('s3Service', new S3Service());
@@ -96,6 +98,8 @@ describe('DataEgressService', () => {
     environmentScService = await dataEgressService.service('environmentScService');
     environmentScService.getMemberAccount = jest.fn().mockResolvedValue({ accountId: 'test-accountId' });
     s3Service = await container.find('s3Service');
+    lockService = await container.find('lockService');
+    dbService = await dataEgressService.service('dbService');
   });
 
   afterEach(() => {
@@ -173,7 +177,7 @@ describe('DataEgressService', () => {
         },
       };
 
-      s3Service.createFolder = jest.fn(() => {
+      s3Service.createPath = jest.fn(() => {
         throw new Error();
       });
       const requestContext = {};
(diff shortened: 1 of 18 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -124,15 +137,104 @@ class DataEgressService extends Service {
...
+    await lockService.tryWriteLockAndRun({ id: lockId }, async () => {
+      await this.addEgressStoreBucketPolicy(requestContext, egressStore, memberAccountId);
+    });
 
     return egressStore;
   }
 
+  async terminateEgressStore(requestContext, environmentId) {
+    const enableEgressStore = this.settings.get(settingKeys.enableEgressStore);
+    const by = _.get(requestContext, 'principalIdentifier.uid');
```
jn1119: Can this API be called by anyone? I am assuming there should be a check to see either the caller is an SWB admin or owns the environment?

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -124,15 +137,104 @@ class DataEgressService extends Service {
...
+          true,
+        );
+      }
+      egressStoreInfo.status = TERMINATED_STATUS_CODE;
+      egressStoreInfo.updatedBy = by;
+      egressStoreInfo.updatedAt = new Date().toISOString();
+      await runAndCatch(
+        async () => {
+          return this._updater()
+            .condition('attribute_exists(id)') // yes we need this to ensure the environment does not exist already
```
jn1119: can you clarify this comment. Did you mean egressStore instead of environment?

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -124,15 +137,104 @@ class DataEgressService extends Service {
...
+    const lockService = await this.service('lockService');
+    const lockId = `bucket-policy-access-${bucketName}`;
+    await lockService.tryWriteLockAndRun({ id: lockId }, async () => {
+      await this.addEgressStoreBucketPolicy(requestContext, egressStore, memberAccountId);
+    });
 
     return egressStore;
   }
 
+  async terminateEgressStore(requestContext, environmentId) {
```
jn1119, line 155: We should probably audit this method. Can you add that if you didn't already?

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/__test__/data-egress-service.test.js
```diff
@@ -285,11 +295,156 @@ describe('DataEgressService', () => {
...
+        // It is better to check using boom.code instead of just the actual string, unless
+        // there are a few errors with the exact same boom code but different messages.
+        // Note: if you encounter a case where a service is throwing exceptions with the
+        // same code but different messages (to address different scenarios), you might
+        // want to suggest to the service author to use different codes.
+        expect.objectContaining({ boom: true, code: 'forbidden', safe: true }),
+      );
+    });
+
+    it('should success delete egress store while egress store is not in PROCESSING status', async () => {
```
jn1119, line 403: Can we check as an assertion that we are removing egress policy for the specific environment in request? Perhaps add test for `removeEgressStoreBucketPolicy` separately and mock `removeEgressStoreBucketPolicy` in this test and verify we called with right params?

addons/addon-base/packages/services/lib/s3-service.js
```diff
@@ -177,13 +177,37 @@ class S3Service extends Service {
...
-  async createFolder(bucketName, folderName) {
+  async createPath(bucketName, folderName) {
     const params = {
       Bucket: bucketName,
       Key: folderName,
     };
     return this.api.putObject(params).promise();
   }
+
+  async clearPath(bucketName, dir) {
```
jn1119, line 188: Add unit test for this method?

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -124,15 +137,104 @@ class DataEgressService extends Service {
...
+    }
+    const egressStoreInfo = egressStoreScanResult[0];
+    const egressStoreStatus = egressStoreInfo.status;
+
+    if (egressStoreStatus.toUpperCase() === PROCESSING_STATUS_CODE) {
+      throw this.boom.forbidden(
+        `Egress store: ${egressStoreInfo.id} is still in processing, workspce can not be terminated before egress request is processed.`,
+        true,
+      );
+    } else if (egressStoreStatus.toUpperCase() === PROCESSED_STATUS_CODE) {
```
jn1119: What is the behavior of the API when the state is neither PROCESSED nor PROCESSING? Do we simply return without any errors. Is that expected?

MBtea: Yes, by design, if the status is in 'CREATED' or 'PENDING', we should do nothing about the egress store and move on terminating the workspace.

addons/addon-base/packages/services/lib/s3-service.js
MBtea: I don't think that would be necessary since the clearPath function is basically calling AWS APIs.

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/__test__/data-egress-service.test.js
MBtea: Fix applied in commit: 3917dc0ed4b0fa6bb7aafbd67d45cd7846a58922

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: Fix applied in commit: 3917dc0ed4b0fa6bb7aafbd67d45cd7846a58922

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: Fix applied in commit: 3917dc0ed4b0fa6bb7aafbd67d45cd7846a58922

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: Fix applied in commit: 3917dc0ed4b0fa6bb7aafbd67d45cd7846a58922

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
jn1119: This is a separate API call for terminating the egress store, right? Shouldn't the API throw an exception for CREATED/PENDING, since technically the call was for terminating the egress store and not for terminating the workspace?

addons/addon-base/packages/services/lib/s3-service.js
```diff
@@ -177,13 +177,37 @@ class S3Service extends Service {
...
+    const deleteParams = {
+      Bucket: bucketName,
+      Delete: { Objects: [] },
+    };
+
+    listedObjects.Contents.forEach(({ Key }) => {
+      deleteParams.Delete.Objects.push({ Key });
+    });
+
+    await this.api.deleteObjects(deleteParams).promise();
```
jn1119: Do we need to verify the response of deleteObjects API and verify that there were no errors? Also, if the bucket is versioned (which I thought was the case by default), what is the behavior we want? Do we want to mark the objects for deletion or actually delete them?

addons/addon-base/packages/services/lib/s3-service.js
jn1119: In general, I don't think it really matters even if there is a single AWS API call- we should write a unit test regardless since the tests are added for regression as well(not for just the fact that things work right now). Having said that, there is a lot going on in that method ranging from list, delete and recursive clearPath call. Overall we need the unit tests to check if we are calling S3 with correct parameters, the base case works, and multiple iterations works (i.e. the recursion ends).

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -71,22 +81,26 @@ class DataEgressService extends Service {
...
+      s3Service.createPath(bucketName, folderName);
     } catch (error) {
       throw this.boom.badRequest(`Error in creating egress store:${folderName} in bucket: ${bucketName}`, true);
     }
 
     // prepare info for ddb and update egress store info
     const egressStoreId = uuid();
+    const creationTime = new Date().toISOString;
     const dbObject = {
       id: egressStoreId,
```
jn1119, line 92: This just caught my attention. Seems like this API is not idempotent (i.e. for multiple calls for the same environment we will keep creating new entries in DDB), and this can happen in case of retries due to bad connection or any other reason. We should fix this. Can we just use something like `egress-{environmentId}` as the identifier to make it easer.

Note that while retrieving the entry, if we are changing the right one. Best to just ensure we have one entry instead of doing a scan and changing one entry.

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -124,15 +138,114 @@ class DataEgressService extends Service {
...
+      egressStoreScanResult = await this._scanner()
+        .limit(1000)
+        .scan()
+        .then(egressStores => {
+          return egressStores.filter(store => store.workspaceId === workspaceId);
+        });
+    } catch (error) {
+      throw this.boom.notFound(`Error in terminating egress store: ${JSON.stringify(error)}`, true);
+    }
+    const egressStoreInfo = egressStoreScanResult[0];
```
jn1119, line 182: probably an error if you see more than one result

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -71,22 +81,26 @@ class DataEgressService extends Service {
...
       id: egressStoreId,
       egressStoreName: `${environment.name}-egress-store`,
+      createdAt: creationTime,
       createdBy: environment.createdBy,
       workspaceId: environment.id,
       projectId: environment.projectId,
       s3BucketName: bucketName,
       s3BucketPath: folderName,
-      status: 'Created',
+      status: CREATED_STATUS_CODE,
```
jn1119, line 100: somewhere we should disallow creating of the egress store if the entry already exists for the environment. Note that current `id` is a new UUID which won't work I think?

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -124,15 +138,114 @@ class DataEgressService extends Service {
...
+            arn: `arn:aws:s3:::${egressStoreInfo.s3BucketName}/${environmentId}/`,
+          },
+        ],
+      };
+
+      // Remove egress store related s3 policy from the s3 bucket
+      const memberAccountId = await this.getMemberAccountId(requestContext, environmentId);
+
+      const lockService = await this.service('lockService');
+      const lockId = `bucket-policy-access-${egressStoreInfo.s3BucketName}`;
```
jn1119, line 255: Given that create/terminate are changing the same DDB row- should we run all create/terminate methods under an environmentId specific lock to avoid potential race conditions.

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -59,6 +68,7 @@ class DataEgressService extends Service {
 
   async createEgressStore(requestContext, environment) {
```
jn1119, line 68: btw, I am assuming that this won't be an API that is exposed by SWB.

edit: it would remain a method call for environment provisioning

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: Yes, the API should be called for terminate the egress store, however, the message here indicates the workspace termination can not executed because of the egress store status is processing. updated message in commit: 160edb79c93f9d709150bd27c0c46acf198b0662

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: Yes, you are right, this would only be served as an internal method called during environment provisioning.

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: Overall, I think the create/terminate methods won't run into race condition since they happens in fixed sequential. But I did applied the lock in the fix just in case the race condition when updating processing and processed status in other API calls. 
fix in commit: c33f4053406ffd30ff50181448e5cba7a7925a90

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: Updated in commit: 1076076507c38a7e39046b29d7fdacae3f50ed21

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: Fixed in commit: c33f4053406ffd30ff50181448e5cba7a7925a90

addons/addon-base/packages/services/lib/s3-service.js
MBtea: Since the bucket is enabled with versioning. Only delete marker will be put on objects.
Test added in commit: 8eccd37803ee539e96dbc9839eab3bae57be8d5a

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: fixed in commit: 1076076507c38a7e39046b29d7fdacae3f50ed21

addons/addon-base/packages/services/lib/s3-service.js
MBtea: Tests added in commit: 8eccd37803ee539e96dbc9839eab3bae57be8d5a

addons/addon-base/packages/services/lib/__test__/s3-service.test.js
```diff
@@ -0,0 +1,89 @@
...
+    });
+    it('should clear s3 object in certain s3 bucket path', async () => {
+      AWSMock.mock('S3', 'listObjectsV2', (params, callback) => {
+        expect(params).toMatchObject({
+          Bucket: 'test-bucketName',
+          Prefix: 'test-dir',
+        });
+        callback(null, {
+          Prefix: [{ key: 'test-key1' }, { key: 'test-key2' }],
+        });
```
jn1119, line 77: Add another test for paginated response of s3 list where we require multiple iterations to clear bucket

addons/addon-base/packages/services/lib/s3-service.js
```diff
@@ -191,7 +191,12 @@ class S3Service extends Service {
       Prefix: dir,
     };
 
-    const listedObjects = await this.api.listObjectsV2(listParams).promise();
+    let listedObjects;
+    try {
+      listedObjects = await this.api.listObjectsV2(listParams).promise();
+    } catch (error) {
+      throw this.boom.badRequest(`S3Service error with listing objects in arn:aws:s3:::${bucketName}/${dir}`, true);
+    }
```
jn1119, line 199: Add test for this error?

addons/addon-base/packages/services/lib/s3-service.js
```diff
@@ -203,10 +208,12 @@ class S3Service extends Service {
...
-    await this.api.deleteObjects(deleteParams).promise();
-
-    if (listedObjects.IsTruncated) await this.clearPath(bucketName, dir);
+    try {
+      await this.api.deleteObjects(deleteParams).promise();
+      if (listedObjects.IsTruncated) await this.clearPath(bucketName, dir);
+    } catch (error) {
+      throw this.boom.badRequest(`S3Service error with deleting objects in arn:aws:s3:::${bucketName}/${dir}`, true);
+    }
   }
```
jn1119: Add test for this error?

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -124,15 +142,124 @@ class DataEgressService extends Service {
...
+    }
+
+    if (egressStoreScanResult.length !== 1) {
+      throw this.boom.internalError(
+        `Error in terminating egress store: multiple result fetched from egrss store table`,
+        true,
+      );
+    }
+    const egressStoreInfo = egressStoreScanResult[0];
+    const isEgressStoreOwner = egressStoreInfo.createdBy === curUser;
```
jn1119, line 183: is it expected that we are allowing environment owner who can be a researcher to terminate the egress store? Note that my earlier comment was around adding permissions but I wasn't sure about the exact use case before.

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -71,37 +80,46 @@ class DataEgressService extends Service {
...
-    );
-    const kmsArn = await this.getKmsKeyIdArn();
 
+    const lockService = await this.service('lockService');
+    const egressStoreDdbLockId = `egress-store-ddb-access-${egressStoreId}`;
+    await lockService.tryWriteLockAndRun({ id: egressStoreDdbLockId }, async () => {
+      await runAndCatch(
+        async () => {
+          return this._updater()
+            .condition('attribute_not_exists(id)') // yes we need this to ensure the environment does not exist already
```
jn1119, line 111: can you change this comment as well

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/__test__/data-egress-service.test.js
```diff
@@ -285,11 +301,234 @@ describe('DataEgressService', () => {
...
+          Bucket: 'test-egressStoreBucketName',
+        });
+        callback(null, {});
+      });
+      // Mock locking so that the putBucketPolicy actually gets called
+      lockService.tryWriteLockAndRun = jest.fn((_params, callback) => callback());
+      AWSMock.mock('S3', 'putBucketPolicy', putBucketPolicyMock);
+
+      await dataEgressService.terminateEgressStore(requestContext, envId);
+      expect(dataEgressService.removeEgressStoreBucketPolicy).toHaveBeenCalledTimes(1);
```
jn1119, line 453: verify the arguments to removeEgressStoreBucketPolicy

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -124,15 +142,124 @@ class DataEgressService extends Service {
...
+    } catch (error) {
+      throw this.boom.notFound(`Error in terminating egress store: ${JSON.stringify(error)}`, true);
+    }
+
+    if (egressStoreScanResult.length !== 1) {
+      throw this.boom.internalError(
+        `Error in terminating egress store: multiple result fetched from egrss store table`,
+        true,
+      );
+    }
```
jn1119, line 181: Add test for this?

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -124,15 +142,124 @@ class DataEgressService extends Service {
...
+      throw this.boom.forbidden(
+        `You are not authorized to terminate the egress store. Please contact your administrator.`,
+        true,
+      );
+    }
+
+    const s3Service = await this.service('s3Service');
+    const egressStoreStatus = egressStoreInfo.status;
+
+    if (egressStoreStatus.toUpperCase() === PROCESSING_STATUS_CODE) {
```
jn1119, line 194: If the store is in CREATED state, I think we should throw an exception stating that store doesn't have the right state to be terminated since it hasn't processed yet

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/__test__/data-egress-service.test.js
```diff
@@ -285,11 +301,234 @@ describe('DataEgressService', () => {
...
+      lockService.tryWriteLockAndRun = jest.fn((_params, callback) => callback());
       AWSMock.mock('S3', 'putBucketPolicy', putBucketPolicyMock);
 
       const result = await dataEgressService.createEgressStore(requestContext, rawEnvironment);
-      expect(putBucketPolicyMock).toHaveBeenCalledTimes(1);
       expect(result).toStrictEqual(mockEgressStoreInfo);
     });
   });
+  describe('delete Egress Store', () => {
+    it('should fail deleting egress store without enable egress store feature', async () => {
```
jn1119, line 313: Also add test for CREATED and TERMINATED state

jn1119 (approved):
I think it is ok to push the changes and address PR feedback if any in a separate PR

PR #555, opened 2021-06-28
feat: Air gap EC2 and Sagemaker workspaces

nguyen102:
Issue #, if available:

Description of changes:

Removed external dependencies and put them into the `environment-files` folder in S3.

### These are the dependencies placed into the S3 folder.

Dependencies for Sagemaker and Ec2-Linux
* jq-1.5
* goofys

Dependencies for Sagemaker
* fuse-2.9.4
* autostop.py
* 
Dependencies for EC2-Linux
* ec2-instance-connect
* fuse-2.9.2

#### EC2-Linux and EC2-Windows requires access to Cloudformation Private Endpoint to signal that they have successfully finished their `goofys` setup script. This allows Cloudformation to mark their respective stack as `Successfully Created`

To allow these respective workspace access to Cloudformation Endpoint, we need these workspaces to set their Security group outbound rules to point to the CF Endpoint's security group. We also need the CF Endpoint to set their Security group inbound rules to be the workspace SG.

To accomplish this we've set up this network topology.

EC2 Linux/ EC2 Windows Workspace has `DefaultSG` and`CustomSG` attached to it
* Outbound Rule of CustomSG: `CFN SG`

CFN Endpoint has `CFN SG` attached to it
* Inbound Rule of CFN SG: `DefaultSG`

By attaching the `DefaultSG` to the workspaces we can have CFN endpoint point to the workspaces when the CFN Endpoint is initially created.

#### To allow AppStream to connect to ec2 instances
EC2 Linux/ EC2 Windows Workspace has `DefaultSG` and`CustomSG` attached to it
* Inbound Rule of CustomSG: `AppStreamSG`

AppStream has `AppStreamSG` attached to it
* Outbound Rule: `DefaultSG`

By attaching the `DefaultSG` to the workspaces we can have AppStream allow access to the workspaces when AppStream resources are initially created during `onboard-account`.

We export `CFN Security Group`, `Default SG`, and `AppStreamSG` in the Cloudformation stack to make it easy for the workspaces to consume these variables.

 ### Testing
 Manually created new Cloudformation stacks with ec2-window, ec2-linux, and sagemaker Cloudformation templates. These stacks were placed in the Private AppStream Workspace. I then logged into each of these workspaces and tested that they didn't have internet access but were able to access S3 studies.

Checklist:

AS review ticket id:


addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-linux-instance.cfn.yml
```diff
@@ -133,6 +136,9 @@ Resources:
           FromPort: -1
           ToPort: -1
           CidrIp: !Ref AccessFromCIDRBlock
+        - DestinationSecurityGroupId:
+            Fn::ImportValue: !Sub "${SolutionNamespace}-CfnEndpointSecurityGroup"
+          IpProtocol: '-1'
```
jn1119: Don't we need isAppStream flag? Does this work without that flag?

scripts/check-dependency-vulnerabilities.sh
```diff
@@ -7,7 +7,7 @@ pnpm audit --json --audit-level high --prod > audit-results.json
 high_vul=$(cat audit-results.json | jq '.metadata.vulnerabilities.high')
 critical_vul=$(cat audit-results.json | jq '.metadata.vulnerabilities.critical')
 
-high_vul_threshold=2
+high_vul_threshold=4
```
jn1119: why was this changed?

nguyen102: Ah yes, so it looks like `prismjs` requires an update to version `2.4.0` but we're using older version of `prismjs` in `base-raas-ui`. That is why that check is failing. This affects our repo as a whole and is not directly related to this PR. 

This [PR](https://github.com/awslabs/service-workbench-on-aws/pull/556) fixes it for our `docs` subdirectory, but not our `base-raas-ui` directory.

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-windows-instance.cfn.yml
```diff
@@ -155,6 +158,9 @@ Resources:
           FromPort: -1
           ToPort: -1
           CidrIp: !Ref AccessFromCIDRBlock
+        - DestinationSecurityGroupId:
```
nguyen102: Note: Add conditional on AppStream flag

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/provision-environment/provision-environment.js
```diff
@@ -141,6 +141,15 @@ class ProvisionEnvironment extends StepBase {
...
       addParam('KeyName', keyName);
     }
+
+    if (type === 'ec2-windows' || type === 'ec2-linux') {
+      addParam('SolutionNamespace', this.settings.get('solutionNamespace'));
+    }
+
+    if (type === 'sagemaker') {
+      addParam('IsAppStreamEnabled', this.settings.get('isAppStreamEnabled').toString());
+    }
```
jn1119: is there a possibility that these parameters will be added to an old template and not to the new template that customers haven't imported yet. Can we fail environment provisioning in case the template is old and not suited for AppStream but the solution has it enabled?

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-linux-instance.cfn.yml
nguyen102: Good catch. Adding the flag.

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/provision-environment/provision-environment.js
nguyen102: Turns out this `provision-environment` file was used for built in workspaces and not SC workspaces. I made the updates in the correct file, which is `env-config-var-service`. I tested this, and even if we pass these parameters to an old template, the workspace still gets started correctly. 

In terms of checking whether an environment is launching the right type of workspace, I think that check should be done earlier in the process than when the workflowLoopRunner is started. When a user switch from normal workspaces to AppStream enabled workspace, we should have a post deployment script that disables all of the "normal" workspace types. And then from then on, we should check that a workspace type is AppStream compatible before letting the user approves the workspace for launch.

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
```diff
@@ -734,4 +757,16 @@ Outputs:
     Condition: isAppStream
     Value: !Ref AppStreamStack
 
+  CFNEndpointSG:
```
jn1119, line 767: Where is the ingress rule for this defined? Do we not need it for EC2 instances to connect to CFN?

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-linux-instance.cfn.yml
```diff
@@ -173,6 +187,10 @@ Resources:
           DeviceIndex: '0'
           GroupSet:
             - !Ref SecurityGroup
+            - !If
+              - AppStreamEnabled
+              - Fn::ImportValue: !Sub "${SolutionNamespace}-SwbVPCDefaultSG"
```
jn1119, line 198: Does this give each EC2 instance (Windows or Linux based) access to other Windows/Linux systems as well? The instances will still have ssh key/password protection etc but I guess those instances will be reachable? Asking this because default's security groups default ingress might allow itself.

nguyen102: Yep, that's right. Since default security group has ingress to itself other EC2 instances have connectivity to each other. However, since those instances still require ssh key, those instances are still secured.

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
nguyen102: Oops, updating the PR to include that now. It'll be on line 601 to 603.

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-windows-instance.cfn.yml
```diff
@@ -155,11 +163,23 @@ Resources:
...
         - IpProtocol: tcp
           FromPort: 3389
           ToPort: 3389
           CidrIp: !Ref AccessFromCIDRBlock
+        - !If
+          - AppStreamEnabled
+          - SourceSecurityGroupId:
+              Fn::ImportValue: !Sub "${SolutionNamespace}-SwbAppStreamSG"
+            IpProtocol: '-1'
+          - !Ref "AWS::NoValue"
```
jn1119, line 182: Can you add similar egress and ingress rules to RStudio as well. Similarly default security group needs to be added in NetworkInterfaces there.

nguyen102: updated

jn1119 (approved):

PR #560, opened 2021-06-30
feat: session expiry

SanketD92:
Issue #, if available:
N/A

Description of changes:
This is the short-term fix for the ID token expiration bug reported by our customers. Currently the access tokens being generated (by Cognito or otherwise) have an expiration time of 60min, while the default inactivity-warning message has a threshold of 30min. This bug is reported to be independent of inactivity.

This PR adds a ForceLogout component keeping track of user's JWT token expiration and auto logout.

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-ui/packages/base-ui/src/AppContainer.js
+++ b/addons/addon-base-ui/packages/base-ui/src/AppContainer.js
@@ -58,8 +58,20 @@ class AppContainer extends Component {
       return false; // This will stop lodash from continuing the forEach loop
     });
 
+    plugins = _.reverse(pluginRegistry.getPluginsWithMethod('app-component', 'getForceLogoutComponent') || []);
+    let ForceLogout = () => <></>;
+    // We ask each plugin in reverse order if they have the ForceLogout component
+    _.forEach(plugins, plugin => {
+      const result = plugin.getForceLogoutComponent({ location, appContext: getEnv(app) });
+      if (_.isUndefined(result)) return;
+      ForceLogout = result;
+      // eslint-disable-next-line consistent-return
+      return false; // This will stop lodash from continuing the forEach loop
+    });
+
     return (
       <>
+        <ForceLogout />
         <AutoLogout />
         <App />
       </>
--- a/addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
+++ b/addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
@@ -46,9 +46,8 @@ class ForceLogout extends React.Component {
 
   componentDidMount() {
     this.timer = setInterval(() => {
-      const expirationInMins = this.hasTokenExpired();
       // eslint-disable-next-line react/no-unused-state
-      if (expirationInMins > 0) {
+      if (!this.hasTokenExpired()) {
         runInAction(() => {
           this.tokenActive = true;
         });
--- a/addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
+++ b/addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
@@ -46,16 +46,9 @@ class ForceLogout extends React.Component {
 
   componentDidMount() {
     this.timer = setInterval(() => {
-      // eslint-disable-next-line react/no-unused-state
-      if (!this.hasTokenExpired()) {
-        runInAction(() => {
-          this.tokenActive = true;
-        });
-      } else {
-        runInAction(() => {
-          this.tokenActive = false;
-        });
-      }
+      runInAction(() => {
+        this.tokenActive = !this.hasTokenExpired();
+      });
     }, 1000);
   }
 
--- a/addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
+++ b/addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
@@ -1,0 +1,128 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License").
+ *  You may not use this file except in compliance with the License.
+ *  A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+import _ from 'lodash';
+import React from 'react';
+import { observable, action, decorate, computed, runInAction } from 'mobx';
+import { observer, inject } from 'mobx-react';
+import { Button, Modal } from 'semantic-ui-react';
+
+const jwtDecode = require('jwt-decode');
+
+// expected props
+// - authentication
+// - app
+class ForceLogout extends React.Component {
+  constructor(props) {
+    super(props);
+    runInAction(() => {
+      this.tokenActive = true;
+    });
+  }
+
+  get app() {
+    return this.props.app;
+  }
+
+  get authentication() {
+    return this.props.authentication;
+  }
+
+  get modalOpen() {
+    return !this.tokenActive;
+  }
+
+  componentDidMount() {
+    this.timer = setInterval(() => {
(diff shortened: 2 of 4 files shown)


addons/addon-base-ui/packages/base-ui/src/AppContainer.js
```diff
@@ -58,8 +58,20 @@ class AppContainer extends Component {
       return false; // This will stop lodash from continuing the forEach loop
     });
 
+    plugins = _.reverse(pluginRegistry.getPluginsWithMethod('app-component', 'getForceLogoutComponent') || []);
+    let ForceLogout = () => <></>;
+    // We ask each plugin in reverse order if they have the ForceLogout component
+    _.forEach(plugins, plugin => {
```
nguyen102, line 64: This code is duplicated three times. I think there's an opportunity to DRY it up by making that code a private method. 

Then we can just call `getComponent(component)`

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
```diff
@@ -0,0 +1,124 @@
...
+  componentDidMount() {
+    this.timer = setInterval(() => {
+      const expirationInMins = this.tokenExpiresIn();
+      // eslint-disable-next-line react/no-unused-state
+      if (expirationInMins > 0) {
+        this.setState({ tokenActive: true });
+      } else {
+        this.setState({ tokenActive: false });
+      }
+    }, 1000);
```
nguyen102, line 52: I think this can be simplified by changing `tokenExpiresIn` to be `hasTokenExpired`. The method returns a boolean and we can use that to determine the state.
```
const tokenExpired = this.hasTokenExpired();
this.setState({ tokenExpired})
```

nguyen102 (commented):
Overall looks good. Couple of small comments

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
```diff
@@ -0,0 +1,124 @@
...
+  get modalOpen() {
+    return !this.state.tokenActive;
+  }
+
+  componentDidMount() {
+    this.timer = setInterval(() => {
+      const expirationInMins = this.tokenExpiresIn();
+      // eslint-disable-next-line react/no-unused-state
+      if (expirationInMins > 0) {
+        this.setState({ tokenActive: true });
```
hatimkhan: Any reason why you are mixing MobX state management with this.setState()?  
In general, there is no need to use this.setState() otherwise you need to worry about implementing shouldUpdate() correctly.  Using MobX makes this easier.  Instead of this.setState{{ tokenActive }).  Just declare 'tokenActive' as an observable instance variables (just like how you declared 'internvalId'.  Then use runInAction() when you need to update the tokenActive value.

addons/addon-base-ui/packages/base-ui/src/AppContainer.js
nguyen102: Looks like this approach involves a double for loop, which makes the code harder to read. Resolving this comment.

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
nguyen102: We don't actually have to implement `shouldComponentUpdate`, right? It's usually only implemented if we want to optimize performance and don't want to rerender the REACT component on every `state` change. In our case the `state` should only change when the token has expire.

SanketD92: Good point. Made the changes.

addons/addon-base-ui/packages/base-ui/src/AppContainer.js
SanketD92: Per conversation, skipping this for readability.

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
hatimkhan: @nguyen102  here are my thoughts:
With MobX,  it automatically provides an extremely efficient implementation of shouldComponentUpdate(), this way we never have to make a decision of when we should implement the shouldComponentUpdate().  It is better not to mix setState() if we are already using MobX to manage the state of this component.

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
```diff
@@ -0,0 +1,128 @@
...
+    clearInterval(this.timer);
+    return this.authentication.logout({ autoLogout: true });
+  };
+
+  hasTokenExpired = () => {
+    let minutesRemaining;
+    try {
+      const idToken = localStorage.getItem('appIdToken');
+      const decodedIdToken = jwtDecode(idToken);
+      const expiresAt = _.get(decodedIdToken, 'exp', 0) * 1000;
```
nguyen102, line 69: this can just be `return _.get(decodedIdToken, 'exp', Date.now()) >= Date.now())`

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
```diff
@@ -0,0 +1,128 @@
...
+    return this.props.authentication;
+  }
+
+  get modalOpen() {
+    return !this.tokenActive;
+  }
+
+  componentDidMount() {
+    this.timer = setInterval(() => {
+      const expirationInMins = this.hasTokenExpired();
```
nguyen102: The return value is a boolean now, so we can just do
```
runInAction(() => {
          this.tokenActive = this.hasTokenExpired();
        });
```

nguyen102 (changes requested):

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
SanketD92: Wanted to still keep this inside the try-catch block in case the token was not received at all, or if it was corrupted.

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
SanketD92: Ah yes, had to revert my changes for lock file merge conflicts with develop. Pushed the fix.

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
nguyen102: I was thinking something like this. Thoughts?
```
hasTokenExpired = () => {
    try {
      const idToken = localStorage.getItem('appIdToken');
      const decodedIdToken = jwtDecode(idToken);
      return _.get(decodedIdToken, 'exp', Date.now()) >= Date.now())
    } catch (e) {
      return false;
    }
  };
  ```

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
nguyen102: @hatimkhan Yeah, I agree. If we're handling local state in a class, we should use either `mobx` or `REACT state`. I personally prefer to use REACT state if we're just handling local state for a component since REACT is already built to handle that.

In terms of worrying about `shouldComponentUpdate`, from what I've read and experienced, you rarely need to implement that method. Most of the time, the default REACT implementation for `shouldComponentUpdate` is sufficient. You only need to implement that method if your component's `render` method is very large and results in long rendering time if it were to be rerender often. 

So basically the heuristic I follow is not to implement `shouldComponentUpdate` unless I see a performance impact. Seeing a performance impact on the UI is rare.

More info about this can be read [here](http://jamesknelson.com/should-i-use-shouldcomponentupdate/)

hatimkhan: @nguyen102 I like your heuristic, thanks for sharing the link. Since I started using MobX, I never had to think about shouldUpdateComponent at all, with MobX we never need to write one, even for data intensive/expensive components, so you completely eliminate the need for heuristic.

addons/addon-base-ui/packages/base-ui/src/parts/ForceLogout.js
SanketD92: We can do it this way too. I originally wanted to convey to the reader computation is being done by the number of minutes left until token expiration.

SanketD92:
Closing PR. CodeCov acting weird.

PR #566, opened 2021-07-08
feat: Allow creating new AppStream enabled account

nguyen102:
Issue #, if available:

Description of changes:
Allow creating new AppStream enabled account

* UI was updated to asks for new AppStream specific fields if AppStream is enabled
* Backend 
 * Updated to pass AppStream specific fields to `onboard-account.cfn.yml`
 * Updated to share AppStream image from `master` account with `member` account
 * Image created by AppStream ImageBuilder changed from `FirefoxPuttyNotepad_` to be `ServiceWorkbench_v1_`. This change was made to allow a more generic Image name that can be updated to different versions in the future.

**Testing**: Deployed changes to my AWS account and I was able to successfully create a new account through SWB UI. I was able to launch AppStream workspaces in the new account.

**Note**: For internal Amazon customers: When the new account is created but before the AppStream stack is deployed, you need to launch an EC2 instance in the account so your account get a containment score. If you do not do so, you will not get a score, and be unable to launch an AppStream fleet.

For external customers: When the new account is created but before the AppStream stack is deployed, you need to navigate to the AppStream console and click Get Started, then click Skip. This will automatically create the required user role [AmazonAppStreamServiceAccess](https://docs.aws.amazon.com/appstream2/latest/developerguide/roles-required-for-appstream.html#AmazonAppStreamServiceAccess).

Checklist:

AS review ticket id:


addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/provision-account/provision-account.js
```diff
@@ -205,11 +251,22 @@ class ProvisionAccount extends StepBase {
...
           roleArn: cfnOutputs.CrossAccountExecutionRoleArn,
           xAccEnvMgmtRoleArn: cfnOutputs.CrossAccountEnvMgmtRoleArn,
-          subnetId: cfnOutputs.VpcPublicSubnet1,
           vpcId: cfnOutputs.VPC,
           encryptionKeyArn: cfnOutputs.EncryptionKeyArn,
         };
-        await this.addAwsAccountTable(requestContext, awsAccountData);
+        let additionalAccountData = {};
+        if (this.settings.get(settingKeys.isAppStreamEnabled) === 'true') {
+          additionalAccountData = {
```
jn1119, line 311: Should we add other properties as well like fleet or workspace subnet?

main/solution/backend/config/infra/cloudformation.yml
```diff
@@ -761,6 +759,11 @@ Resources:
               - sts:AssumeRole
             Resource:
               - !Sub 'arn:aws:iam::${AWS::AccountId}:role/${self:custom.settings.namespace}-prep-raas-master-MasterRole-*'
+          - Effect: Allow
+            Action:
+              - appstream:UpdateImagePermissions
+            Resource:
+              - !Sub 'arn:aws:appstream:${AWS::Region}:${AWS::AccountId}:image/FirefoxPuttyNotepad*'
```
jn1119: This suffix needs to change I suppose to the new ServiceWorkbenchImage name convention.

nguyen102: Ahh, good catch, I'll make the update

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/provision-account/provision-account.js
nguyen102: @SanketD92 correct me if I'm wrong, but the only additional fields that you need to get the connect API working is `appStreamStackName` and `appStreamSecurityGroup`?

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-aws-accounts.json
```diff
@@ -39,6 +39,14 @@
     "encryptionKeyArn": {
       "type": "string",
       "pattern": "^arn:aws:kms:.*$"
+    },
+    "appStreamStackName": {
+      "type": "string",
+      "minLength": 10
+    },
+    "appStreamSecurityGroup": {
```
SanketD92: How about we name it as: `appStreamSecurityGroupId`

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/provision-account/provision-account.js
SanketD92: That's correct. The stack name is enough to get the associated AppStream resources.

However if there are more than 1 fleets associated to that AppStream stack (edge-case), that would result in an error while connecting. So if we're adding fields anyways I'd vote to add `appStreamFleetName` as well.

Workspace subnet is not needed for connect API changes, so I'll leave that up to you.

nguyen102: Added `appStreanFleetName`. Workspace subnet is needed so SWB knows which subnet to launch workspaces in.

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-aws-accounts.json
nguyen102: updated

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/provision-account/provision-account.js
SanketD92: Yeah, that's correct. And since the existing `subnetId` value won't be available anymore, we should probably replace the value with AppStream's subnetId instead of creating a new field.

jn1119 (dismissed):

jn1119 (commented):
AWS Support commented on our ticket and mentioned that we can create AppStream roles ourselves. It seems like there was a path option in create-role CLI where `service-role` can be specified.

addons/addon-base-raas/packages/base-raas-workflow-steps/lib/steps/provision-account/provision-account.js
```diff
@@ -96,22 +97,102 @@ class ProvisionAccount extends StepBase {
...
+    const [accountService] = await this.mustFindServices(['accountService']);
+    const requestContext = await this.payload.object('requestContext');
+    const accountId = await this.state.string('ACCOUNT_ID');
+    const appStreamImageName = await this.payload.string('appStreamImageName');
+    await accountService.shareAppStreamImageWithMemberAccount(requestContext, accountId, appStreamImageName);
+
+    return this.wait(10).thenCall('createAppStreamRoles');
+  }
+
+  async createAppStreamRoles() {
```
jn1119, line 118: Can you add some unit tests for the newly added code in this file. We should also add tests around isAppStream flag and verify the expected behavior.

jn1119: Let's merge this PR for now to speed up development for other devs. Should be ok since this is a feature branch. Unit tests can follow in a separate PR.

jn1119 (approved):

PR #567, opened 2021-07-08
feat: [GALI-887] - Zhaxan/gali 887 egress object tagging

MBtea:
Description of changes:

- Add egress object tagging feature, when an object put into a egress store s3 bucket, a lambda should be triggered to put tag on the object.
- Add Lambda handler test

Checklist:

AS review ticket id:


addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -65,11 +65,34 @@ class DataEgressService extends Service {
...
+    const workspaceId = environmentId;
+    let egressStoreScanResult = [];
+
+    try {
+      egressStoreScanResult = await this._scanner()
+        .limit(1000)
+        .scan()
+        .then(egressStores => {
+          return egressStores.filter(store => store.workspaceId === workspaceId);
+        });
```
jn1119, line 78: I know we have done this before but I didn't notice before think this is something that won't scale with a lot of workspaces. My suggestion would be to use the hashkey here (I am guessing its workspaceId, right?)

jn1119 (commented):
LGTM overall. Had one unrelated comment that we can resolve outside this PR too. Let me know what you think.

jn1119 (dismissed):

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
MBtea: Yes, make sense, I will push another fix PR for this.

jn1119 (approved):

PR #571, opened 2021-07-13
feat: Story update connect

SanketD92:
Issue #, if available:
GALI-918 & GALI-925 (partial)

Description of changes:
This contains backend additions for using AppStream resources in the hosting account. The PR also contains UI changes for workspace types with the connection scheme set to use https (like SageMaker).

Checklist:

AS review ticket id:


addons/addon-base-raas-appstream/packages/base-raas-appstream-services/lib/appstream/appstream-sc-service.js
```diff
@@ -0,0 +1,207 @@
...
+    const environment = await environmentScService.mustFind(requestContext, { id: environmentId });
+
+    const { stackName, fleetName } = await this.getStackAndFleet(requestContext, {
+      environmentId,
+      indexId: environment.indexId,
+    });
+    if (!stackName || !fleetName) {
+      // return original url as is
+      return finalDestination;
+    }
```
jn1119: Shouldn't this be an exception? Correct me if I am wrong but shouldn't this code path only execute if AppStream is enabled

SanketD92: That's a fair question. Yes, it could be changed to an exception. 

Why I did it this way: I believe we were following the pattern of allowing addons to "contribute" to the final result without needing to check if the feature specific to that addon is enabled. I'm okay to sway from that pattern.

I'll add a guard that returns the finalDestination if the feature is disabled, and throw an exception if it's enabled and AppStream-resources are not found

jn1119 (approved):

PR #574, opened 2021-07-14
feat: [APU] New account onboarding, lambda functions

ahl27:
Issue #, if available: [GALI-952](https://sim.amazon.com/issues/28482fe8-d542-4dc2-8228-dd9e1a184165), [AWSHLT-14](https://sim.amazon.com/issues/e65e0fec-a235-46b9-a033-3f0f52770334)

Description of changes:

This update adds the last set of features to the APU feature, primarily reworking the onboarding process for a new account.

Detailed list of changes:
- New API calls to try to onboard any accounts in PENDING state. The UI has a new button to manually trigger this call.
- bugfixes for ERRORED state
- permissionStatus cases streamlined; NOSTACKNAME case removed
- xAccEnvMgmt dependency removed, new more limited IAM role introduced and added to AwsAccount
- AddAwsAccount and create-aws-account.json forms streamlined, unnecessary inputs removed
- AddAwsAccount now redirects user to onboarding page upon successful creation
- minor reworks to some UI status messages
- minor bugfix for environment filter buttons, apparently 'gold' isn't recognized
- new lambda built to onboard and check permissions of accounts in the background (currently triggers 1/min)
- new IAM role and policy created for lambda (lambda also assumes other new IAM role stored in AwsAccount)
- adds unit tests and refactors some old ones
- updates openapi.yaml with new api call and fixes an issue with another openapi entry being overindented
- SWB now checks stack status for changing account permissionStatus value; stacks that are in progress switch account to PENDING state, stacks that failed switch account to ERRORED, and completed stacks are updated based on template
- When the account is pulled out of PENDING state, the backend DB is updated with stack outputs. This happens after a stack is created or updated (either when the user onboards for the first time or reonboards)

The next PR (after this one) will include changes to naming and wording ~~including the requested change from last PR on switching from SWB-Onboard-* to initial-stack-*~~, and support for SWB-created accounts

Edit: `initial-stack` wording change has been added, and permissionscheck removed from store heartbeat (now exclusively in the lambda function). 

Note: since accounts can be in PENDING state, they may not have finished onboarded yet. A flag or something should probably be added to prevent resource creation for accounts that aren't in `CURRENT` or `NEEDSUPDATE` state. Current error handling should be sufficient for the time being, workspaces will just error out if you try to create them with an account that hasn't finished onboarding.

Checklist:

AS review ticket id:


ahl27:
permissionStatusFlow

This is the current flowchart for permissionStatus values. `[NEW]` is not a status that accounts can have, it just indicates the initial state before the account is even created in SWB. Unknown is a catch-all base case for situations that are totally unexpected.

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
```diff
@@ -316,6 +316,44 @@ Resources:
...
+      RoleName: !Join ['-', [Ref: Namespace, 'cfn-status-role']]
+      Path: '/'
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Effect: 'Allow'
+            Principal:
+              AWS:
+                - !Join [':', ['arn:aws:iam:', Ref: CentralAccountId, 'root']]
+                - !Ref ApiHandlerArn
```
Bingjiling: ApiHandlerArn and WorkflowRoleArn is not needed here actually. Since the root account is granted the permission already, ApiHandlerArn and WorkflowRoleArn can assume this role as long as they got the sts:assumeRole permission for this role.

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
```diff
@@ -316,6 +316,44 @@ Resources:
...
+                - !Ref WorkflowRoleArn
+            Action:
+              - 'sts:AssumeRole'
+            Condition:
+              StringEquals:
+                sts:ExternalId: !Ref ExternalId
+      ManagedPolicyArns:
+        - !Ref PolicyCfnStatus
+      PermissionsBoundary: !Ref PolicyCfnStatus
```
Bingjiling, line 333: Nice!

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-accounts-service.test.js
```diff
@@ -159,12 +159,7 @@ describe('AwsAccountService', () => {
   describe('create', () => {
     const awsAccount = {
       name: 'my-aws-account',
-      externalId: '012345678998',
```
Bingjiling, line 162: Will the externalId be a random generated string since user is not defining it anymore?

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-cfn-service.test.js
```diff
@@ -224,16 +263,17 @@ describe('AwsAccountService', () => {
       expect(res).toEqual('CURRENT');
     });
 
-    it('should correctly set account with undefined cfnStackName to NEEDSONBOARD', async () => {
-      // This account's status should change to 'NEEDSONBOARD'
+    it('should correctly set account with missing cfnStackName to ERRORED', async () => {
```
Bingjiling: Does this mean pre-existing account will be in ERRORED status?

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -236,16 +266,20 @@ class AwsCfnService extends Service {
...
     const checkPermissions = async account => {
-      if (account.cfnStackName === '') {
-        res = 'NEEDSONBOARD';
+      errorMsg = '';
+      if (
+        account.permissionStatus === 'NEEDSONBOARD' ||
+        account.permissionStatus === 'PENDING' // Backend workflow will bring the account out of PENDING
+      ) {
+        res = account.permissionStatus;
         errorMsg = `Account ${account.accountId} needs to be onboarded.`;
```
Bingjiling: Why is this an error situation?

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -236,16 +266,20 @@ class AwsCfnService extends Service {
...
     let res;
     let errorMsg = '';
 
     const checkPermissions = async account => {
-      if (account.cfnStackName === '') {
-        res = 'NEEDSONBOARD';
+      errorMsg = '';
+      if (
+        account.permissionStatus === 'NEEDSONBOARD' ||
+        account.permissionStatus === 'PENDING' // Backend workflow will bring the account out of PENDING
```
Bingjiling: Update comment

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/aws-accounts-controller.js
```diff
@@ -49,6 +49,18 @@ async function configure(context) {
...
   );
 
+  // ===============================================================
+  //  GET /onboard (mounted to /api/aws-accounts)
+  // ===============================================================
+  router.get(
+    '/onboard',
+    wrap(async (req, res) => {
+      const requestContext = res.locals.requestContext;
+      const result = await awsCfnService.attemptOnboardAccounts(requestContext);
```
Bingjiling: Update this method to checkPendingAccounts and route to /api/aws-accounts/pending ?

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -212,19 +217,44 @@ class AwsCfnService extends Service {
       { action: 'check-aws-permissions', conditions: [allowIfActive, allowIfAdmin] },
       { accountId },
     );
+    // see https://github.com/rvedotrc/aws-cloudformation-stack-states for all states
+    const PENDING_STATES = [
```
Bingjiling: These states should be defined at file or class level since they're constants used by multiple methods.

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-aws-accounts.json
```diff
@@ -1,45 +1,22 @@
...
-      "type": "string",
-      "pattern": "^subnet-[a-f0-9]{8,17}$"
-    },
-    "encryptionKeyArn": {
-      "type": "string",
-      "pattern": "^arn:aws:kms:.*$"
-    }
-  },
-  "required": ["name", "roleArn", "externalId", "accountId", "vpcId", "subnetId", "encryptionKeyArn"]
-}
```
Bingjiling, line 45: This might break the `Create Account` functionality. Could you please do a global search on the createAccount method and make sure it's not breaking anything?

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-aws-accounts.json
```diff
@@ -31,11 +31,15 @@
...
     },
+    "cfnStackName": {
+      "type": "string",
+      "maxLength": 2048
+    },
     "roleArn": {
       "type": "string",
       "minLength": 10
     },
-    "xAccEnvMgmtRoleArn": {
```
Bingjiling, line 38: Emmmm we still need the EnvMgmtRole right?

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -305,6 +339,96 @@ class AwsCfnService extends Service {
...
+      const output = _.find(_.get(stack, 'Outputs', []), item => item.OutputKey === prop);
+      return output.OutputValue;
+    };
+
+    fieldsToUpdate.cfnStackId = stack.StackId;
+    fieldsToUpdate.externalId = accountEntity.externalId || 'workbench';
+    fieldsToUpdate.vpcId = findOutputValue('VPC');
+    fieldsToUpdate.subnetId = findOutputValue('VpcPublicSubnet1');
+    fieldsToUpdate.encryptionKeyArn = findOutputValue('EncryptionKeyArn');
+    fieldsToUpdate.roleArn = findOutputValue('CrossAccountEnvMgmtRoleArn');
```
Bingjiling: There should be 3 roles in total in member accounts including: onboard status role, envMgmtRole and crossAccountRole. Are they saved in DDB after onboarding?

main/solution/backend/config/infra/cloudformation.yml
```diff
@@ -744,6 +745,54 @@ Resources:
...
+              - dynamodb:Query
+              - dynamodb:Scan
+              - dynamodb:UpdateItem
+            Resource:
+              - !GetAtt [AwsAccountsDb, Arn]
+              - !Join ['', [!GetAtt [AwsAccountsDb, Arn], '/index/*']]
+          - Effect: Allow
+            Action:
+              - dynamodb:GetItem
+            Resource: !GetAtt [WorkflowsDb, Arn]
```
Bingjiling: Why is workflowsDb needed?

main/solution/backend/config/infra/cloudformation.yml
```diff
@@ -744,6 +745,54 @@ Resources:
...
+          - Effect: Allow
+            Action: dynamodb:UpdateItem
+            Resource: !GetAtt [WorkflowInstancesDb, Arn]
+          - Effect: Allow
```
Bingjiling: Double check the log is printed successfully with these permission? I remember it's a bit tricky restricting CloudWatch permissions.

main/solution/backend/src/lambdas/aws-account-onboarding/handler.js
```diff
@@ -0,0 +1,43 @@
...
+  await container.initServices();
+  await handlerWithContainer(container);
+};
+
+const handlerWithContainer = async container => {
+  const awsCfnService = await container.find('awsCfnService');
+  const userContext = getSystemRequestContext();
+
+  // Check permission status of all accounts
+  await awsCfnService.batchCheckAccountPermissions(userContext);
```
Bingjiling: Emmmm should the batchCheckAccountPermissions method calls attemptOnboardAccounts method? What's the benefit of exposing both methods?

Bingjiling (commented):
Reviewed backend. Will continue on front end part tmr 👍

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-accounts-service.test.js
ahl27: it will not, the externalId is always expected to be 'workbench' by the internals, so it'll always be 'workbench'. The user wasn't expected to define it previously, all of the instructions that reference externalId also ask the user to just enter workbench.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-cfn-service.test.js
ahl27: uhhhhh....yes 😅
I could add a special case for this, but a missing stackname is equivalent to an invalid one, so in theory it should fail to find the stack name, and then return errored status.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: it could go either way--my logic was that the method is trying to check permissions, so accounts that are being handled by the workflow can't be checked, and thus it's like an error. I could probably remove this line and it would be better, since it's sort of a weird response type.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: which one? This comment is still current, I just wanted to indicate that accounts in PENDING state are skipped because they're being handled by another resource

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/aws-accounts-controller.js
ahl27: sure, that makes sense

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: 👍🏻 makes sense

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-aws-accounts.json
ahl27: sure, I'll look into that today

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-aws-accounts.json
ahl27: we don't, it was actually never saved on the backend. If you look at the develop branch, the `AwsAccount` object doesn't even have an entry for xAccEnvMgmtRoleArn, I think I must've added it later just for this. This role ARN is actually the same for every account, so it shouldn't need to be stored anyway

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: only two, onboardStatusRole and the roleArn one. onboardStatusRole is saved in the account's entry in the DDB when it is first created, and then roleArn is stored in this method (it's pulling the CrossAccountEnvMgmtRoleArn value). The third arn shouldn't be necessary to store at the account level.

main/solution/backend/config/infra/cloudformation.yml
ahl27: uhhhh......good catch haha

main/solution/backend/config/infra/cloudformation.yml
ahl27: hmmm I don't think it is honestly, it was giving me some error message. I'll have to look at it again

main/solution/backend/src/lambdas/aws-account-onboarding/handler.js
ahl27: editing away my first comment because I misinterpreted your comment haha

That's a good point, I suppose we could do that. It depends on if we want the permissions check to be callable without trying to onboard accounts. The way I have it currently set up, the lambda runs both functions, and then the permissions checker is on the heartbeat method to run more frequently when the user is actually on the page. I guess we could move it into a single function and just call that here (and possibly in the heartbeat), which would probably simplify things.

the more I think about it the more it seems to make sense to just wrap it all into a single function, so I'll see if I can refactor that.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: removed it 👍🏻

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/aws-accounts-controller.js
ahl27: actually if we're going to roll this into the account permissions check, this change won't be needed

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
ahl27: gotcha, good catch 👍🏻

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/aws-accounts-controller.js
ahl27: just refactored this 👍

main/solution/backend/src/lambdas/aws-account-onboarding/handler.js
ahl27: just refactored this 👍

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-aws-accounts.json
ahl27: it doesn't look like it's breaking any functionality. `Create Account` uses a completely separate service, store, and schema for its stuff, so this schema is only used for the add account feature. The Create Account functionality actually just adds the information directly to the backend db, so it doesn't reference the `aws-accounts-service` at all

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
```diff
@@ -95,13 +96,14 @@ const AwsAccount = types
       self.description = rawAwsAccounts.description || self.description;
       self.accountId = rawAwsAccounts.accountId || rawAwsAccounts.accountId;
       self.externalId = rawAwsAccounts.externalId || self.externalId;
-      self.permissionStatus = rawAwsAccounts.permissionStatus || self.permissionStatus || 'PENDING';
+      self.permissionStatus = rawAwsAccounts.permissionStatus || self.permissionStatus || 'NEEDSONBOARD';
```
nguyen102: Is it a big change to rename these to have underscore separate the word? `NEEDS_ONBOARD` would be easier to read than the current format.

nguyen102: How do we differentiate between `PENDING` and `NEEDS_ONBOARD`. Does `NEEDS_ONBOARD` now replace `PENDING`? Do we still need `PENDING` as a state in the file?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
```diff
@@ -15,19 +15,20 @@
...
 import { createLink } from '@aws-ee/base-ui/dist/helpers/routing';
 
 const statusDisplay = {
-  CURRENT: { color: 'green', display: 'Up-to-Date' },
-  NEEDSUPDATE: { color: 'orange', display: 'Needs Update' },
-  NEEDSONBOARD: { color: 'purple', display: 'Needs Onboarding' },
-  ERROR: { color: 'red', display: 'Error' },
-  UNKNOWN: { color: 'grey', display: 'Unknown' },
+  CURRENT: { color: 'green', display: 'Up-to-Date', spinner: false },
+  NEEDSUPDATE: { color: 'orange', display: 'Needs Update', spinner: false },
```
nguyen102: If we can make this `NEEDS_UPDATE`

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
```diff
@@ -149,13 +155,44 @@ class AccountCard extends React.Component {
...
       <Label attached="top left" size="mini" color={state.color}>
+        {state.spinner && <Icon name="spinner" loading />}
         {state.display}
       </Label>
     );
   }
 
+  renderPendingDetailsAccordion(account) {
+    const expanded = this.detailsExpanded;
+    const onboardMessage = account.permissionStatus === 'NEEDSONBOARD';
```
nguyen102: since this is a boolean, I would name it `shouldShowOnboardMessage`.

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
```diff
@@ -149,13 +155,44 @@ class AccountCard extends React.Component {
...
     return (
       <Label attached="top left" size="mini" color={state.color}>
+        {state.spinner && <Icon name="spinner" loading />}
         {state.display}
       </Label>
     );
   }
 
+  renderPendingDetailsAccordion(account) {
+    const expanded = this.detailsExpanded;
```
nguyen102: `expanded` -> `isExpanded`

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
```diff
@@ -149,13 +155,44 @@ class AccountCard extends React.Component {
...
+    return (
+      <Accordion className="mt2">
+        <Accordion.Title active={expanded} index={0} onClick={this.handleDetailsExpanded}>
+          <Icon name="dropdown" />
+          <b>Details</b>
+        </Accordion.Title>
+        <Accordion.Content active={expanded}>
+          {expanded && (
+            <div className="mb2">
+              {onboardMessage && (
```
nguyen102: For this we can do 
```
{
  shouldShowOnboardMessage ? (This account needs to be onboard..) : (Service workbench is waiting... 
 }
```

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
```diff
@@ -316,6 +311,42 @@ Resources:
...
+    Properties:
+      Description: Allows main account to onboard and check status of aws accounts
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Effect: Allow
+            Action:
+              - cloudformation:DescribeStacks
+              - cloudformation:GetTemplate
+            Resource: !Sub 'arn:aws:cloudformation:${AWS::Region}:${AWS::AccountId}:stack/SWB-*'
```
nguyen102: will need to change to `initial-stack-*`

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -236,17 +268,20 @@ class AwsCfnService extends Service {
 
     const awsAccountsService = await this.service('awsAccountsService');
     const accountsList = await awsAccountsService.list();
-
+    // const stackOutputs = ['externalId', 'vpcId', 'subnetId', 'encryptionKeyArn', 'cfnStackId', 'roleArn'];
```
nguyen102: delete

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-cfn-service.test.js
Bingjiling: We need to avoid showing ERRORED status for pre-existing account, or customers would be filing bugs for it. A missing stackname can be differentiated from an invalid one right?

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
Bingjiling: Ah, I thought you forgot to update this one. Backend workflow in the context of SWB would be interpreted as an actual workflow, since we're not using a workflow here, it's clearer to say Backend Lambda <lambda name> will bring the account out of PENDING.

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-aws-accounts.json
Bingjiling: Gotcha. That's interesting. Thanks for looking into this :)

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-aws-accounts.json
Bingjiling: xAccEnvMgmtRoleArn is present in the [create-aws-accounts schema](https://github.com/awslabs/service-workbench-on-aws/blob/mainline/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-aws-accounts.json#L23), and is needed for most of the Service Catalog related calls in the member account. The ARN is different for every account, since it's a role exist in every member account. 

You might have mixed it up with the EnvMgmtRole in the main account, which is used to manage Service Catalog resources in the main account.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-cfn-service.test.js
ahl27: yep, it can be--should I just default to UNKNOWN if there's no stack name?

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
Bingjiling: xAccEnvMgmtRole is needed, it's used for Service Catalog related calls in member account, the only case it's not needed is if the member account is the same as the main account(which is a rare edge case). You can do a [global search](https://github.com/awslabs/service-workbench-on-aws/search?q=xAccEnvMgmtRoleArn) and see its use case.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: ah okay, i'll make that fix

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-aws-accounts.json
ahl27: hm, but it's not present in the [AwsAccount object](https://github.com/awslabs/service-workbench-on-aws/blob/mainline/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js). Is it only kept in the backend?

ahl27: I can add this back in, no problem

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: nvm, resolved with the other comment

main/solution/backend/config/infra/cloudformation.yml
Bingjiling: You can reference the example here -> https://github.com/awslabs/service-workbench-on-aws/blob/35f6cce3ccc301921ead742240c15c1a7e332f0c/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/sagemaker-notebook-instance.cfn.yml#L86-L96

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
ahl27: shouldn't be a big deal to rename them.

PENDING accounts are handled separately, so the permission checker only looks at accounts not in PENDING state (auto-skipping pending accounts) and then there's a separate functionality that looks at accounts in PENDING and pulls them out of it. Accounts go into PENDING state when their CFN stack is in a pending state. PENDING is definitely required because it indicates to the user that there's changes happening, otherwise there can be issues where the user makes changes but it doesn't look like anything has happened

ahl27: as for this line specifically, it's just the default value if something goes wrong and we don't have a permission status...I think UNKNOWN is probably the right choice to have there rather than PENDING or NEEDS_ONBOARD

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
ahl27: sure, i'll replace it everywhere it occurs

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
ahl27: got it, just made that change

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
ahl27: got it 👍🏻

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
ahl27: oh does that work? nice

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: ah good catch

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-cfn-service.test.js
ahl27: actually I'll probably default it to needs onboard

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
ahl27: Renamed that everywhere it occurs, and I changed this line to UNKNOWN as the default 'missing' value

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
ahl27: yes, I'll make that change today

ahl27: fixed 👍🏻

main/solution/backend/config/infra/cloudformation.yml
ahl27: thanks! looks like I was missing two allowed actions, deploying it now to verify that everything is working

ahl27: everything is working properly now 👍🏻

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-cfn-service.test.js
ahl27: added an additional check to set the status to UNKNOWN if stackname is either missing or empty. UNKNOWN state provides a button to deploy a new stack, and I just tested the flow and it seems to work pretty well.

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-aws-accounts.json
Bingjiling: Yes, it's a miss on the front end side that it's not showed, good catch there! This role is added when the Service Catalog feature is added, and the front end was not updated properly to show this role.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-accounts-service.test.js
Bingjiling: Got it! Yes, that makes sense.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccount.js
nguyen102: yep, UNKNOWN sounds good.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
```diff
@@ -113,6 +113,23 @@ const AwsAccountsStore = BaseStore.named('AwsAccountsStore')
...
+      forceCheckAccountPermissions: async () => {
+        await getAllAccountsPermissionStatus();
+      },
+
+      hasPendingAccounts: () => {
+        const pendingAccountIds = _.map(
+          _.filter(self.awsAccounts, acct => acct.permissionStatus === 'PENDING'),
+          'id',
+        );
+        return _.isEmpty(pendingAccountIds);
```
nguyen102: Should this be negated? Should it be 
`return !_.isEmpty(pendingAccountIds)`

This method would return `true` when the list of `pendingAccounts` is not empty.

nguyen102: Actually could this just be one line?
`return !_.isEmpty(_.filter(self.awsAccounts, acct => acct.permissionStatus === 'PENDING')`

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
```diff
@@ -48,13 +49,17 @@ class AccountCard extends React.Component {
...
+  }
+
   get isSelectable() {
     return this.props.isSelectable;
   }
 
   get permissionStatus() {
-    // Possible Values: CURRENT, NEEDSUPDATE, NEEDSONBOARD, ERRORED, NOSTACKNAME
-    return this.props.permissionStatus;
+    // Possible Values: CURRENT, NEEDS_UPDATE, NEEDS_ONBOARD, ERRORED, NOSTACKNAME
```
nguyen102: `NOSTACKNAME` -> `NO_STACK_NAME`

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AwsAccountsList.js
```diff
@@ -67,6 +67,11 @@ class AwsAccountsList extends React.Component {
     return store.list;
   }
 
+  getPendingAccounts() {
```
nguyen102: getPendingAccounts -> getHasPendingAccounts

Name change since we're getting a boolean and not the accounts themselves.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -152,15 +163,15 @@ class AwsCfnService extends Service {
     const account = await awsAccountsService.mustFind(requestContext, { id: accountId });
     cfnTemplateInfo.template = await cfnTemplateService.getTemplate('onboard-account');
     cfnTemplateInfo.region = this.settings.get(settingKeys.awsRegion);
-    cfnTemplateInfo.name = account.cfnStackName || ['SWB-Onboard-', randomString(20)].join('');
+    cfnTemplateInfo.name = account.cfnStackName || ['initial-stack-', randomString(20)].join('');
```
nguyen102: Let's keep this consistent with the `provision-account` workflow and have this be the Date time instead of randomString.
`initial-stack-${new Date().getTime()}`

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -305,6 +350,76 @@ class AwsCfnService extends Service {
...
+    const accounts = await awsAccountsService.list();
+    const pendingAccountIds = _.map(
+      _.filter(accounts, acct => acct.permissionStatus === 'PENDING'),
+      'id',
+    );
+
+    const auditLog = {};
+    const newStatus = {};
+
+    const processor = async awsAccountId => {
```
nguyen102: `processor` -> `onboardAccount`

Let's give it a more descriptive name than `processor`

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -305,6 +350,76 @@ class AwsCfnService extends Service {
     }
   }
 
+  async checkPendingAccounts(requestContext) {
```
nguyen102: It looks like this method actually onboard pending accounts? Should it be renamed to that? `onboardPendingAccounts`.

`checkPendingAccounts` would lead me to think it's a method that doesn't mutate state, but this method actually does change state.

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
```diff
@@ -212,19 +228,40 @@ class AwsCfnService extends Service {
...
+
+      // still hash values
+      return trimmedExpPermString !== trimmedCurPermString ? 'NEEDS_UPDATE' : 'CURRENT';
+    }
+
+    // we should never make it here unless something really goes wrong
+    return 'UNKNOWN';
   }
 
   async batchCheckAccountPermissions(requestContext, batchSize = 5) {
```
nguyen102: should this be `batchCheckAndUpdateAccountPermissions`?

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/aws-accounts/AwsAccountsStore.js
ahl27: hmmmm indeed it could be haha, good thought

ahl27: actually this made me realize that this isn't needed at all anyway, i just removed it entirely

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AccountCard.js
ahl27: thanks, NOSTACKNAME is deprecated so I'll just remove it

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/accounts/AwsAccountsList.js
ahl27: removed this function since it's not needed

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: will do!

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: 👍🏻

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: uhhhhh.....it probably be should be haha, at first it was just checking but now it's updating--I'll change it everywhere it appears

addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/aws-cfn-service.js
ahl27: that's true, I can change this--it calls the attemptOnboard function to actually do the onboarding, but I agree since it's calling that as a subfunction the parent function is then mutating state

nguyen102 (approved):

PR #578, opened 2021-07-15
feat: Appstream connection updates

SanketD92:
Issue #, if available:
GALI-925

Description of changes:
UI Changes for AppStream

Checklist:

AS review ticket id:


addons/addon-base-raas-appstream/packages/base-raas-appstream-services/lib/appstream/appstream-sc-service.js
```diff
@@ -111,7 +116,10 @@ class AppStreamScService extends Service {
     return `${uid}-${sessionSuffix}`.replace(/[^\w+=,.@-]+/g, '').slice(0, 32);
   }
 
-  async urlForFirefox(requestContext, { environmentId }) {
+  async getStreamingUrl(requestContext, { environmentId, applicationId }) {
+    const isAppStreamEnabled = this.settings.get(settingKeys.isAppStreamEnabled);
+    if (!isAppStreamEnabled) return undefined;
```
jn1119: I guess we can just throw an error here

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvSshConnRowExpanded.js
```diff
@@ -88,25 +99,73 @@ class ScEnvSshConnRowExpanded extends React.Component {
...
+    const interfaces = this.networkInterfaces;
+    const moreThanOne = _.size(interfaces) > 1;
+
+    return (
+      <div>
+        <b>Connection instructions for your AppStream workspace:</b>
+        <List bulleted>
+          <List.Item>
+            The IP Address or DNS of the instance.{' '}
+            {moreThanOne ? 'Ask your administrator if you are not sure which one to use:' : ''}
```
jn1119: We shouldn't have moreThanOne condition true for AppStream for now. We can keep this check around but just wanted to make sure we don't display it currently for any environment when AppStream is enabled..

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvSshConnRowExpanded.js
```diff
@@ -88,25 +99,73 @@ class ScEnvSshConnRowExpanded extends React.Component {
...
+                <List.Item key={network.value} className="flex">
+                  {this.renderHostLabel(network)}
+                  <CopyToClipboard text={network.value} />
+                </List.Item>
+              ))}
+            </List>
+          </List.Item>
+          <List.Item>Click the &apos;Connect&apos; button below to navigate to the AppStream instance</List.Item>
+          <List.Item>You downloaded the private key when you created the SSH key.</List.Item>
+          <List.Item>Save this key&apos;s content into AppStream&apos;s Notepad application in .PEM format</List.Item>
```
jn1119: Do we need to add guidance on the directory they should use. Something like `C:\Users\Photon\Documents`. I think a bunch of dirs are admin only, so might be worth adding this info.

Even though the file would eventually be deleted, I suggest that we add a line for manual deletion once SSH is established

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvSshConnRowExpanded.js
```diff
@@ -88,25 +99,73 @@ class ScEnvSshConnRowExpanded extends React.Component {
...
+          <List.Item>Open PuttyGen in AppStream and convert your private PEM key to PPK format</List.Item>
+          <List.Item>Enter the PPK file and private IP address details in Putty to connect to EC2</List.Item>
```
jn1119: Before this step shouldn't there be some step suggesting that the user needs to click `use this key` or something in SWB UI?

scripts/app-stream/SETUP.md
```diff
@@ -29,7 +29,7 @@ Note: Please set up your [AWS Profile](https://docs.aws.amazon.com/cli/latest/us
 cd ~\Documents
 
 # Pull the Image Builder script from Github
-Invoke-WebRequest -Uri https://raw.githubusercontent.com/awslabs/service-workbench-on-aws/feat-secure-workspace-egress/scripts/app-stream/buildImage.ps1 -OutFile buildImage.ps1
+Invoke-WebRequest -Uri https://raw.githubusercontent.com/awslabs/service-workbench-on-aws/feat-appstream-ui/scripts/app-stream/buildImage.ps1 -OutFile buildImage.ps1
```
jn1119: I think this branch needs to be mainline?

scripts/app-stream/buildImage.ps1
```diff
@@ -3,6 +3,8 @@ Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManage
 
 choco install -y putty.install
 
+Invoke-WebRequest -Uri https://raw.githubusercontent.com/awslabs/service-workbench-on-aws/feat-appstream-ui/scripts/app-stream/ec2linux.ps1 -OutFile 'C:\Users\Public\Documents\EC2Linux.ps1'
```
jn1119: This too, branch should be mainline eventually, so should we change it now?

addons/addon-base-raas-appstream/packages/base-raas-appstream-services/lib/appstream/appstream-sc-service.js
SanketD92: I think these methods should not throw errors. These are called via plugins and they only contribute to the URL generation if AppStream is enabled. If not, the regular (SC) URLs take effect.

The logic in `addons/addon-base-raas-appstream/packages/base-raas-appstream-services/lib/plugins/env-sc-connection-url-plugin.js` ensures that if AppStream URL returned by this service is `undefined`, then it just returns back the original SC connection info.

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvSshConnRowExpanded.js
SanketD92: These instructions appear after the user clicks on the 'Use this key' button. Since the 'Use this key' button is the only  component while selecting the key pair, I don't think guidance to click would be necessary.

scripts/app-stream/SETUP.md
SanketD92: Eventually, yes. I can change it now too.

scripts/app-stream/buildImage.ps1
SanketD92: Eventually, yes. I can change it now too.

addons/addon-base-raas-appstream/packages/base-raas-appstream-services/lib/appstream/appstream-sc-service.js
jn1119: I see, I think this is related to my original question on an earlier PR where I mentioned that we shouldn't even call these methods if AppStream isn't enabled. Is that hard to achieve?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvSshConnRowExpanded.js
jn1119: Hmm, that's a lot of work to do in one minute then. Can't we show these steps when the user clicks on connections? I thought if we decouple `use this key` button with instructions then the user has more time.

scripts/app-stream/SETUP.md
jn1119: Either keep it `feat-secure-workspace-egress` or change to `mainline`. My preference is change to mainline now since we might forget it later.

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvSshConnRowExpanded.js
SanketD92: As discussed, we'll be directing users to store their PEM files in Downloads and deleting once connected.

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvSshConnRowExpanded.js
SanketD92: Moving these instructions, as discussed.

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/parts/ScEnvSshConnRowExpanded.js
SanketD92: True. Doesn't hurt to have it, but I don't see it getting used with AppStream as well.

jn1119 (dismissed):

jn1119 (approved):

PR #594, opened 2021-07-23
feat: workspace provisioning mgmt

SanketD92:
Issue #, if available:
GALI-921

Description of changes:
[TRE] Disable workspace provisioning if AppStream is not setup in AWS account.

UI: Disable provision button
Backend: Throw error if workspace provision API is called
In scope: end to end test

In a gist:
Show workspace provisioning button only if any project account is AppStream-configured.
Finally, env provisioning should only list projects configured with AppStream (If a user tries this for non-AppStream configured projects, API throws error)

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/setup/CreateInternalEnvForm.js
+++ b/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/setup/CreateInternalEnvForm.js
@@ -141,6 +141,7 @@ class CreateInternalEnvForm extends React.Component {
             </Header.Subheader>
           </Header>
         </Segment>
+        {this.renderButtons()}
       </>
     );
   }
--- a/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/setup/CreateInternalEnvForm.js
+++ b/addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/setup/CreateInternalEnvForm.js
@@ -2,7 +2,7 @@ import _ from 'lodash';
 import React from 'react';
 import { decorate, computed, runInAction, observable, action } from 'mobx';
 import { observer, inject } from 'mobx-react';
-import { Segment, Button, Header } from 'semantic-ui-react';
+import { Segment, Button, Header, Icon } from 'semantic-ui-react';
 import { displayError } from '@aws-ee/base-ui/dist/helpers/notification';
 import Dropdown from '@aws-ee/base-ui/dist/parts/helpers/fields/DropDown';
 import Form from '@aws-ee/base-ui/dist/parts/helpers/fields/Form';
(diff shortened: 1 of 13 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-service.js
```diff
@@ -385,7 +402,15 @@ class EnvironmentScService extends Service {
     const { envTypeId, envTypeConfigId, projectId } = environment;
 
     // Lets find the index id, by looking at the project and then get the index id
-    const { indexId } = await projectService.mustFind(requestContext, { id: projectId, fields: ['indexId'] });
+    // The isAppStreamConfigured attribute value will be returned by project service. No other fields needed to be added
+    const { indexId, isAppStreamConfigured } = await projectService.mustFind(requestContext, {
+      id: projectId,
+      fields: ['indexId'],
+    });
```
jn1119, line 412: Outside scope of this but I just thought about this. It seems like we will need to look at other APIs like createConnectionUrl, find etc and throw similar errors

jn1119 (dismissed):

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/setup/CreateInternalEnvForm.js
```diff
@@ -21,20 +21,37 @@ import SelectConfigurationCards from './SelectConfigurationCards';
 // - defaultCidr (via props)
 // - clientInformationStore (via injection)
 // - userStore (via injection)
+// - projectsStore (via injection)
 class CreateInternalEnvForm extends React.Component {
```
nguyen102, line 25: Out of curiosity, which screen is this? Is this for creating internal workspaces? Do we still support internal workspaces?

addons/addon-base-raas/packages/base-raas-services/lib/project/project-service.js
```diff
@@ -209,13 +224,74 @@ class ProjectService extends Service {
...
+      const [awsAccountsService, indexesService] = await this.service(['awsAccountsService', 'indexesService']);
+      const accounts = await awsAccountsService.list(systemContext, {
+        fields: ['id', 'appStreamFleetName', 'appStreamSecurityGroupId', 'appStreamStackName'],
+      });
+      const indexes = await indexesService.list(systemContext, { fieldsToGet: ['id', 'awsAccountId'] });
+      const accountIdsWithAppStream = _.map(
+        _.filter(
+          accounts,
+          account =>
+            !_.isUndefined(account.appStreamFleetName) &&
```
nguyen102, line 271: Thoughts on making the operation an `||` instead of `&&`? I believe if any of those fields are missing, AppStream will not function correctly when trying to generate an AppStream link.

addons/addon-base-raas/packages/base-raas-services/lib/project/project-service.js
```diff
@@ -209,13 +224,74 @@ class ProjectService extends Service {
...
+  /**
+   * Returns the same object type as the input with an added isAppStreamConfigured boolean flag for each project entity
+   *
+   * IMPORTANT: The system context is used instead of the user's request context.
+   * This is because non-admin users do not have access to resources like indexes and awsAccounts,
+   * which are required for config verification. No other information from these resources are made available to non-admins
+   *
+   * @param input The Project entity, or list of Project entities
+   */
+  async verifyAppStreamConfig(input) {
```
nguyen102: The term `verify` usually means no change is made to the `input`. Since we're modifying the `input`, what are your thoughts on renaming this method to `updateWithAppStreamConfig`?

main/end-to-end-tests/cypress/support/commands.js
```diff
@@ -51,6 +51,12 @@ Cypress.Commands.add('login', role => {
       email: Cypress.env('researcherEmail'),
       password: Cypress.env('researcherPassword'),
     };
+  }
+  if (role === 'restrictedResearcher') {
+    loginInfo = {
+      email: Cypress.env('restrictedResearcherEmail'),
```
nguyen102, line 57: Remember to add this to GH Actions. Also add the creds [here](https://quip-amazon.com/4gSRAngBjLcO/Trigger-E2E-Test-GitHub-Action-Locally#QTb9CArMy7D)

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/setup/CreateInternalEnvForm.js
SanketD92: No, the one for internal workspaces is here: `addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/compute/parts/CreateInternalPlatformForm.js`.

This component is the final form in SC env provisioning, which comes up after selecting the workspace type.

addons/addon-base-raas/packages/base-raas-services/lib/project/project-service.js
SanketD92: You're right. Making the change

main/end-to-end-tests/cypress/support/commands.js
SanketD92: Yes, it's on my list. I wanted to wait on doing that till we merge the feature into develop, similar to the other e2e tests for appstream.

addons/addon-base-raas/packages/base-raas-services/lib/project/project-service.js
SanketD92: We can't, my unit test caught it :)

If any of those fields are missing we want to exclude that account from the array. Therefore we need to use `&&` instead of `||`, because here we're only accepting accounts with _None_ of these 3 fields set to `undefined`.

nguyen102 (approved):

Issue #590, opened 2021-07-21
[Bug] Looking at Study details in data-sources fails


avillach:
**Describe the bug**
From Admin page https://service-workbench.hms.harvard.edu/data-sources
when I try to open ANY study I get a new total blank page

**To Reproduce**
https://service-workbench.hms.harvard.edu/data-sources
click on ANY study

**Versions (please complete the following information):**
v3.3.0

**Additional context**
This is the first time that I configure studies as BYOB in SWB PROD.   worked fine before on SWB DEV
So I don't know if this is related to a newer version

avillach:
My bad, I didn't meant to close it. Still open issue

jn1119:
Thanks Paul for reaching out, I have started looking into this issue and will post an update shortly.

jn1119:
I used 3.3.0 release and created couple of new data sources in my test environment. Unfortunately, I wasn't able to reproduce the issue. We can try the following as next steps:

1. Do the studies show up under studies/organization tab and are you able to use them via a workspace?
2. Can you try another browser and see if you see the same issue?
3. Do you see any errors in network tab when you go to the data-sources tab. Note that the video you uploaded expands the study but the relevant data-sources API calls had already been called. In order to see the errors coming from data-sources tab just login to service workbench like you normally do, and before clicking on data-sources tab open the network tab. Any related cloudwatch logs would also be useful.

jn1119:
Also, can you make sure that both your data sources stacks are up to date (perhaps you can re-run them?). I wonder if something has changed in the account which has those buckets.

jn1119:
Hey Paul, we will be releasing v3.3.1 today. We were able to reproduce the bug if study permissions contain a user that was deleted by admin. If that's what is happening in your case, v3.3.1 should fix this issue.

jn1119:
Version [3.3.1](https://github.com/awslabs/service-workbench-on-aws/releases/tag/v3.3.1) is now released. Can you try it out and let us know if your issue is fixed. Also instead of using the `network` tab in developer tools, can you select the `console` instead and paste the error you end up seeing if UI crashes. Thanks.

avillach:
issue resolved with 3.3.1. Thanks!

PR #595, opened 2021-07-24
fix: Remove delete user feature from UI and handle study permissions which have stale users

jn1119:
Issue #, if available: [Issue-590](https://github.com/awslabs/service-workbench-on-aws/issues/590)

Description of changes:

* In [Issue-590](https://github.com/awslabs/service-workbench-on-aws/issues/590) it was observed that expanding study permissions on a study which has already deleted users causes SWB website to crash.
* Since deleting a user is can cause unknown side effects on resources that the user has permission to, I have removed the delete user button from UI. Note that the API still exists and the documentation has been updated to reflect the warning.
* I have updated permissions loading to account for deleted users so now UI wouldn't crash but we would still log the stale user on console as an error.
* Our update permission workflow has been changed to account for deleted users as well. I have made it such that our UI would now send stale user ids to remove list of update study permissions API.

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-ui/packages/base-ui/src/models/users/UsersStore.js
+++ b/addons/addon-base-ui/packages/base-ui/src/models/users/UsersStore.js
@@ -142,7 +142,6 @@ const UsersStore = BaseStore.named('UsersStore')
             if (userSnapshot) {
               result.push(User.create(userSnapshot));
             }
-            // result.push(User.create(getSnapshot(userIdentifier)));
           }
         }
       });
--- a/addons/addon-base-ui/packages/base-ui/src/models/users/UsersStore.js
+++ b/addons/addon-base-ui/packages/base-ui/src/models/users/UsersStore.js
@@ -131,7 +131,18 @@ const UsersStore = BaseStore.named('UsersStore')
           if (user) {
             result.push(user);
           } else {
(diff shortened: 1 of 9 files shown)


addons/addon-base-ui/packages/base-ui/src/models/users/UsersStore.js
```diff
@@ -131,7 +131,18 @@ const UsersStore = BaseStore.named('UsersStore')
...
+              userSnapshot = getSnapshot(userIdentifier);
+            } catch (error) {
+              // Note that user might be already deleted. In order to prevent UI from crashing log the error instead
+              // and not add it to User list
+              console.log(`User ${userIdentifier.uid} doesn't exist`, error);
+            }
+            if (userSnapshot) {
+              result.push(User.create(userSnapshot));
+            }
+            // result.push(User.create(getSnapshot(userIdentifier)));
```
nguyen102: should this line be deleted?

nguyen102 (dismissed):

addons/addon-base-ui/packages/base-ui/src/models/users/UsersStore.js
jn1119: Yeah, removed :)

nguyen102 (approved):

PR #608, opened 2021-07-29
feat: Restrict SageMaker presigned URL access to private VPC endpoint

jn1119:
Issue #, if available: GALI-1042

Description of changes:
1. Added a new policy for each SageMaker instance which if used would ensure that presigned URLs are only accessible from SageMaker interface endpoint in AppStream
2. For the actual code change, note that we are updating the policy and giving IP address of the caller (Lambda function) access to create presigned URL. Given that caller IP address would also have access to the URL, we are updating the role and removing that access once URL is generated. The presigned URLs that are generated use the latest role policy that they were created from. This way we ensure that URL is accessible only from AppStream instances which are using the specific VPC endpoint we created for SageMaker.

Testing:
1. Added unit tests
2. Also tested the change from UI and ensured that only AppStream instances have access to presigned URLs

- Page shown when SageMaker presigned URL is opened outside AppStream
Screen Shot 2021-07-29 at 7 42 04 AM

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas-appstream/packages/base-raas-appstream-services/lib/plugins/env-sc-connection-url-plugin.js
+++ b/addons/addon-base-raas-appstream/packages/base-raas-appstream-services/lib/plugins/env-sc-connection-url-plugin.js
@@ -26,6 +26,7 @@ async function createConnectionUrl({ envId, connection }, { requestContext, cont
   const isSsh = connection.scheme === 'ssh';
   const isRdp = connection.scheme === 'rdp';
   const appStreamScService = await container.find('appStreamScService');
+  const environmentScConnectionService = await container.find('environmentScConnectionService');
   const settings = await container.find('settings');
   const isAppStreamEnabled = settings.optionalBoolean(settingKeys.isAppStreamEnabled, false);
 
@@ -34,6 +35,10 @@ async function createConnectionUrl({ envId, connection }, { requestContext, cont
   // it will only be triggered during the URL creation API call
   if (!isAppStreamEnabled || connection.operation === 'list') {
     return { envId, connection };
+  }
+
+  if (_.toLower(_.get(connection, 'type', '')) === 'sagemaker') {
+    connection.url = await environmentScConnectionService.createPrivateSageMakerUrl(requestContext, envId, connection);
   }
 
   // Only wrap via AppStream if the connection.url exists
--- a/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
@@ -19,7 +19,7 @@ const crypto = require('crypto');
 const NodeRSA = require('node-rsa');
 const querystring = require('querystring');
 const Service = require('@aws-ee/base-services-container/lib/service');
-const { retry, sleep, linearInterval } = require('@aws-ee/base-services/lib/helpers/utils');
+const { retry, linearInterval } = require('@aws-ee/base-services/lib/helpers/utils');
 const sshConnectionInfoSchema = require('../../schema/ssh-connection-info-sc');
 const { connectionScheme } = require('./environment-sc-connection-enum');
 const { cfnOutputsToConnections } = require('./helpers/connections-util');
@@ -386,8 +386,6 @@ class EnvironmentScConnectionService extends Service {
       );
       const currentPolicyResponse = await this.getCurrentRolePolicy(iam, connection);
       await this.updateRoleToIncludeCurrentIP(iam, connection, currentPolicyResponse);
-      // wait for permission to propagate before using the role
-      await sleep(2000);
       const createPresignedURLFn = async () => {
         const stsEnvMgmt = await environmentScService.getClientSdkWithEnvMgmtRole(
           requestContext,
--- a/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
@@ -14,13 +14,20 @@
  */
 
 const _ = require('lodash');
+let fetch = require('node-fetch');
 const crypto = require('crypto');
 const NodeRSA = require('node-rsa');
 const querystring = require('querystring');
 const Service = require('@aws-ee/base-services-container/lib/service');
+const { sleep, retry } = require('@aws-ee/base-services/lib/helpers/utils');
 const sshConnectionInfoSchema = require('../../schema/ssh-connection-info-sc');
 const { connectionScheme } = require('./environment-sc-connection-enum');
 const { cfnOutputsToConnections } = require('./helpers/connections-util');
+
+// Webpack messes with the fetch function import and it breaks in lambda.
+if (typeof fetch !== 'function' && fetch.default && typeof fetch.default === 'function') {
+  fetch = fetch.default;
+}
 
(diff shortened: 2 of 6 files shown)


addons/addon-base-raas-appstream/packages/base-raas-appstream-services/lib/plugins/env-sc-connection-url-plugin.js
```diff
@@ -26,6 +26,7 @@ async function createConnectionUrl({ envId, connection }, { requestContext, cont
   const isSsh = connection.scheme === 'ssh';
   const isRdp = connection.scheme === 'rdp';
   const appStreamScService = await container.find('appStreamScService');
+  const environmentScConnectionService = await container.find('environmentScConnectionService');
   const settings = await container.find('settings');
   const isAppStreamEnabled = settings.optionalBoolean(settingKeys.isAppStreamEnabled, false);
```
nguyen102, line 31: Sorry not related to this PR, but I believe AppStream is pulled in as a string here. We are assuming it's a string which is a bug.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
```diff
@@ -360,6 +369,108 @@ class EnvironmentScConnectionService extends Service {
     return { password, networkInterfaces: this.toNetworkInterfaces(instanceInfo) };
   }
 
+  async createPrivateSageMakerUrl(requestContext, envId, connection, presign_retries = 10) {
+    const [lockService] = await this.service(['lockService']);
```
nguyen102: This can be 
`const lockService = await this.service('lockService');`

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
```diff
@@ -360,6 +369,108 @@ class EnvironmentScConnectionService extends Service {
...
+      );
+      const currentPolicyResponse = await iam
+        .getRolePolicy({
+          RoleName: connection.role,
+          PolicyName: connection.policy,
+        })
+        .promise();
+
+      // Construct new statement which will allow the caller IP address permission to generate the presigned URL
+      const currentIpAddress = await fetch('http://checkip.amazonaws.com/').then(function(res) {
```
nguyen102: Should we log a warning incase we fail to get the IP address?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
```diff
@@ -360,6 +369,108 @@ class EnvironmentScConnectionService extends Service {
     return { password, networkInterfaces: this.toNetworkInterfaces(instanceInfo) };
   }
 
+  async createPrivateSageMakerUrl(requestContext, envId, connection, presign_retries = 10) {
```
nguyen102, line 372: This method is getting quite large. Should we consider breaking up the code to contain two helper methods.
```
createPrivateSagemakerUrl(...) {
    updateSagemakerPolicy(newPolicy);
    generatePrivateSagemakerUrl();
    updateSagemakerPolicy(originalPolicy);
}
```

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
```diff
@@ -360,6 +369,108 @@ class EnvironmentScConnectionService extends Service {
...
+      }
+      const putRolePolicyParams = {
+        RoleName: connection.role,
+        PolicyName: connection.policy,
+        PolicyDocument: JSON.stringify(policyToUpdate),
+      };
+      await iam.putRolePolicy(putRolePolicyParams).promise();
+
+      // wait for permission to propagate before using the role
+      await sleep(1000);
```
nguyen102: We might want to do a while loop to deterministically wait for the connection to be available. Maybe max wait time of 10 seconds while we keep trying to get the url. If the request fails with a permission error, wait one second and try again. 

We should probably also log if we fail to generate the url after 10 seconds.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
```diff
@@ -360,6 +369,108 @@ class EnvironmentScConnectionService extends Service {
...
+    });
+    return signedURL;
+  }
+
+  async getSageMaker({ accessKeyId, secretAccessKey, sessionToken }) {
+    const aws = await this.getAWS();
+    return new aws.sdk.SageMaker({ accessKeyId, secretAccessKey, sessionToken });
+  }
+
+  async getAWS() {
```
nguyen102, line 481: We only use this method once, and it's only 2 lines. We might not need to create a method for this.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
jn1119: This isn't a warning. It's actually an error which will make the call fail and the error would be reflected in CloudFormation. Should we be doing more?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
jn1119: Yeah, I agree

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
jn1119: Actually this sleep is completely unnecessary. The call to generate the URL is happening with retries (exponential backoff). So I will just remove it.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
jn1119: This is done for making it easier to mock in unit tests.

addons/addon-base-raas-appstream/packages/base-raas-appstream-services/lib/plugins/env-sc-connection-url-plugin.js
SanketD92: Maybe you mean `settingKeys.isAppStreamEnabled` value is a string and not a boolean. If so, the optionalBoolean method is meant to return the boolean value for the passed key string.

@nguyen102 Are you seeing a bug when connecting to AppStream enabled instances?

nguyen102: Ah nvm, I tested the `settings.optionalBoolean` [code](https://github.com/awslabs/service-workbench-on-aws/blob/d862fccc3912a5b41aefdd934a55d27cfe214a7a/addons/addon-base/packages/services/lib/settings/env-settings-service.js#L163) and it converts a string to a boolean.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
jn1119: Added a comment

nguyen102 (approved):

PR #615, opened 2021-07-30
fix: static namespace bug fix

maghirardelli:
Issue #, if available: GALI-871

Description of changes: added a function to check if the resolved namespace is static and if so, to provide the necessary prefix (analysis-) and suffix (unique datetime string) so Cloudformation behaves as expected. Does not alter how dynamic namespaces are generated.

Checklist:

AS review ticket id:


addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
```diff
@@ -90,7 +91,9 @@ class LaunchProduct extends StepBase {
 
     // Read input params specified in the environment type configuration
     // The params may include variable expressions, resolve the expressions by using the resolveVars
-    const resolvedInputParams = await this.resolveVarExpressions(envTypeConfig.params, resolvedVars);
+    const resolvedInputParamsRaw = await this.resolveVarExpressions(envTypeConfig.params, resolvedVars);
+    // Additional layer to check the namespace is valid and unique
```
jn1119: I am guessing that namespace might have become static after going through `resolveVarExpressions` method. We should probably add a comment to explain why this additional layer of check is required because just before `resolveVarExpressions` we do have a namespace ending with current date.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,79 @@
...
+            // Build
+            const resolvedInputParams = [{Key: 'Namespace', Value: 'staticname-2626262626'}];
+            const datetime = Date.now();
+    
+            // Operate
+            const after = await lp.checkNamespace(resolvedInputParams, datetime);
+    
+            // Check
+            const numberSequence = resolvedInputParams[0].Value.split('-').pop();
+            await expect(after[0].Value.startsWith('analysis-')).toBeTruthy() && expect(Number(after[0].Value.split('-').pop())).not.toBe(NaN) && expect(after[0].Value.split('-').pop() === numberSequence).toBeFalsy();
```
jn1119: Would comparing `resolvedInputParams[namespaceIndex]` and `analysis-staticname-2626262626-${datetime}` to be equal be simpler?

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,79 @@
...
+describe('LaunchProduct', ()=> {
+      const meta = {};
+      const input = {};
+    
+      const lp = new LaunchProduct({
+        step: { config: {} },
+        workflowPayload: new WorkflowPayload({ meta, input, workflowInstance: { steps: [] } }),
+      });
+
+    describe('checkNamespace', () => {
```
jn1119, line 29: Perhaps also add a test which starts with `staticname-${datetime}` and ends in `analysis-staticname-${datetime}` where datetime is `${Date.now}`. Just to cover all the cases.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,79 @@
...
+        it('Dynamic name should not be altered', async () => {
+            // Build
+            const datetime = Date.now();
+            const resolvedInputParams = [{Key: 'Namespace', Value: `analysis-${datetime}`}];
+    
+            // Operate
+            const after = await lp.checkNamespace(resolvedInputParams, datetime);
+    
+            // Check
+            await expect(after[0].Value === resolvedInputParams[0].Value).toBeTruthy();
```
jn1119: Generally using `expect(string1).toBe(string2)` is better because you can see both the strings that don't match in errors in case your test fails- so it helps with debugging. With `toBeTruthy` I am guessing we would only learn if the condition failed or succeeded.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,79 @@
...
+        it('Static name that begins with analysis- should be transformed to end with number string', async () => {
+            // Build
+            const resolvedInputParams = [{Key: 'Namespace', Value: 'analysis-staticname'}];
+            const datetime = Date.now();
+    
+            // Operate
+            const after = await lp.checkNamespace(resolvedInputParams, datetime);
+    
+            // Check
+            await expect(Number(after[0].Value.split('-').pop())).not.toBe(NaN);
```
jn1119: I feel like we can use `toBe` here and to other test cases directly since we do want to check if we added datetime at the end which is a strong guarantee compared to checking if a number was added.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
maghirardelli: Added comment

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/.pnpm-debug.log
```diff
@@ -0,0 +1,20 @@
+{
```
nguyen102: this file should be removed. We don't want to add it to git. You can remove it with `git rm <filePath>`

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: Had to change the function to use the parameter datetime to do this, but I think that does make sense because it does make this test more straightforward and as you said, does get rid of any unknown function values.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: This is unrealistic that a user would be able to input the exact datetime that would be generated from the dynamic namespace generation, but I added it for robustness

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: Got it!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: Got it!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
```diff
@@ -356,6 +359,34 @@ class LaunchProduct extends StepBase {
     }
     return result.LaunchPathSummaries[0];
   }
+
+  /**
+   * Method to check if the resolved input parameter contained a static namespace param. If so, this method augments the
```
nguyen102: Nitpick text change: Check if the resolved input parameter contains a static namespace param. If so, augments the namespace to begin with 'analysis-' for permissions purposes (if it does not already start with that) and to end with a unique datetime string so Cloudformation doesn't create duplicate stacks for separate workspaces.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
```diff
@@ -356,6 +359,34 @@ class LaunchProduct extends StepBase {
...
+    const namespaceIndex = resolvedInputParams.findIndex(element => element.Key === 'Namespace');
+    let namespaceParam = resolvedInputParams[namespaceIndex].Value;
+
+    // Check to make sure the resolved namespace variable begins with 'analysis-' so our templates will allow it
+    if(!namespaceParam.startsWith('analysis-')){
+      namespaceParam = 'analysis-' + namespaceParam;
+    }
+
+    // Check to make sure the resolved namespace variable ends with a unique datetime string so it will be unique for each deployment of a configuration with a static namespace
+    if(namespaceParam.split('-').pop() !== datetime.toString()){
```
nguyen102, line 386: Correct me if I'm wrong, but I think this statement will always evaluate to `true`. I don't think a user can pass in a `dateTime` value with the exact time that this code segment is executed. We probably always want to use the `dateTime` value being passed in as the suffix.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
```diff
@@ -356,6 +359,34 @@ class LaunchProduct extends StepBase {
...
+   * Method to check if the resolved input parameter contained a static namespace param. If so, this method augments the
+   * namespace to begin with 'analysis-' for permissions purposes (if it does not already start with that) and to end with 
+   * a unique datetime string so Cloudformation doesn't make create duplicate stacks on seperate deployments of the same
+   * workspace configuration. 
+   * 
+   * @param resolvedInputParams 
+   * @param datetime 
+   * @returns {Promise<{Value: string, Key: string}[]>}
+   */
+  async checkNamespace(resolvedInputParams, datetime){
```
nguyen102: This method has a side effect, in that it's changing `resolvedInputParams`. Let's change the method name to reflect that. `checkNamespace` => `updateNamespaceIfNecessary`.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
maghirardelli: It will not evaluate to true if the resolveVarExpressions() function call does not change resolvedVars.namespace I believe. I had tested it and with a dynamic placeholder the code never entered into this if branch.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
maghirardelli: Yeah, I just learned the hard way and need to re-evaluate my logic in testing it because it does this. Is there a way to NOT make it do this? If it is easier to stop the changing of the parameter passed, I would prefer that so my tests would work *I think*. Thank you!

jn1119: We can always change `checkNamespace` method to `getNewNamespaceIfNecessary` to remove side effect. And just return the new namespace. In the start method we will have `resolvedInputParams[namespaceIndex].Value = getNewNamespaceIfNecessary(..`. For your unit tests you can just check the namespaces then and not need to look at resolvedInputParams.

We don't even need to pass the entire resolvedInputParams array to the method and just get away with passing the namespace.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
nguyen102: Oh I see. Namespace is defined in line 87. Seems like we can update `Namespace` value once instead of twice. This method can be 
```
getNamespace(defaultNamespace, resolvedInputParams) {
//If user provides us with a namespace, we format it with `analysis` prefix and `dateTIme` prefix. 

// If user does not provide namespace, we use the `stackName` variable which in this case would be `analysis-${datetime}`

// We pass in the stackName as the param `defaultNamespace`

}
```

On line 98 `resolveInputParams.namespace = getNamespace(stackName, resolvedInputParams`

And delete line 87.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
maghirardelli: Re: only passing the namespace, I tried to keep most of the new logic in the new function so I only had to unit test the new function and not the start function. 

I think I will do Tim's method name change and just redo the logic in my unit tests as the renaming of the namespace works this way--the unit tests just don't because I didn't know about the side effect.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
```diff
@@ -356,6 +363,35 @@ class LaunchProduct extends StepBase {
...
+   * so Cloudformation doesn't create duplicate stacks for separate workspaces.
+   * 
+   * @param resolvedInputParams 
+   * @param datetime 
+   * @returns {Promise<{namespaceParam:string, namespaceIndex:string}[]>}
+   */
+  async getNamespaceAndIndexIfNecessary(resolvedInputParams, datetime){
+    const namespaceIndex = resolvedInputParams.findIndex(element => element.Key === 'Namespace');
+    let namespaceParam = resolvedInputParams[namespaceIndex].Value;
+    console.log(namespaceParam);
```
jn1119: If you are logging something try to add more description. Something along the lines of `Current namespace in resolvedInputParams is ${namespaceParam}`

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
```diff
@@ -356,6 +363,35 @@ class LaunchProduct extends StepBase {
...
+    }
+
+    // Check to make sure the resolved namespace variable ends with a unique datetime string so it will be unique for each deployment of a configuration with a static namespace
+    if(namespaceParam.split('-').pop() !== datetime.toString()){
+      namespaceParam += '-' + datetime;
+    }
+    console.log(datetime);
+
+    console.log(namespaceParam);
+    return [namespaceParam, namespaceIndex];
```
jn1119: Usually it's recommended that you return a dictionary like { namespaceParam, namespaceIndex }. And you can access it by doing something like 

```
const { namespaceParam, namespaceIndex } = await this.getNamespaceAndIndexIfNecessary(resolvedInputParams, datetime);
```

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,161 @@
...
+            // Operate
+            const after = await lp.getNamespaceAndIndexIfNecessary(resolvedInputParams, datetime);
+            const after2 = await lp.getNamespaceAndIndexIfNecessary(resolvedInputParams2, datetime2);
+    
+            // Check
+            const afterNamespace = after[0];
+            const afterNamespace2 = after2[0];
+            await expect(afterNamespace.startsWith('analysis-')).toBeTruthy();
+            await expect(afterNamespace).toBe
+            await expect(afterNamespace).not.toBe(afterNamespace2);
```
jn1119: Aren't these checks redundant? Given that you added the last one. I think other tests have similar redundant checks.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,161 @@
...
+            const originalNamespace = resolvedInputParams[0].Value;
+    
+            // Operate
+            const after = await lp.getNamespaceAndIndexIfNecessary(resolvedInputParams, datetime);
+            const after2 = await lp.getNamespaceAndIndexIfNecessary(resolvedInputParams2, datetime2);
+    
+            // Check
+            const afterNamespace = after[0];
+            const afterNamespace2 = after2[0];
+            // await expect(after[0]).toBe(resolvedInputParams);
```
jn1119: Do we need these comments?

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/.pnpm-debug.log
jn1119: It seems like this file is still around. Is it required?

Nope. I thought I took it out as well but I will double check before pushing again.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,161 @@
...
+    
+        //     // Operate
+        //     const { afterNamespace, index } = await lp.getNamespaceAndIndexIfNecessary(resolvedInputParams, datetime);
+        //     const { afterNamespace2, index2 } = await lp.getNamespaceAndIndexIfNecessary(resolvedInputParams2, datetime2);
+    
+        //     // Check
+        //     await expect(afterNamespace.startsWith('analysis-')).toBeTruthy();
+        //     await expect(afterNamespace).not.toBe(afterNamespace2);
+        //     await expect(afterNamespace).toBe('analysis-' + originalNamespace + `-${datetime}`);
+        // });
```
jn1119: The commented code can be removed I think unless there is a reason we want to keep it around.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
maghirardelli: I actually meant to take this out--thanks for the advice and for flagging it for me!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
maghirardelli: So I tried to do this and it did not work. The tests failed and said the things returned were undefined. Do I have to name the things I assign to the returned values the same as I did within the function? I had said

`const { afterNamespace, afterIndex } = await this.getNamespaceAndIndexIfNecessary(...`

and it said that afterNamespace was undefined. Or, is it just good practice to name them the same?

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: Ah, they are. Thank you for pointing that out.

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: Nope! Taking them out!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: Yes, I will remove it. Sorry about that!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
maghirardelli: Nevermind--got it to work this time...don't know what I was doing differently the first time. I made it a dictionary!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
```diff
@@ -356,6 +363,32 @@ class LaunchProduct extends StepBase {
...
+  /**
+   * Check if the resolved input parameter contains a static namespace param. If so, augments the namespace to begin with 
+   * 'analysis-' for permissions purposes (if it does not already start with that) and to end with a unique datetime string 
+   * so Cloudformation doesn't create duplicate stacks for separate workspaces.
+   * 
+   * @param resolvedInputParams 
+   * @param datetime 
+   * @returns {Promise<{namespaceParam:string, namespaceIndex:string}[]>}
+   */
+  async getNamespaceAndIndexIfNecessary(resolvedInputParams, datetime){
```
Bingjiling: Function does not need to be async here

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,124 @@
...
+      });
+
+    // afterEach(() => {
+    //     const resolvedInputParams = [];
+    //     const datetime = 0;
+    //     const datetime2 = 0;
+    // })
+
+    describe('checkNamespace', () => {
+        it('Static name should be transformed to start with analysis- and end with number string and be unique between calls', async () => {
```
Bingjiling, line 30: Nice work catching all the test cases here!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,124 @@
...
+      const lp = new LaunchProduct({
+        step: { config: {} },
+        workflowPayload: new WorkflowPayload({ meta, input, workflowInstance: { steps: [] } }),
+      });
+
+    // afterEach(() => {
+    //     const resolvedInputParams = [];
+    //     const datetime = 0;
+    //     const datetime2 = 0;
+    // })
```
jn1119: I am assuming this can be deleted?

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
```diff
@@ -0,0 +1,124 @@
...
+            const datetime2 = Date.now() + 1;
+            const originalNamespace = resolvedInputParams[0].Value;
+    
+            // Operate
+            const {namespaceParam, namespaceIndex } = await lp.getNamespaceAndIndexIfNecessary(resolvedInputParams, datetime);
+            const {namespaceParam2, namespaceIndex2 } = await lp.getNamespaceAndIndexIfNecessary(resolvedInputParams2, datetime2);
+    
+            // Check
+            await expect(namespaceParam).not.toBe(namespaceParam2);
+            await expect(namespaceParam).toBe('analysis-' + originalNamespace + `-${datetime}`);
```
jn1119: nitpick: this can be written as 

```
`analysis-${originalNamespace}-${datetime}`
```

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
maghirardelli: Got it!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: Thanks!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: Yup, deleting it!

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/__test__/launch-product.test.js
maghirardelli: Thanks! Put this fix in to make it cleaner!

Bingjiling (dismissed):
Nice work! Remember to squash and clean up the commit message when merging :)

jn1119 (approved):

addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js
```diff
@@ -90,7 +91,13 @@ class LaunchProduct extends StepBase {
 
     // Read input params specified in the environment type configuration
     // The params may include variable expressions, resolve the expressions by using the resolveVars
+    // By doing this resolution, we might overwrite the dynamic, unique value defined above with a static name 
+    // that would not be unique between deployments of the same workspace configuration
     const resolvedInputParams = await this.resolveVarExpressions(envTypeConfig.params, resolvedVars);
+    // Additional layer to check the namespace is valid and unique. If not, make new namespace from
+    // static and get index of namespace param to change
+    const { namespaceParam, namespaceIndex } = this.getNamespaceAndIndexIfNecessary(resolvedInputParams, datetime);
+    resolvedInputParams[namespaceIndex].Value = namespaceParam;
```
nguyen102, line 100: Instead of assigning the value of `namspace` to `resolvedVars` twice, we could do it just once right? Is there a reason why we're setting the value twice? Setting the value twice could lead to user error in the future as they might not realized that `namespace` is being rewritten soon after it was initially set on line 87.

For example, we could set it once like this.
https://github.com/awslabs/service-workbench-on-aws/pull/615#discussion_r680236315

maghirardelli: I was setting it twice because the original set provides us some way to check if it was changed during the resolveVarExpressions/envTypeConfigService.mustfind call. I thought it was a simple way to check if the param was static or not. Do you still think I should find a way to only do it once?

Bingjiling: I think line 87 is needed actually, since in function [resolveVarExpressions](https://github.com/awslabs/service-workbench-on-aws/blob/3acd6d983db7d51f502a7a8e17793992be1234a7/addons/addon-environment-sc-api/packages/environment-sc-workflow-steps/lib/steps/launch-product/launch-product.js#L324), it'll be resolving a placeholder value `${namespace}` to what's set in line 87. If the value does not exist, lodash will throw error `namespace is not defined`. Tested it with this code block: 

```
let placeHolder = `${namespace}`;
result = _.template(placeHolder);
result(undefined);
```

nguyen102: that makes sense. thank you for the explanation

PR #641, opened 2021-08-06
fix: go bug during deployment is handled

maghirardelli:
Issue #, if available: GALI-1000

Description of changes: fixes the lack of handling for the error if Go lang is not installed during the deployment of SWB.

The script for my prompts:
When an error is caught, the command line will ask: "If you want to deploy Windows workspaces within Service Workbench, you must have Go lang installed before deployment. This is recommended."
If the user enters y, Y, yes, Yes, YES: "Failing building due to lack of Go lang. Please install Go lang." and throws and error with "Go lang not installed"
If the user enters n, N, no, No, NO: "Any Windows workspaces deployed will not function properly. You have been warned."
If the user enters another answer not listed here, it will re-ask the first question.

Checklist:

AS review ticket id:


addons/addon-base/packages/serverless-go-build-tools/index.js
```diff
@@ -71,6 +75,15 @@ class GoBuildTools {
     });
   }
 
+  // From: https://stackoverflow.com/questions/12042534/node-js-synchronous-prompt/46700053#46700053
```
SanketD92: optional: Instead of a link, could you describe what this method does as a one-liner?

addons/addon-base/packages/serverless-go-build-tools/index.js
```diff
@@ -79,6 +92,45 @@ class GoBuildTools {
...
+          // If Windows workspaces are desired, fail build
+          this.cli.warn(messagePrefix, 'Failing building due to lack of Go lang. Please install Go lang.');
+          readline.close();
+          validAnswer = true;
+          throw new Error('Go lang not installed');
+        } else if (response.match(/^(n|N|no|No|NO)$/)) {
+          // If Windows workspaces are not desired, log another warning and do not fail build
+          this.cli.warn(
+            messagePrefix,
+            'Any Windows workspaces deployed will not function properly. You have been warned.',
```
SanketD92: suggestion: 'Any Windows workspaces deployed will not function properly _until Go is installed and SWB is redeployed_. You have been warned.'

SanketD92 (dismissed):

addons/addon-base/packages/serverless-go-build-tools/index.js
```diff
@@ -79,6 +92,45 @@ class GoBuildTools {
       return;
     }
 
+    // Test out Go command
+    try {
+      execSync('go version');
```
Bingjiling, line 104: What would happen if go is installed but the go build failed for other reason? Will the deployment script success despite the failure?

maghirardelli: Yes, it would have done that to begin with as well. I suppose I could check the error in the second try-catch (at line 176) to make sure the error is missing go command. Would you like me to implement that?

addons/addon-base/packages/serverless-go-build-tools/index.js
maghirardelli: Included both!

addons/addon-base/packages/serverless-go-build-tools/index.js
maghirardelli: Got it!

addons/addon-base/packages/serverless-go-build-tools/index.js
Bingjiling: Yes, I think we can move what's in the catch block to line 176 directly, and catch go not installed and go build failed altogether. And we can use a more generic error message and remediation recommendation  so it takes care of both scenarios.

maghirardelli: We did it this way for the prompts to be synchronous.

maghirardelli: We did the prompting first so that we can have the prompts be synchronous. In the second try-catch, they are nested within a few operations so the prompt prints wrong.

Bingjiling (approved):

PR #642, opened 2021-08-06
fix: Various TRE bugfixes

nguyen102:
Issue #, if available:

Description of changes:

List of bug fixes below
* Unable to launch linux workspaces with AppStream is turned off
* When removing CIDR from AppStream enabled workspaces, our validator failed because it expects the CIDR field
* When user try to  view a workspace by clicking `Connect` there is no sign the the connection is happening because the `Connect` button does not change to the "loading" state
* User needs to turn off popup blocker for the domain so we can open a new AppStream tab for them. Instructions is added for users. 
* When AppStream is turned off, we're still asking users for AppStream specific configs
* We don't need to ask users `IsAppStreamEnabled`. We should be pulling that from our `settings` file.
Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/forms/AddUpdateAwsAccountForm.js
+++ b/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/forms/AddUpdateAwsAccountForm.js
@@ -36,13 +36,14 @@ const addUpdateBaseAwsAccountFormFields = {
 const addUpdateAwsAccountAppStreamFormFields = {
   appStreamFleetDesiredInstances: {
     label: 'AppStream Fleet Desired Instance',
-    placeholder: 'Number of users that can concurrently access a workspace through AppStream',
+    placeholder:
+      'Maximum number of concurrently running AppStream sessions. Each researcher uses one AppStream session when viewing a workspace',
     rules: 'required|integer',
   },
   appStreamDisconnectTimeoutSeconds: {
     label: 'AppStreamDisconnectTimeoutSeconds',
-    placeholder: 'The amount of time that a streaming session remains active after users disconnect',
-    rules: 'required|integer',
+    placeholder: 'The amount of time that a streaming session remains active after users disconnect.  (Minimum of 60)',
+    rules: ['required', 'integer', 'min:60'],
   },
   appStreamIdleDisconnectTimeoutSeconds: {
     label: 'AppStreamIdleDisconnectTimeoutSeconds',
--- a/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-config-vars-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-config-vars-service.js
@@ -307,9 +307,7 @@ class EnvironmentConfigVarsService extends Service {
       adminKeyPairName,
     };
 
-    if (enableEgressStore && enableEgressStore.toUpperCase() === 'TRUE') {
-      result.egressStoreIamPolicyDocument = JSON.stringify(egressStoreIamPolicyDocument);
-    }
+    result.egressStoreIamPolicyDocument = JSON.stringify(egressStoreIamPolicyDocument);
 
     return result;
   }
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-egress-store.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-egress-store.json
@@ -100,7 +100,6 @@
     "rev",
     "inWorkflow",
     "status",
-    "cidr",
     "createdAt",
     "updatedBy",
     "createdBy",
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-egress-store.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-egress-store.json
@@ -88,7 +88,12 @@
       "minLength": 1,
       "maxLength": 2048
     },
-    "hasConnections": { "type": "boolean" }
+    "hasConnections": { "type": "boolean" },
+    "studyRoles": {
+      "type": "object",
+      "properties": {},
+      "additionalProperties": true
+    }
   },
   "required": [
     "projectId",
--- a/addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/__tests__/env-type-config-service.test.js
+++ b/addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/__tests__/env-type-config-service.test.js
@@ -86,7 +86,7 @@ describe('EnvTypeService', () => {
       if (settingKey === 'isAppStreamEnabled') {
         return true;
       }
-      return undefined;
+      throw Error(`${settingKey} not found`);
     });
   });
 
--- a/addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/__tests__/env-type-config-service.test.js
+++ b/addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/__tests__/env-type-config-service.test.js
@@ -62,6 +62,7 @@ describe('EnvTypeService', () => {
     service = await container.find('envTypeConfigService');
     envTypeService = await container.find('envTypeService');
     s3Service = await container.find('s3Service');
+    const settingsService = await container.find('settings');
 
     // skip authorization
     service.assertAuthorized = jest.fn();
@@ -80,6 +81,13 @@ describe('EnvTypeService', () => {
       putObject: jest.fn().mockReturnThis(),
       promise: jest.fn().mockReturnThis(),
     };
+
+    settingsService.getBoolean = jest.fn(settingKey => {
+      if (settingKey === 'isAppStreamEnabled') {
+        return true;
+      }
+      return undefined;
+    });
   });
 
   describe('list function', () => {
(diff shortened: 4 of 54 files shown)


addons/addon-base-raas-ui/packages/base-raas-ui/src/models/forms/AddUpdateAwsAccountForm.js
```diff
@@ -36,13 +36,14 @@ const addUpdateBaseAwsAccountFormFields = {
 const addUpdateAwsAccountAppStreamFormFields = {
   appStreamFleetDesiredInstances: {
     label: 'AppStream Fleet Desired Instance',
-    placeholder: 'Number of users that can concurrently access a workspace through AppStream',
+    placeholder:
```
nguyen102, line 39: Various text enhancements suggested by Jolie

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-config-vars-service.js
```diff
@@ -307,9 +307,7 @@ class EnvironmentConfigVarsService extends Service {
       adminKeyPairName,
     };
 
-    if (enableEgressStore && enableEgressStore.toUpperCase() === 'TRUE') {
-      result.egressStoreIamPolicyDocument = JSON.stringify(egressStoreIamPolicyDocument);
-    }
+    result.egressStoreIamPolicyDocument = JSON.stringify(egressStoreIamPolicyDocument);
```
nguyen102, line 310: Our workspace cfn template expect a value for this attribute, and will error out if we don't provide it. If the attribute is `{}` then `EgressPolicy` is not added. Incase where `Egress` is turned off, the `egressStoreIamPolicyDocument` will be `{}`

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-egress-store.json
```diff
@@ -100,7 +100,6 @@
     "rev",
     "inWorkflow",
     "status",
```
nguyen102, line 102: With Jeet's change of removing `CIDR` from AppStream enabled accounts, we don't always pass in `cidr`. Therefore we need to remove it from here, or it will cause a validation error.

addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/env-type-config-service.js
```diff
@@ -241,6 +240,33 @@ class EnvTypeConfigService extends Service {
     return savedConfig;
   }
 
+  async checkAndUpdateParams(envType, config) {
```
nguyen102, line 244: Setting default values for the env params, depending on if AppStream is turned on or not.

addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/env-type-config-service.js
```diff
@@ -241,6 +240,33 @@ class EnvTypeConfigService extends Service {
...
+      key: 'IsAppStreamEnabled',
+      value: isAppStreamEnabled.toString(),
+    });
+    let params = [...updatedConfig.params];
+    if (isAppStreamEnabled) {
+      params.push({
+        key: 'AccessFromCIDRBlock',
+        value: '',
+      });
+    } else {
```
jn1119: I guess this is ok but was wondering if this is required if all the templates have a default value already?

addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/__tests__/env-type-config-service.test.js
```diff
@@ -80,6 +81,13 @@ describe('EnvTypeService', () => {
       putObject: jest.fn().mockReturnThis(),
       promise: jest.fn().mockReturnThis(),
     };
+
+    settingsService.getBoolean = jest.fn(settingKey => {
+      if (settingKey === 'isAppStreamEnabled') {
+        return true;
+      }
+      return undefined;
```
jn1119: throw `error` instead to replicate the settingsService?

addons/addon-environment-sc-ui/packages/environment-type-mgmt-ui/src/models/forms/CfnParamsForm.js
```diff
@@ -34,8 +34,22 @@ import { createForm } from '@aws-ee/base-ui/dist/helpers/form';
...
+  const filteredCfnParams = cfnParams.filter(cfnParam => {
+    const { ParameterKey } = cfnParam;
+    // We don't need to provide `IsAppStreamEnabled` as a config option to user because we can pull this value from settings on the backend
+    const keysToFilterOut = ['IsAppStreamEnabled'];
+    if (isAppStreamEnabled) {
+      keysToFilterOut.push('AccessFromCIDRBlock');
+    } else {
+      keysToFilterOut.push('EgressStoreIamPolicyDocument');
+      keysToFilterOut.push('SolutionNamespace');
+    }
```
jn1119: I didn't quite get why `EgressStoreIamPolicyDocument ` and `SolutionNamespace` need to be filtered out only if AppStream is disabled.

* Is there an advantage of making `SolutionNamespace` configurable when we are deriving it anyways from stackId.
* `EgressStoreIamPolicyDocument` seems like something that should either not be configurable or only show up when egress config is enabled.

Also update unit tests for this method?

scripts/app-stream/SETUP.md
```diff
@@ -29,7 +29,7 @@ Note: Please set up your [AWS Profile](https://docs.aws.amazon.com/cli/latest/us
 cd ~\Documents
 
 # Pull the Image Builder script from Github
-Invoke-WebRequest -Uri https://raw.githubusercontent.com/awslabs/service-workbench-on-aws/mainline/scripts/app-stream/buildImage.ps1 -OutFile buildImage.ps1
+Invoke-WebRequest -Uri https://raw.githubusercontent.com/awslabs/service-workbench-on-aws/feat-secure-workspace-egress/scripts/app-stream/buildImage.ps1 -OutFile buildImage.ps1
```
jn1119, line 34: make sure we add a task to change it back to mainline when releasing

addons/addon-environment-sc-ui/packages/environment-type-mgmt-ui/src/models/forms/CfnParamsForm.js
jn1119: Seems like this can be defined outside the filter

addons/addon-environment-sc-ui/packages/environment-type-mgmt-ui/src/models/forms/CfnParamsForm.js
```diff
@@ -34,8 +34,22 @@ import { createForm } from '@aws-ee/base-ui/dist/helpers/form';
...
+    // Include keys that are not in keysToFilterOut array
+    return !keysToFilterOut.includes(ParameterKey);
+  });
+
   const fields = {};
-  _.forEach(cfnParams, ({ ParameterKey, Description, DefaultValue }) => {
+  filteredCfnParams.forEach(({ ParameterKey, Description, DefaultValue }) => {
     const existingValue = _.get(_.find(existingParamValues, { key: ParameterKey }), 'value') || DefaultValue;
     if (!isAppStreamEnabled || (isAppStreamEnabled && !(ParameterKey === 'AccessFromCIDRBlock'))) {
```
jn1119: seems like `ParameterKey === 'AccessFromCIDRBlock'` can be removed now?

scripts/app-stream/SETUP.md
nguyen102: let me add todos for these

addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/__tests__/env-type-config-service.test.js
nguyen102: updated

addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/env-type-config-service.js
nguyen102: The validation of the fields will fail if we don't provide an attribute for them, which is why we're setting the values here. But yes, an alternative is to remove these fields from the validation schema and set default values in all of the templates.

addons/addon-environment-sc-ui/packages/environment-type-mgmt-ui/src/models/forms/CfnParamsForm.js
nguyen102: fixed

addons/addon-environment-sc-ui/packages/environment-type-mgmt-ui/src/models/forms/CfnParamsForm.js
nguyen102: removed

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-egress-store.json
jn1119: Can cidr be completely removed? I wonder why egress store needs it.

addons/addon-environment-sc-ui/packages/environment-type-mgmt-ui/src/models/forms/CfnParamsForm.js
nguyen102: Good point. Let's not make `EgressStoreIamPolicyDocument` and `SolutionNamespace` configurable by the customer. 

I'll also update the unit test.

jn1119 (approved):

addons/addon-environment-sc-api/packages/environment-type-mgmt-services/lib/environment-type/env-type-config-service.js
jn1119: Yeah, that's what I meant. I think all the templates have a default value for `AccessFromCIDRBlock` so the validation shouldn't fail. I approved your change though since this seemed a minor point.

PR #643, opened 2021-08-08
fix: update kms policy when no study selected

MBtea:
Issue #, if available:
Original issue: Researchers can not create object in egress store when workspace is launched in another account

Description of changes:
add fix to update egress store kms key policy when no study is selected

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
@@ -19,11 +19,18 @@ const Service = require('@aws-ee/base-services-container/lib/service');
 const { getSystemRequestContext } = require('@aws-ee/base-services/lib/helpers/system-context');
 const { processInBatches } = require('@aws-ee/base-services/lib/helpers/utils');
 
+const { addAccountToStatement } = require('../../../helpers/utils');
+
 /**
  * This service is responsible for allocating and de-allocating AWS resources for the environment so that
  * the environment can access what it needs, such as studies. This service implements the roles only
  * study access strategy.
  */
+const settingKeys = {
+  enableEgressStore: 'enableEgressStore',
+  egressStoreKmsKeyArn: 'egressStoreKmsKeyArn',
+  egressStoreKmsPolicyWorkspaceSid: 'egressStoreKmsPolicyWorkspaceSid',
+};
 class EnvironmentResourceService extends Service {
   constructor() {
     super();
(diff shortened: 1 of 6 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/data-source/access-strategy/roles-only/environment-resource-service.js
```diff
@@ -254,6 +261,71 @@ class EnvironmentResourceService extends Service {
     return s3Mounts;
   }
 
+  async allocateEgressResourcesOnly(requestContext, { studies: allStudies, memberAccountId }) {
```
nguyen102: I'm a little confuse with the name, `allocateEgressResourcesOnly`. It looks like we're updating the KMS policy if Egress store is turned on and there are no studies attached to the environment.

Should we rename this method to reflect that?

On another note, do we need two methods for updating KMS policy. Whether a study is attached or not, we'll need to update the KMS policy. Therefor can we combine `allocateEnvStudyResources` with `allocateEgressResourceOnly` to be one method? Maybe just a method called `updateKMSPolicyForEgress`.

MBtea: I think we can use `updateKMSPolicyForEgress` for the method name but I don't think we should merge two methods `allocateEnvStudyResources` and `allocateEgressResourceOnly` into one because:
1. There is if statement in `allocateStudyResources` which is invoked by `allocateEnvStudyResources` says it dose not want to handle the empty study situation.
2. Merging two method could bring risk for other services who has dependency on `allocateEnvStudyResources` method. 
I prefer to create a new method and update the method name to `updateKMSPolicyForEgress` without merging it with existing methods.

nguyen102: Ah got it. I had thought `allocateStudyResources` was a new method created for KMS policies as well. But if it's not it makes sense not to merge them.

Updating the method name makes sense. Btw, is there another method that update the KMS policy of environments that do have a study attached? Could that method be combined with this method so we have one method that handles updating KMS policy whether there are studies attached or not?

MBtea: Yeah, I had the same thought but I decided to not put the method into util file because the function which is updating the KMS policy is called `updateKMSPolicyForEgress` and `updateEgressKMSPolicy` but majority of them are getting setting's key and invoke kms APIs which makes them not a good fit to be put into util file.

nguyen102 (approved):

PR #651, opened 2021-08-12
Egress/pre deployment validation

MBtea:
Issue #, if available:

Description of changes:
Add pre-deployment validation step for checking BYOB studies have read write access. Block the deployment if user have existing BYOB study has read write access to external bucket. 

Use case:
- If this is the user's first time deployment, the pre-deployment stack will be deployed before backend stack with no error thrown. All other deployment goes as the old way.
- If this is **NOT** user's first time deployment and egress store feature is **NOT** enabled, the pre-deployment stack will be deployed before backend stack with no error thrown. All other deployment goes as the old way.
- If this is **NOT** user's first time deployment and egress store feature is enabled, the pre-deployment stack will be deployed before backend stack, the pre-deployment lambda would be invoked to check if there is existing BYOB study that has read&write access to the original bucket. If there is **NO** existing BYOB study that has read&write access to the original bucket, the deployment should proceed to backend deployment/update without error thrown. If there is existing BYOB study that has read&write access to the original bucket, the lambda in pre-deployment should throw error and block and deployment and no stack should be updated as followed.

Checklist:

AS review ticket id:


SanketD92 (commented):
Changes look good to me. Could you add to the PR description which test cases you've covered?

SanketD92 (approved):
LGTM

PR #663, opened 2021-08-17
test: Split integration and E2E tests for AppStream/NonAppStream

nguyen102:
Issue #, if available:

Description of changes:
* Added three subfolders for` end-to-end-tests` and `integration-tests`. The folders are: `appstream-disabled`, `appstream-enabled`, and `common`.
   * AppStream enabled tests will run all tests in `common` as well as tests in `appstream-enabled`
   * AppStream disabled tests will run all tests in `common` as well as tests in `appstream-disabled`
* I've moved the tests to their appropriate location. 
   * End to end workspace tests were moved to `appstream-disabled` as that current test is testing the AppStream disabled experience. We'll be writing new workspace test for `appstream-enabled`
   * For integration tests, studies and data-sources tests were move to `appstream-disabled` because those tests are [creating Studies](https://github.com/awslabs/service-workbench-on-aws/blob/b799b5bfb1305c07b04742725697c6b0e5f47bdc/main/integration-tests/support/resources/data-sources/studies.js#L50) with `accessType=readWrite` which is not allowed by AppStream [registerStudy code](https://github.com/awslabs/service-workbench-on-aws/blob/64c67609961295a343f5b90a859ff2046ee2c44e/addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js#L199)
* `package.json` scripts were update to include an option for running `appstream-enabled` tests.
* `run-integration-tests.sh` was updated to include an option for running `appstream-enabled` tests.

To review the changes, click the file filter and only include `.json`, `.md`, and `.sh`. The `.js` files are just files being moved into the correct folder

**Testing:** Check that the script for running tests ran the correct types of test (AppStream Enabled tests or AppStream disabled tests)

Checklist:

AS review ticket id:


main/integration-tests/__test__/api-tests/common/data-sources/register-ds-study.test.js
```diff
@@ -13,8 +13,8 @@
  *  permissions and limitations under the License.
  */
 
-const { runSetup } = require('../../../support/setup');
-const errorCode = require('../../../support/utils/error-code');
+const { runSetup } = require('../../../../support/setup');
+const errorCode = require('../../../../support/utils/error-code');
 
 describe('Register data source study scenarios', () => {
```
jn1119, line 19: We seem to have the same filename `register-ds-study.test.js` in both common and non-appstream path. I that expected? I didn't go through the contents in detail but it does seem like they are different.

nguyen102: Ah yes. Each of the two files is a subset of the original `register-ds-study.test.js`. The test cases that passes for the AppStream enabled path is in the common folder. The test cases that doesn't pass for AppStream enabled is in the `appstream-disabled` folder.

jn1119: I see, I wonder why `should return study registration information if admin` doesn't work for AppStream then? That doesn't seem intentional, right? Perhaps we should move it to common cases and open a separate ticket for debugging if it isn't intentional?

nguyen102: `should return study registration information if admin`fails because it trying to [create Studies](https://github.com/awslabs/service-workbench-on-aws/blob/b799b5bfb1305c07b04742725697c6b0e5f47bdc/main/integration-tests/support/resources/data-sources/studies.js#L50) with `accessType=readWrite` which is not allowed by AppStream [registerStudy code](https://github.com/awslabs/service-workbench-on-aws/blob/64c67609961295a343f5b90a859ff2046ee2c44e/addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js#L199). We don't allow that because if a user can create a study in which they have write access then they'll be able to export data without using the `EgressDataStore`.

jn1119: Ah, so it is an egress store setting as opposed to AppStream setting. I guess we should clarify that with folder structure then. Something like `appstream-and-egress-enabled` perhaps?

nguyen102: good point, I'll change the folder name to reflect that. I'll use `appstream-egress-enabled` to keep the name shorter

main/integration-tests/README.md
```diff
@@ -45,22 +48,32 @@ Once test resources and config file are created, you can run the integration tes
...
+
 - In config file `main/integration-tests/config/settings/<STAGE>.yml`
   - set `isBuildServer` to `false`
   - set `isLocal` to `false`
  
 ##### run all integration tests from the root directory with: 
+
+Run AppStream tests. AppStream and Egress should be enabled in the testing environment
+```bash
+$ scripts/run-integration-tests.sh <STAGE> AppStream
```
jn1119: Should we include egress in the stage- `AppStreamEgressEnabled`?

scripts/run-integration-tests.sh
```diff
@@ -8,7 +8,16 @@ pushd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null
 [[ ${UTIL_SOURCED-no} != yes && -f ./util.sh ]] && source ./util.sh
 popd > /dev/null
 
-ENV_NAME=$@
+PARAMETERS=$@
+PARAMETER_ARRAY=($PARAMETERS)
+ENV_NAME=${PARAMETER_ARRAY[0]}
+APPSTREAM_ENABLED=${PARAMETER_ARRAY[1]}
```
jn1119: Should parameter key and value also reflect egress?

scripts/run-integration-tests.sh
```diff
@@ -8,7 +8,16 @@ pushd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null
...
+
+if [ "$APPSTREAM_ENABLED" == "AppStream" ]; then
+  echo "Testing with AppStream Enabled"
```
jn1119: Mention Egress too?

main/integration-tests/README.md
nguyen102: updated

scripts/run-integration-tests.sh
nguyen102: updated

scripts/run-integration-tests.sh
nguyen102: updated

scripts/run-integration-tests.sh
```diff
@@ -31,7 +40,12 @@ if [ "$TEST_CONFIG_EXISTS" == true ]; then
     # shellcheck disable=SC1091
 
     printf "\n\nExecuting integration tests against env %s\n" "$ENV_NAME"
-    pnpm run intTest --recursive --if-present -- --stage="$ENV_NAME"
+
+    if [ "$APPSTREAM_ENABLED" == "AppStream" ]; then
```
jn1119: This flag should be `APPSTREAM_EGRESS_ENABLED`, right?

nguyen102: sorry about that, good catch. Updated.

jn1119 (approved):
LGTM. I guess just make sure you run the tests again once before pushing this out.

nguyen102:
Just tested again and all tests passes. I'll wait till after the demo on Friday to merge this in.

PR #669, opened 2021-08-18
feat: display Configuration Name and Instance Type on Workspace details card

maghirardelli:
Issue #, if available: GALI-284

Description of changes: Added two new rows at the bottom of the details table for the Workspace cards that display the Configuration Name and Instance Type of the workspace for researchers to better understand the capacity of the workspace:
Screen Shot 2021-08-18 at 4 17 34 PM

Checklist:

AS review ticket id:


SanketD92 (commented):
Unit and E2E tests needed.

Bingjiling:
We need to add these two fields to the env type config cards for selection before launching workspace as well.

maghirardelli:
Display for the select configuration cards:
Screen Shot 2021-08-19 at 1 54 38 PM

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/ScEnvironmentCard.js
```diff
@@ -31,7 +34,57 @@ class ScEnvironmentCard extends React.Component {
...
+  }
+
+  getInstanceTypeFromConfigParams(config) {
+    const configParams = Object.entries(config.params);
+    const instanceTypeElement = _.findIndex(configParams, param => {
+      const key = Object.entries(param[1])[0][1];
+      return key === 'InstanceType' || key === 'MasterInstanceType';
+    });
+    let instanceType;
+    // If no instance type param found
```
nguyen102: Should this comment be on line 58 where the `else` is?

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/ScEnvironmentCard.js
```diff
@@ -31,7 +34,57 @@ class ScEnvironmentCard extends React.Component {
...
+  getConfiguration(envTypeConfigId) {
+    const configsStore = this.getEnvTypeConfigsStore();
+    const config = configsStore.getEnvTypeConfig(envTypeConfigId);
+    return config;
+  }
+
+  getInstanceTypeFromConfigParams(config) {
+    const configParams = Object.entries(config.params);
+    const instanceTypeElement = _.findIndex(configParams, param => {
+      const key = Object.entries(param[1])[0][1];
```
nguyen102: Is there a possibility that the value doesn't exist in the param at the location specified? You can use the lodash [get](https://lodash.com/docs/#get) method to try and retrieve the value at the specified path. If that value is not there, you can set the default value to be undefined.

Same concept for line 57, incase the value is not there. In that case the default value should probably be `Not available`.

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/ScEnvironmentCard.js
```diff
@@ -74,12 +131,14 @@ class ScEnvironmentCard extends React.Component {
     );
 
     return (
-      <Table definition>
+      <Table data-testid="detail-table" definition>
```
nguyen102: let's be more specific with this id, and name it something like `environment-card-details-table`

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/__tests__/ScEnvironmentCard.test.js
```diff
@@ -0,0 +1,130 @@
...
+    component = wrapper.instance();
+
+    // mock error function because it doesn't function properly in enzyme
+    displayErrorMock.displayError = jest.fn(x => x);
+  });
+
+  it('should get env type configs store', async () => {
+    // BUILD
+
+    // OPERATE
```
nguyen102: If we don't need to BUILD anything then for the comment, we can do 
`//BUILD, OPERATE`

Otherwise it looks like we're missing code on line 78

main/end-to-end-tests/cypress/integration/workspaces.spec.js
```diff
@@ -54,6 +54,49 @@ describe('Launch a new sagemaker workspace', () => {
     launchWorkspace(ec2, 'EC2');
   });
 
+  it('should contain Configuration Name in Workspace Details Table', () => {
+    // const workspaces = Cypress.env('workspaces');
+    navigateToWorkspaces();
+    cy.get('[data-testid=workspaces]')
```
nguyen102: In end to end tests, we want to combine multiple test cases into one larger test. This is normally done to speed up the tests so that common steps aren't repeated. Let's make line 59 to 62 into a function, and we can call that function on line 49 and 55. So after we launch a workspace we check that `Configuration Name` and `Instance Type` is in the Details card for that workspace.

You'll want to modify the function `launchWorkspace` to return the `workspaceName`. Then with the `workspaceName` you can do
```
cy.contains(workspaceName)
      .parent()
      .get('[data-testid=workspaces]')
       .get('[data-testid=detail-table]')
      .contains('Configuration Name');
```

main/end-to-end-tests/cypress/integration/workspaces.spec.js
```diff
@@ -102,4 +145,39 @@ describe('Launch a new sagemaker workspace', () => {
...
+
+    // Select the type of environment you want to launch
+    cy.get('[data-testid=env-type-card]')
+      .contains(workspaceParam.workspaceTypeName)
+      .click();
+
+    // Click next
+    cy.get('button')
+      .contains('Next')
+      .click();
```
nguyen102: Nitpick: I think we can just put the content of this method in the method `checkInstanceTypeInSelectionCard` since we're not calling method `navigateToConfigurationCard` from anywhere else.

main/end-to-end-tests/cypress/integration/workspaces.spec.js
```diff
@@ -102,4 +145,39 @@ describe('Launch a new sagemaker workspace', () => {
...
+  };
+
+  const exitConfiguration = () => {
```
nguyen102: Does the code still work if we don't call `exitConfiguration`. At the end of the test, we don't have to reset it to a previous state. Each new test should `navigateToWorkspace` by itself and start from there. Therefore, we don't need to exit the configuration page.

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/ScEnvironmentCard.js
maghirardelli: Yes--i'll move it

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/ScEnvironmentCard.js
maghirardelli: I originally was using the lodash `get`, but with the layering of the params (it's got multiple layers which are not very nice data structures which is why I was using the `Object.entries()` to make them things I could easily index on), the `get` doesn't actually GET the necessary information that is paired with the key.

As for the default, I went with "Not available" because the code broke the UI if I made it undefined or it didn't find the right parameter. I thought this was a graceful way to display that case (which should never happen in the wild because you can't make a working workspace without having the configuration have an instance type).

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/ScEnvironmentCard.js
maghirardelli: Got it!

addons/addon-base-raas-ui/packages/base-raas-ui/src/parts/environments-sc/__tests__/ScEnvironmentCard.test.js
maghirardelli: Got it--will do

main/end-to-end-tests/cypress/integration/workspaces.spec.js
maghirardelli: Oh, I see what you meant the other day now. I'll do this.

main/end-to-end-tests/cypress/integration/workspaces.spec.js
maghirardelli: Got it! I thought modular was better in case it had other uses, but you're right that it doesn't so I will combine them.

main/end-to-end-tests/cypress/integration/workspaces.spec.js
maghirardelli: Nope. This was what caused the "side effect" I told you about when we talked on Friday. Since the test is mimicking creating a workspace without actually clicking create workspace, clicking the workspaces icon on the left menu (which is what `navigateToWorkspace` does) doesn't exit the process from the last test. So, this is why it was getting stuck and timing out. Although, could I possibly combine this check during the actual creation of workspaces in `launchWorkspace`? Or is that combining the tests of too many components into one case? I just fear it wouldn't be clear why it failed if we combine too many test cases.

nguyen102: That's a good idea, let's do this check when we're launching the test, since launching a test requires going pass the configuration screen anyways.

In Cypress it's ok to have larger tests with more assertions.
[Link here](https://docs.cypress.io/guides/references/best-practices#Creating-tiny-tests-with-a-single-assertion)

maghirardelli: Okay--I think I will also add a test to launch EMR in that case because we need to do that to test the configuration card information is there for that workspace as well.

maghirardelli: Doing all ^^ this worked! Now, there are only 3 E2E tests but they are pretty comprehensive!

nguyen102 (approved):
Nice! Looks good!

PR #678, opened 2021-08-27
Rstudio AppStream integration

jn1119:
Issue #, if available: GALI-1063

Description of changes:
1. Updated account onboarding template to create private hosted zone and SSM interface endpoint in case a custom domain is provided and AppStream is enabled. SSM is required for RStudio instances to publish secret that is used for creating a presigned URL.
2. Updated RStudio template to remove CIDR access when AppStream is enabled. Also output private IP here which is required for creating private DNS entries in Route53
3. Account onboarding workflow has been updated to include domainName in template if it exists. Also a new entry related to hosted zone id is being added in DynamoDB which is required for creating private entries in Route53 when provisioning RStudio instances
4. RStudio provisioning and termination workflow has been updated to ensure that we create/remove private DNS entries. Note that on start/stop we don't need to delete and recreate these entries since private IP address doesn't change unlike before in non-appstream case where public ip address changes when instance is restarted.

Testing done
1. Account onboarding updates
2. RStudio workspace provisioning when AppStream is enabled

Note: I will add a separate task for testing RStudio in non-AppStream case and use someone else's account since it would take quite a bit of time to set RStudio again.

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
+++ b/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml
@@ -97,6 +97,11 @@ Parameters:
     AllowedValues: [ALWAYS_ON, ON_DEMAND]
     Default: ON_DEMAND
 
+  DomainName:
+    Description: Optional custom Domain name to be created in Route53
+    Type: String
+    Default: ''
+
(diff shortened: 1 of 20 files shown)


SanketD92 (dismissed):
LGTM

SanketD92 (approved):

PR #680, opened 2021-08-27
Feat: Pre-populate variable values in input section of new workspace configuration

maghirardelli:
Issue #, if available: GALI-1092

Description of changes: pre-populated variables into the input section of creating a new workspace. Most people keep these default placeholder variables anyway so this will speedup making a new configuration for customers as well as reduce user error

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
+++ b/addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
@@ -63,6 +63,39 @@ class DropDown extends React.Component {
     this.optionsInState = _.concat({ text: data.value, value: data.value }, this.optionsInState);
   };
 
+  /**
+   * Uses the dropdown options to extract the placeholder variable values for the parameters when configuring
+   * a new workspace configuration.
+   * @param currentValue: holds the current value of the dropdown
+   * @param dropdownOptions: holds the options given in the droopdown
+   * @param field: the current field to find a defautl variable for
+   * @returns either the same currentValue or the variable value default to be displayed
+   */
+  getDefaultValue(currentValue, dropdownOptions, field) {
+    // if there is not a current value and there are valid dropdown options and the field is valid
+    if (currentValue === '' && dropdownOptions[0].key !== undefined && field.key !== undefined) {
+      // first element has no values
+      const currentValueIndex = dropdownOptions.slice(1).findIndex(option => {
+        const currentKeyOriginal = option.key;
+        const currentKey = currentKeyOriginal.charAt(0).toUpperCase() + currentKeyOriginal.slice(1);
+        const fieldKey = field.key;
+        return (
+          currentKey === fieldKey ||
+          (fieldKey.toLowerCase().includes(currentKeyOriginal) && currentKeyOriginal !== 'name') ||
+          currentKeyOriginal.slice(0, -2) === fieldKey.toLowerCase() ||
+          currentKeyOriginal === `admin${fieldKey.slice(0, 3)}Pair${fieldKey.slice(-4)}`
+        );
+      });
+      // if no value was found
+      if (currentValueIndex < 0) {
+        return '';
+      }
+      // offset of one because we sliced the first element out in our search
+      return dropdownOptions[currentValueIndex + 1].value;
+    }
+    return currentValue;
+  }
+
   render() {
     const {
       field,
@@ -119,7 +152,7 @@ class DropDown extends React.Component {
 
     const attrs = {
       id,
-      value,
+      value: this.getDefaultValue(value, mergeOptions, field),
 
       // applicable only when allowAdditions = true
       onAddItem: this.onAddItem,
--- a/addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
+++ b/addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
@@ -68,11 +68,10 @@ class DropDown extends React.Component {
    * Uses the dropdown options to extract the placeholder variable values for the parameters when configuring
    * a new workspace configuration.
    * @param currentValue: holds the current value of the dropdown
-   * @param dropdownOptions: holds the options given in the dropdown
    * @param field: the current field to find a default variable for
    * @returns either the same currentValue or the variable value default to be displayed
    */
-  getDefaultValue(currentValue, dropdownOptions, field) {
+  getDefaultValue(currentValue, field) {
     // Make a dict of the field values to the proper variables
     const fieldToVariableMap = {
       EncryptionKeyArn: '${encryptionKeyArn}',
@@ -148,7 +147,7 @@ class DropDown extends React.Component {
 
     const attrs = {
       id,
-      value: this.getDefaultValue(value, mergeOptions, field),
+      value: this.getDefaultValue(value, field),
 
       // applicable only when allowAdditions = true
       onAddItem: this.onAddItem,
--- a/addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
+++ b/addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
@@ -84,7 +84,7 @@ class DropDown extends React.Component {
       EnvironmentInstanceFiles: '${environmentInstanceFiles}',
       Subnet: '${subnetId}',
     };
-    // if there is not a current value and the field is a key in the above dict
+    // if current value is empty and the field is a key in the above dict
     if (currentValue === '' && field.key in fieldToVariableMap) {
       return fieldToVariableMap[field.key];
     }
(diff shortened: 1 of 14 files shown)


addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
```diff
@@ -63,6 +63,39 @@ class DropDown extends React.Component {
...
+   * @param field: the current field to find a defautl variable for
+   * @returns either the same currentValue or the variable value default to be displayed
+   */
+  getDefaultValue(currentValue, dropdownOptions, field) {
+    // if there is not a current value and there are valid dropdown options and the field is valid
+    if (currentValue === '' && dropdownOptions[0].key !== undefined && field.key !== undefined) {
+      // first element has no values
+      const currentValueIndex = dropdownOptions.slice(1).findIndex(option => {
+        const currentKeyOriginal = option.key;
+        const currentKey = currentKeyOriginal.charAt(0).toUpperCase() + currentKeyOriginal.slice(1);
```
SanketD92: Perhaps lodash [capitalize](https://lodash.com/docs/4.17.15#capitalize) method could achieve the same result, while making it more readable.

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
```diff
@@ -63,6 +63,39 @@ class DropDown extends React.Component {
...
+   * @param currentValue: holds the current value of the dropdown
+   * @param dropdownOptions: holds the options given in the droopdown
+   * @param field: the current field to find a defautl variable for
+   * @returns either the same currentValue or the variable value default to be displayed
+   */
+  getDefaultValue(currentValue, dropdownOptions, field) {
+    // if there is not a current value and there are valid dropdown options and the field is valid
+    if (currentValue === '' && dropdownOptions[0].key !== undefined && field.key !== undefined) {
+      // first element has no values
+      const currentValueIndex = dropdownOptions.slice(1).findIndex(option => {
```
SanketD92: Although the first element does not have any values, it doesn't optimize the findIndex operation a whole lot by slicing it off. Personally, I'd stay away from slicing if not required.

Let me know if I'm misunderstanding the logic.

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
```diff
@@ -63,6 +63,39 @@ class DropDown extends React.Component {
...
+        const currentKeyOriginal = option.key;
+        const currentKey = currentKeyOriginal.charAt(0).toUpperCase() + currentKeyOriginal.slice(1);
+        const fieldKey = field.key;
+        return (
+          currentKey === fieldKey ||
+          (fieldKey.toLowerCase().includes(currentKeyOriginal) && currentKeyOriginal !== 'name') ||
+          currentKeyOriginal.slice(0, -2) === fieldKey.toLowerCase() ||
+          currentKeyOriginal === `admin${fieldKey.slice(0, 3)}Pair${fieldKey.slice(-4)}`
```
SanketD92: Instead of slicing, could we simply compare it to the actual key we're expecting?
for eg. `currentKeyOriginal === 'adminKeyPairName'` (assuming that's the key we're trying to get)

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
```diff
@@ -63,6 +63,39 @@ class DropDown extends React.Component {
     this.optionsInState = _.concat({ text: data.value, value: data.value }, this.optionsInState);
   };
 
+  /**
+   * Uses the dropdown options to extract the placeholder variable values for the parameters when configuring
+   * a new workspace configuration.
+   * @param currentValue: holds the current value of the dropdown
+   * @param dropdownOptions: holds the options given in the droopdown
```
SanketD92: typo `dropdown`

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
```diff
@@ -63,6 +63,39 @@ class DropDown extends React.Component {
...
+   * @param field: the current field to find a defautl variable for
```
SanketD92: typo `default`

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
SanketD92: Could you also add comments for lines 82 through 86 to explain these checks?

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
maghirardelli: The problem is that the first element has a key value that is always the field key. I don't know why. But the value and the name attributes for that first element are blank. So the findIndex ALWAYS selects the first one, which is not the behavior we want

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
maghirardelli: Yeah, I will add comments. 

We only care about cond 1: `currentKeyOriginal === 'adminKeyPairName'` if cond 2:`fieldKey === KeyName` because the dropdown options for all dropdown fields will have adminKeyPair so ever field would return that if it came before the proper variable we want in the list. So, I could make it an and of those two conditions (cond 1 && cond 2) or just leave it like this. Let me know which you prefer for readability/in general.

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
maghirardelli: Got it!

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
maghirardelli: Got it!

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
maghirardelli: That doesn't work exactly. For example, if `currentKeyOriginal = iamPolicyDocument`, we want `currentKey = IamPolicyDocument` not `currentKey = Iampolicydocument` because the key for that field is IamPolicyDocument. I think I could just `.toLowerCase()` both values and then compare if you think that makes more sense than what I have in line 80. LMK!

SanketD92: I see. Yes, lowercase both and compare makes more sense.

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
SanketD92: Okay, we can leave it as-is then.

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
SanketD92: Yes, AND-ing makes sense if two conditions are interdependent. 

Optional: If there is a way to remove slicing and compare with lowercase strings, that would help with readability IMO. (If not possible, add examples in comments)

maghirardelli: That condition is for fieldKey = Subnet and currentKeyOriginal = subnetId. VPC also needs to be mapped to vpcId. Those two strings would never be the same unless we sliced on Id from currentKeyOriginal or we added Id to fieldKey. Adding would make it more specific to their cases I guess. LMK what you think about that case.

I'll make the change for the last case.

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
maghirardelli: Got it! I'll make it do that

addons/addon-base-ui/packages/base-ui/src/parts/helpers/fields/DropDown.js
```diff
@@ -63,6 +64,33 @@ class DropDown extends React.Component {
...
+      VPC: '${vpcId}',
+      AccessFromCIDRBlock: '${cidr}',
+      S3Mounts: '${s3Mounts}',
+      Namespace: '${namespace}',
+      KeyName: '${adminKeyPairName}',
+      IamPolicyDocument: '${iamPolicyDocument}',
+      EnvironmentInstanceFiles: '${environmentInstanceFiles}',
+      Subnet: '${subnetId}',
+    };
+    // if there is not a current value and the field is a key in the above dict
```
SanketD92: optional: "if current value is empty and ..."

SanketD92 (dismissed):

SanketD92 (dismissed):
Ship it!

SanketD92 (approved):

PR #692, opened 2021-09-01
fix: Use bucket region if available while mounting BYOB studies

jn1119:
Issue #, if available:

Description of changes:

* Note that this change only works when isAppStream is false. When AppStream is enabled cross-region studies will not work since access to S3 happens via a private gateway endpoint. According to [S3 documentation](https://docs.aws.amazon.com/vpc/latest/privatelink/vpc-endpoints-s3.html) private endpoint cannot be used to access buckets cross-region

> Endpoints currently do not support cross-Region requests—ensure that you create your endpoint in the same Region as your bucket. You can find the location of your bucket by using the Amazon S3 console, or by using the get-bucket-location command. Use a Region-specific Amazon S3 endpoint to access your bucket; for example, mybucket.s3.us-west-2.amazonaws.com. For more information about Region-specific endpoints for Amazon S3, see Amazon Simple Storage Service (S3) in Amazon Web Services General Reference. If you use the AWS CLI to make requests to Amazon S3, set your default Region to the same Region as your bucket, or use the --region parameter in your requests.

Checklist:

AS review ticket id:

### DIFF
--- a/main/solution/post-deployment/config/environment-files/bin/mount_s3.sh
+++ b/main/solution/post-deployment/config/environment-files/bin/mount_s3.sh
@@ -78,19 +78,26 @@ do
             printf 'Mounting internal study "%s" at "%s"\n' "$study_id" "$study_dir"
             goofys --region $region --acl "bucket-owner-full-control" "${s3_bucket}:${s3_prefix}" "$study_dir"
         else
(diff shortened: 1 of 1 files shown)


SanketD92 (approved):
LGTM

nguyen102 (approved):

PR #703, opened 2021-09-10
fix: no sagemaker autostop lag

maghirardelli:
Issue #, if available: GALI-965

Description of changes: Reduces lag of Sagemaker instances showing up as available when they have autostopped or are autostopping to a few seconds from a few minutes (or infinity in some cases)

Checklist:

AS review ticket id:


addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
```diff
@@ -1422,4 +1422,112 @@ describe('EnvironmentSCService', () => {
...
+      // OPERATE
+      const sagemakerUpdated = await service.pollAndSyncSageMakerStatus(
+        roleArn,
+        externalId,
+        sagemakerInstances,
+        requestContext,
+      );
+
+      // CHECK
+      await expect(sagemakerUpdated).not.toEqual(empty);
```
nguyen102: Nitpick: What's the actual value that is returned here? Can we just check it returns that value?

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
```diff
@@ -1422,4 +1422,112 @@ describe('EnvironmentSCService', () => {
...
+
+    it('should finish updating before returning', async () => {
+      // BUILD
+      const sagemakerInstances = { 'notebook-instance-name': {} };
+      service.pollSageMakerRealtimeStatus = jest.fn(() => {
+        return { 'notebook-instance-name': 'InService' };
+      });
+      service.updateStatus = jest.fn(async () => {
+        // sleep
+        await new Promise(r => setTimeout(r, 4000));
```
nguyen102: Can we make this `1000` and the original code without the bug fix will still fail this unit test? If yes, let's shorten the time frame so the test runs faster.

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
maghirardelli: Oh yeah, I think we can do that too. I'll change

addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
maghirardelli: let me see. If so, I will make it shorter.

nguyen102 (approved):

Issue #699, opened 2021-09-08
Increase the number of supported studies per workspace


aGutierrezSacristan:
**Describe the bug**
When creating a workspace in SWB with multiple studies associated, when having more than 10 studies, the workspace is not created, and stayed in permanent "pending" status, with not even the option to terminate it.

**To Reproduce**
1. Select 44 studies. 
2. Create a SageMaker Notebook-v3 Workspace Type
Large-Sized Sagemaker
Instance Type : ml.t2.2xlarge
vCpus : 8
Memory : 32GiB
Auto-stop idle-time : 60min

**Expected behavior**
That the workspace is created and available and if there is a maximum number of studies, or some other related issues, to get some type of message and to have the chance to terminate it from SWB. 

**Screenshots**
![Screen Shot 2021-09-08 at 2 11 23 PM](https://user-images.githubusercontent.com/19646530/132562938-6226ca02-03ab-40f2-ab22-74b4a9df5c5b.png)

**Versions (please complete the following information):**
Release Version installed: 3.3.1

**Additional context**
If worked when selecting 10 studies but it is always in pending when selecting all the studies that we require (in this specific case around 40)

jol-ster:
Thank you for reporting this.  Can you please let me know which category of studies are involved?  There are three types..
1) MyStudies
2) Organizational Studies
3) Studies that are "Data Sources" from the Bring Your Own Bucket registration

If it is a mix of types, that would be helpful to understand as well.

aGutierrezSacristan:
Thanks @jol-ster 
The study types are: Organization studies.

SanketD92:
Hello @aGutierrezSacristan, could you check a few things in your account for us that would help us in debugging further:

1. BYOB (or Data Source) studies are also listed under the Organization studies category in SWB, but users cannot view their files from the SWB UI. The way SWB handles these studies is different than the rest of the Organization studies. Could you confirm the studies you selected did not include any BYOB studies?
2. Could you check if there was any error on the CloudFormation stack corresponding to this workspace, usually starts with `'SC-'` on the hosting account (ie. account linked to the SWB project)?

carvantes:
Hi,

There are 2 issues:

1.- Mounting many studies causes errors. There isn't a specific max number of studies, but the study s3 locations are concatenated as a param and eventually breach the max CFN string param size of 4096

[...]

PR #705, opened 2021-09-10
fix: properly handle very long error messages on env update

carvantes:
Issue #, if available: https://github.com/awslabs/service-workbench-on-aws/issues/699

Provisioning workspaces with lots of studies resulted in the provisioning workflow crashing and the workspace to be stuck in PENDING. See the detailed description on the related gh issue.

With this fix provisioning still fails, but now the workflow finishes and the workspace is correctly set to FAILED and displays the error message.

Screen Shot 2021-09-10 at 15 24 42

Checklist:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
@@ -605,6 +605,63 @@ describe('EnvironmentSCService', () => {
       expect(storageGatewayService.updateStudyFileMountIPAllowList).not.toHaveBeenCalled();
     });
 
+    it('should truncate excessively long error messages and succeed to update', async () => {
+      // BUILD
+      const requestContext = {
+        principalIdentifier: {
+          username: 'uname',
+          ns: 'user.ns',
+        },
+      };
+
+      const oldEnv = {
+        id: 'oldId',
+        name: 'exampleName',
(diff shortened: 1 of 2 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/__tests__/environment-sc-service.test.js
```diff
@@ -605,6 +605,63 @@ describe('EnvironmentSCService', () => {
       expect(storageGatewayService.updateStudyFileMountIPAllowList).not.toHaveBeenCalled();
     });
 
+    it('should truncate excessively long error messages and succeed to update', async () => {
```
jn1119, line 608: Would this test fail if you didn't have the truncate fix? I am not sure where we are testing if the truncate happened in this test before the ddb update.

carvantes: yes, it would fail at: https://github.com/awslabs/service-workbench-on-aws/blob/70d989172013fe9213194723e791f230d730ca6f/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-service.js#L458

That because the validation schema has a max length of 2048:
https://github.com/awslabs/service-workbench-on-aws/blob/b799b5bfb1305c07b04742725697c6b0e5f47bdc/addons/addon-base-raas/packages/base-raas-services/lib/schema/update-environment-sc.json#L25-L27

jn1119 (approved):

PR #710, opened 2021-09-15
feat: swb main study role

nguyen102:
Issue #, if available:

Description of changes:
Allows egress store in member account to access Egress Store S3 bucket in main account

* Create a role called `study-<egressId>` in the main account when a workspace is created
* Create a policy called `study-<egressId>` in the main account and attach it to the study role above
* Delete the policy and role above when the workspace is terminated and an egress request to export data is not pending
* Modify `egress-store-object-handler` lambda function to not mark EgressStore as `isAbleToSubmitEgressRequest` if the egress store is empty.
* Added unit tests for adding new study role and deleting the role. 

**Tests**
Launch a new workspace, and role/policy was created.
Terminated the new workspace, and role/policy was deleted

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
@@ -27,6 +27,7 @@ const {
   getRevisedS3Statements,
   removeAccountFromStatement,
 } = require('../helpers/utils');
+const { StudyPolicy } = require('../helpers/iam/study-policy');
 
 const settingKeys = {
   tableName: 'dbEgressStore',
@@ -566,45 +567,21 @@ class DataEgressService extends Service {
     const kmsArn = await this.getKmsKeyIdArn();
     const memberAccountId = await this.getMemberAccountId(requestContext, egressStoreId);
     const mainAccountRoleName = this.getMainAccountEgressStoreRole(egressStoreId);
-    const permissionPolicy = {
-      Version: '2012-10-17',
-      Statement: [
-        {
-          Sid: 'S3StudyReadAccess',
-          Effect: 'Allow',
-          Action: [
-            's3:GetObject',
-            's3:GetObjectVersion',
-            's3:GetObjectTagging',
-            's3:AbortMultipartUpload',
-            's3:ListMultipartUploadParts',
-            's3:PutObject',
-            's3:PutObjectAcl',
-            's3:PutObjectTagging',
-            's3:DeleteObject',
-            's3:DeleteObjectVersion',
-          ],
-          Resource: `arn:aws:s3:::${egressStoreBucketName}/${egressStoreId}`,
-        },
-        {
-          Sid: 'S3StudyList',
-          Effect: 'Allow',
-          Action: ['s3:ListBucket', 's3:ListBucketVersions'],
-          Resource: `arn:aws:s3:::${egressStoreBucketName}`,
-          Condition: {
-            StringLike: {
-              's3:prefix': `${egressStoreId}/*`,
-            },
-          },
-        },
-        {
-          Sid: 'studyKMSAccess',
-          Action: ['kms:Decrypt', 'kms:DescribeKey', 'kms:Encrypt', 'kms:GenerateDataKey', 'kms:ReEncrypt*'],
-          Effect: 'Allow',
-          Resource: kmsArn,
-        },
-      ],
+    const permissionBoundaryArn = this.settings.get('permissionBoundaryPolicyStudyBucketArn');
+
+    const egressStudyPolicy = new StudyPolicy();
+
+    const study = {
+      bucket: egressStoreBucketName,
+      folder: [egressStoreId],
+      permission: {
+        read: true,
+        write: true,
+      },
+      kmsArn,
     };
+    egressStudyPolicy.addStudy(study);
+    const permissionPolicy = egressStudyPolicy.toPolicyDoc();
 
     const iam = await this.getIAM();
     const createPolicyResponse = await iam
@@ -630,7 +607,7 @@ class DataEgressService extends Service {
         }),
         Path: '/',
         RoleName: mainAccountRoleName,
-        PermissionsBoundary: createPolicyResponse.Policy.Arn,
+        PermissionsBoundary: permissionBoundaryArn,
       })
       .promise();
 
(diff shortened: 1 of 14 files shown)


main/solution/backend/config/infra/cloudformation.yml
```diff
@@ -623,49 +635,8 @@ Resources:
...
-              - !GetAtt [ResourceUsagesDb, Arn]
-              - !If
-                - EnableEgressStore
-                - !Sub 'arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${self:custom.settings.dbEgressStore}'
-                - !Ref 'AWS::NoValue'
-              - !If
-                - EnableEgressStore
-                - !Sub 'arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${self:custom.settings.dbEgressStore}/index/*'
-                - !Ref 'AWS::NoValue'
+              - !Sub 'arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${self:custom.settings.dbPrefix}-*'
```
nguyen102, line 640: The policy for this role was getting too long, and CFN was throwing an error. This change was made to shorten the policy length.

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -537,6 +546,118 @@ class DataEgressService extends Service {
...
+   * @param requestContext
+   * @param egressStoreId
+   * @returns role Arn
+   */
+  async createMainAccountEgressStoreRole(requestContext, egressStoreId) {
+    const egressStoreBucketName = this.settings.get('egressStoreBucketName');
+    const kmsArn = await this.getKmsKeyIdArn();
+    const memberAccountId = await this.getMemberAccountId(requestContext, egressStoreId);
+    const mainAccountRoleName = this.getMainAccountEgressStoreRole(egressStoreId);
+    const permissionPolicy = {
```
jn1119: Can you look into [study-policy.js](https://github.com/awslabs/service-workbench-on-aws/blob/feat-secure-workspace-egress/addons/addon-base-raas/packages/base-raas-services/lib/helpers/iam/study-policy.js) and see if that works for you? Most likely it should, but in case it doesn't we can extend it if we need to rather than hardcoding the policy here.

main/solution/backend/config/infra/cloudformation.yml
```diff
@@ -767,6 +738,24 @@ Resources:
...
             Resource:
               - !Sub 'arn:aws:appstream:${AWS::Region}:${AWS::AccountId}:image/ServiceWorkbench*'
+          - Effect: Allow
+            Action:
+              - iam:CreateRole
+            Resource:
+              - !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:role/study-*'
+            Condition:
+              StringLike:
+                iam:PermissionsBoundary: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:policy/study-*'
```
jn1119: Can we create a permission boundary here which has read/write access to egress bucket and reference it here. It's better to avoid wildcards in permission boundary since there can be an overly permissive policy created which starts with `study-`.

main/solution/post-deployment/src/lambdas/egress-store-object-handler/handler.js
```diff
@@ -45,11 +45,23 @@ const handlerWithContainer = async (container, event) => {
   }
   const dataEgressService = await container.find('dataEgressService');
   const egressStoreInfo = await dataEgressService.getEgressStoreInfo(egressStoreId);
-  if (!egressStoreInfo.isAbleToSubmitEgressRequest) {
+  if (!egressStoreInfo.isAbleToSubmitEgressRequest && S3FileWasAdded(event.Records)) {
```
jn1119, line 48: Is there a test for this?

nguyen102: added

main/solution/backend/config/infra/cloudformation.yml
nguyen102: Added the permissionBoundary

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
nguyen102: Updated to use study-policy

main/solution/backend/config/infra/cloudformation.yml
```diff
@@ -767,6 +740,68 @@ Resources:
...
+              - !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:role/study-*'
+          - Effect: Allow
+            Action:
+              - iam:CreatePolicy
+            Resource:
+              - !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:policy/study-*'
+
+  PermissionBoundaryPolicyStudyBucket:
+    Type: AWS::IAM::ManagedPolicy
+    Properties:
```
nguyen102, line 768: I didn't add a condition here to only create this Policy if `EnableEgressStore` is true, because then we would have to do additional `if` statements on line 748. I wasn't able to get the CF template to compile successfully with an if Statement for line 748. In terms of value given effort spent, I think it should be ok to not add additional `if` statements for the Egress state, given that we will soon transition internal studies to use roles as well. When that happens this role and policy will be needed for both Egress/NonEgress state.

jn1119: If EnableEgressStore is false,  it just looks odd to have a policy created in main account for a bucket that doesn't even exist. Couple of things to try:

1. We either add an If/Else condition on the resource itself on this permission boundary policy. We can go with either of the following approaches here: 
    * (Preferable for now but not sure if possible) When egressStore is enabled we allow read/write access to egress bucket and when egressStore is disabled our permission boundary will be empty suggesting that the IAM role created using the permission boundary doesn't have any access.
    * (Includes what we want to do in future eventually) When egressStore is enabled we allow read/write access to egress bucket and only read access to study bucket. When egressStore is disabled our permission boundary will only have read/write access to study bucket

2. I know it is hard to get nested If conditions in CFN templates. Have you tried applying the If condition for line 743 instead? I had to do something similar for account-onboard yaml template [here](https://github.com/awslabs/service-workbench-on-aws/blob/feat-secure-workspace-egress/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/onboard-account.cfn.yml#L265-L275). The main difference is that in my case I had a condition for an inline policy and what you have here is a managed policy. Not sure if you already tried this.

nguyen102: I tried doing Option 1.a, but that didn't work since with `egressStore`disabled there was an empty policy and CF didn't like it. 

I'm going to go with approach 2 instead.

addons/addon-base-raas/packages/base-raas-services/lib/data-egress/data-egress-service.js
```diff
@@ -215,17 +215,13 @@ class DataEgressService extends Service {
...
     if (egressStoreStatus.toUpperCase() === PROCESSING_STATUS_CODE) {
       throw this.boom.forbidden(
-        `Egress store: ${egressStoreInfo.id} is still in processing. The egress store is not terminated and the workspce can not be terminated before egress request is processed.`,
+        `Egress store: ${egressStoreInfo.id} is still in processing. The egress store is not terminated and the workspace can not be terminated before egress request is processed.`,
         true,
       );
-    } else if (egressStoreStatus.toUpperCase() === PROCESSED_STATUS_CODE || isEgressStoreNotTouched) {
-      // ONLY terminate the egress store if it has been processed or the egress store is empty
-
+    } else if ([PROCESSED_STATUS_CODE, CREATED_STATUS_CODE].includes(egressStoreStatus.toUpperCase())) {
```
jn1119, line 224: Was this branch not being unit tested before? Can we add a test?

nguyen102: added

jn1119 (approved):

PR #713, opened 2021-09-16
fix: externalId requirement for add account form

SanketD92:
Issue #, if available:
`externalId` not being stored in DDB when required.

Description of changes:
Adding `externalId` form field for aws accounts.

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/forms/AddUpdateAwsAccountForm.js
+++ b/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/forms/AddUpdateAwsAccountForm.js
@@ -31,6 +31,11 @@ const addUpdateBaseAwsAccountFormFields = {
     placeholder: 'Type description for this AWS account',
     rules: 'required|string',
   },
+  externalId: {
+    label: 'External ID',
+    placeholder: 'Type external ID for this AWS account',
+    rules: 'required|string|between:1,300',
+  },
 };
 
 const addUpdateAwsAccountAppStreamFormFields = {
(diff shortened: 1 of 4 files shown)


maghirardelli (commented):
Are we not going to add a note that it cannot be changed in the CloudFormation stack input part or things would not work? I don't know if this is that important but Jeet had mentioned providing a warning.

SanketD92:
> Are we not going to add a note that it cannot be changed in the CloudFormation stack input part or things would not work? I don't know if this is that important but Jeet had mentioned providing a warning.

I'm going to test this out. I have a feeling we may not need to since it will be updated in both places (CFN stack and DDB entry) at once.

nguyen102:
Doesn't SWB automatically set the `externalId` to be `workbench`? Given that if a user sets a value besides `workbench`, the code will not work, should we not allow them to change this value?

We could either have the UI automatically send a default value for `externalId` or on the API side we set it to be `workbench` for `externalId`

SanketD92:
> Doesn't SWB automatically set the `externalId` to be `workbench`? Given that if a user sets a value besides `workbench`, the code will not work, should we not allow them to change this value?
> 
> We could either have the UI automatically send a default value for `externalId` or on the API side we set it to be `workbench` for `externalId`

No, the externalId value was not present from the form which is why it's never entered in the DDB entry (you'll get `undefined` if you try to get from the DDB entity), while it's read as the string "undefined" on the CfN stack that add account process creates. 

No matter what the user changes the external ID to in their CfN stack parameters, the cfn status check role cannot be assumed since the externalId does not match.

nguyen102 (approved):

PR #715, opened 2021-09-16
fix: enableEgressStore flag is boolean, not string

maghirardelli:
Issue #, if available: GALI-1071

Description of changes: enableEgressStore flag in the configuration file for SWB is now a boolean and not a string. Serverless accepts this by invoking a script to convert the boolean and read the string value of "true" if it was true and "false" otherwise. This affected the logic of many files that used this flag as well as unit tests of those files. All are updated. Tested with unit tests as well as locally and through the API.

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/data-source/__tests__/environment-resource-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/data-source/__tests__/environment-resource-service.test.js
@@ -494,6 +494,23 @@ describe('EnvironmentResourceService', () => {
 
   describe('addToKmsKeyPolicy', () => {
     it('add new principal to KMS policy with no principals', async () => {
+      environmentResourceService._settings = {
+        get: settingName => {
+          if (settingName === 'studyDataKmsPolicyWorkspaceSid') {
+            return 'KMS Policy';
+          }
+          if (settingName === 'studyDataKmsKeyArn') {
+            return 'studyKmsKeyAlias';
+          }
+          return undefined;
+        },
+        getBoolean: settingName => {
+          if (settingName === 'enableEgressStore') {
+            return false;
+          }
+          return undefined;
+        },
+      };
       AWSMock.mock('KMS', 'describeKey', (params, callback) => {
         expect(params).toMatchObject({
           KeyId: 'studyKmsKeyAlias',
@@ -545,9 +562,6 @@ describe('EnvironmentResourceService', () => {
           if (settingName === 'studyDataKmsKeyArn') {
             return 'studyKmsKeyAlias';
           }
-          if (settingName === 'enableEgressStore') {
-            return 'true';
-          }
           if (settingName === 'egressStoreBucketName') {
             return 'test-egressStoreBucketName';
           }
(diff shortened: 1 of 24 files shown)


main/config/settings/.defaults.yml
```diff
@@ -204,4 +204,8 @@ egressNotificationTopicName: ${self:custom.settings.namespace}-EgressTopic
 # =========================================================================================================
 
 # Determine whether workspace should be accessible only through AppStream
+# NOTE: Once the isAppStremEnabled is set to true, the AppStrem feature will be enabled and can NOT be toggled off.
```
SanketD92: AppStream typo (x2)

addons/addon-base-raas/packages/base-raas-services/lib/data-source/__tests__/environment-resource-service.test.js
```diff
@@ -494,6 +494,23 @@ describe('EnvironmentResourceService', () => {
 
   describe('addToKmsKeyPolicy', () => {
     it('add new principal to KMS policy with no principals', async () => {
+      environmentResourceService._settings = {
```
SanketD92: Why are these needed now but weren't needed earlier?

SanketD92 (commented):
Since the changes touch multiple components, could you list out the manual testing you did in the description (both UI and backend)?

main/config/settings/.defaults.yml
maghirardelli: Got it--will fix

addons/addon-base-raas/packages/base-raas-services/lib/data-source/__tests__/environment-resource-service.test.js
maghirardelli: Because there is an environmentResourceService setup in the beforeEach but it didn't include getBoolean so I needed to have the implementation of the mocked getBoolean function. Putting it here overrode the beforeEach one and meant that the environmentResourceService mock this test was using didn't have the methods that the beforeEach one did. Thanks for pointing this out and making me figure this out though--I can just put the getBoolean mock in the beforeEach one which reduces code repetition and still works!

maghirardelli:
> Since the changes touch multiple components, could you list out the manual testing you did in the description (both UI and backend)?

Below are the tests I ran after making the changes everywhere to treat this flag as a boolean instead of a string.

First, I re-ran every unit test suite that used this value. Most of them I had to either change the mocked return value (from `'false'` to `false` or `'true'` to `true`) as well as change the method that was mocked. Before, the mocking of this value was a `get()` mocking and that had to be changed to a `getBoolean()` mocking. All of these unit tests passed after my code changes and testing set-up changes.

Second, I deployed an environment (upgraded from non TRE) with AppStream and Egress Store enabled. I was able to see and click the "Egress Store" button in the UI. Additionally, hitting the `/api/data-egress/:id` API fielded a successful response.

Third, I deployed an environment with AppStream enabled but not Egress Store. After creating a workspace, I confirmed that the "Egress Store" button did not appear in the Workspace details UI page. Additionally, I hit the `/api/data-egress/:id` API and got a not successful response in line with what would be expected if the egress store was not enabled.

Lastly, I took the last environment and turned the Egress Store on and redeployed. After this, the "Egress Store" button appeared on Workspace details UI. For the already made workspace, clicking this button did nothing but give an error (which is to be expected). After adding a new configuration for a workspace type, I created a new workspace with this new configuration. It's detail page displayed the "Egress Store" button and clicking it displayed the table correctly. Additionally, I hit the `/api/data-egress/:id` API with the new workspace id and got a success response.

SanketD92 (approved):
LGTM

PR #716, opened 2021-09-20
feat: dynamic version number from Changelog

maghirardelli:
Issue #, if available: GALI-824

Description of changes: Now, instead of having the developer have to change the version number and version date in the .defaults.yml file for every release, as part of the deployment script, the version number and date will be added as parameters to the global stage file (location change suggestion from Sanket) as gotten from the CHANGELOG.md file. 

HOWEVER, Sanket would like to change our process a little bit for releases to allow for the version deployed in an environment to be flagged as "Beta." This would involve the developer cutting the release to, after the release has been published, push a change to the CHANGELOG file that is a new header "## Beta" at the top of the file ahead of the latest release information. Case matters (I can change that if you want but figured it was easier to keep it consistent). This way, anyone pulling from mainline will know that their version is not released and potentially unstable. In this case, the date next to Beta will be the date of deployment like this:
Screen Shot 2021-09-20 at 1 12 50 PM

I didn't make this an SSM param (which was one of Jeet's initial ideas when this ticket was created) because I think this change propagates the proper information to everything that it needs to.

Checklist:

AS review ticket id:

### DIFF
--- a/scripts/check-and-add-beta.sh
+++ b/scripts/check-and-add-beta.sh
@@ -2,7 +2,7 @@
 set -e
 
 # Get the first header (not Changelog) in CHANGELOG.md
-versionLine="$(cat CHANGELOG.md | grep -m 1 "[0-9]\.[0-9]\.[0-9]\|Beta")"
+versionLine="$(cat CHANGELOG.md | grep -m 1 "[0-9]\+\.[0-9]\+\.[0-9]\+\|Beta")"
 
 # Check if it contains the word Beta
 if (echo "$versionLine" | grep -q "Beta")
@@ -16,7 +16,7 @@ else
     # Add Beta header to changelog
     echo "Need to add to changelog"
     # Get latest release number
-    latestReleaseVersion="$(cat CHANGELOG.md | grep -o "[0-9]\.[0-9]\.[0-9]" | head -n 1)"
+    latestReleaseVersion="$(cat CHANGELOG.md | grep -o "[0-9]\+\.[0-9]\+\.[0-9]\+" | head -n 1)"
     # Create ed file
     echo "5i
 ## Beta
@@ -31,6 +31,6 @@ q" > add-beta.ed
     rm add-beta.ed
     # Commit and push new changelog
     git add CHANGELOG.md
-    git commit -m "Add Beta"
+    git commit -m "docs: Add Beta"
     git push origin develop
 fi
--- a/scripts/check-and-add-beta.sh
+++ b/scripts/check-and-add-beta.sh
@@ -20,7 +20,7 @@ else
     # Create ed file
     echo "5i
 ## Beta
-[This deployment is in beta. Click here to see changes since ${latestReleaseVersion}.](https://github.com/maghirardelli/service-workbench-on-aws-github-actions/compare/v${latestReleaseVersion}...mainline)
+[This branch is in beta. Click here to see changes since ${latestReleaseVersion}.](https://github.com/maghirardelli/service-workbench-on-aws-github-actions/compare/v${latestReleaseVersion}...mainline)
 
 .
 w
--- a/scripts/check-and-add-beta.sh
+++ b/scripts/check-and-add-beta.sh
@@ -15,9 +15,12 @@ else
     git checkout develop
     # Add Beta header to changelog
     echo "Need to add to changelog"
+    # Get latest release number
+    latestReleaseVersion="$(cat CHANGELOG.md | grep -o "[0-9]\.[0-9]\.[0-9]" | head -n 1)"
     # Create ed file
     echo "5i
 ## Beta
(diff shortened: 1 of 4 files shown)


maghirardelli:
This is the display when the version is in Beta now:
Screen Shot 2021-09-22 at 5 10 39 PM

Additionally, we increased the scope to add another github action after develop is merged into mainline. This step runs if the last commit message does not contain "chore(release):" (because this is the format of the standard-version generated commit messaged) or "Add Beta" (because this is the commit message of the commit made by this step so there is no need to check immediately after the Beta change has taken place). Here is how the three kinds of commits that are relevant react with this:
1. It is a release commit. This step doesn't run.
2. It is the first push to develop after the release has been cut. This step runs, merging develop into mainline and then changing changelog to have the "## Beta" header. The changelog is then commited (by a github-actions bot) to develop, which re-triggers the deploy-integ.yml files and this entire pipeline to run. On this second run, this step does not run. If we want, we can add the `if: "!contains(github.commit, "Add Beta")"` to the other steps in the pipeline so they always would be bypassed in this second run. Should we add this?
3. It is a push to develop after "## Beta" has already been added to changelog. This step runs but the script finds the Beta header and does not change changelog or commit anything to develop.

Here, we see how the release didn't trigger an "Add Beta" commit but the commit after did:
Screen Shot 2021-09-22 at 4 39 00 PM

Here, we see that the "Add Beta" commit skipped the last step in the pipeline:
Screen Shot 2021-09-22 at 4 46 09 PM

Here, we see that it was "actions-user" that committed the change to changelog:
Screen Shot 2021-09-22 at 12 57 21 PM

maghirardelli:
This is the additional information that is displayed below the Beta header in CHANGELOG.md. The version number in the text and in the hyperlink are dynamically scraped from CHANGELOG.md as well: 
Screen Shot 2021-09-27 at 2 56 45 PM

maghirardelli:
This is the additional information in CHANGELOG. The version number in the print and in the link are dynamically scraped from CHANGELOG as well:
Screen Shot 2021-09-27 at 4 19 15 PM
The hyperlink leads to: https://github.com/maghirardelli/service-workbench-on-aws-github-actions/compare/v1.0.10...mainline

scripts/check-and-add-beta.sh
```diff
@@ -0,0 +1,36 @@
...
+.
+w
+q" > add-beta.ed
+    # Change CHANGELOG.md with ed file
+    ed CHANGELOG.md < add-beta.ed
+    # delete ed file
+    rm add-beta.ed
+    # Commit and push new changelog
+    git add CHANGELOG.md
+    git commit -m "Add Beta"
```
jn1119: In order to be consistent with semantic versioning probably add the prefix `docs: `?

scripts/environment-deploy.sh
```diff
@@ -53,6 +53,56 @@ pushd "$SOLUTION_DIR/post-deployment" > /dev/null
 $EXEC sls invoke -f postDeployment -l -s "$STAGE"
 popd > /dev/null
 
+# Get the first header (not Changelog) in CHANGELOG.md
+versionLine="$(cat CHANGELOG.md | grep -m 1 "[0-9]\.[0-9]\.[0-9]\|Beta")"
```
jn1119: Use `[0-9]+` for the first number in case our versioning crosses `9`.

scripts/environment-deploy.sh
```diff
@@ -53,6 +53,56 @@ pushd "$SOLUTION_DIR/post-deployment" > /dev/null
...
+
+# Is there a stage.yml file?
+FILE=main/config/settings/${STAGE}.yml
+if [ -f "$FILE" ]
+then
+    # Yes-->Is there a versionDate and versionNumber key?
+    if (cat "$FILE" | grep -q "versionDate") && (cat "$FILE" | grep -q "versionNumber")
+    then
+        # Yes-->Are they different from above?
+        oldVersionNumber="$(cat "$FILE" | grep -o "[0-9]\.[0-9]\.[0-9]\|Beta" | head -n 1)"
```
jn1119: Same here, we might need to extend the first number to account for more than one digit.

scripts/check-and-add-beta.sh
maghirardelli: Will do!

scripts/environment-deploy.sh
maghirardelli: Got it. I am going to add that to all three digits in case we need it.

scripts/environment-deploy.sh
maghirardelli: Yup, I will put `[0-9]+` for all the release number `grep` calls

maghirardelli: Had to escape the special character so it was `[0-9]\+` for each number in `<number>.<number>.<number>`

jn1119 (approved):

PR #721, opened 2021-09-23
feat: Configurable open data filter

nguyen102:
Issue #, if available:

Description of changes:
* Update `open-data-scrape-handler` to pull Open Data Tag Filters from `config` so customers can choose what Open Data set to include. The studies list API will also use the new filters
* Update `open-data-scrape-handler` to use the [Open Data Registry](https://registry.opendata.aws/registry-open-data/). This simplifies our code so we don't have to pull one S3 Metadata file for each Open Data.
Checklist:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
@@ -32,12 +32,6 @@
     "uploadLocationEnabled": {
       "type": "boolean"
     },
-    "sha": {
-      "type": "string",
-      "maxLength": 64,
-      "pattern": "^([A-Fa-f0-9]{40})$",
-      "description": "A unique identifier for Open Data in MD5 hash, hexadecimal"
-    },
     "tags": {
       "type": "array"
     },
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
@@ -38,6 +38,9 @@
       "pattern": "^([A-Fa-f0-9]{40})$",
       "description": "A unique identifier for Open Data in MD5 hash, hexadecimal"
     },
+    "tags": {
+      "type": "array"
+    },
     "resources": {
       "type": "array",
       "items": [
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/update-study.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/update-study.json
@@ -34,6 +34,9 @@
       "maxLength": 2048,
       "pattern": "^(arn:aws[a-zA-Z-]*:iam::[0-9]{12}:role[/]{1}[a-zA-Z0-9-]+)$"
     },
+    "tags": {
+      "type": "array"
+    },
     "resources": {
       "type": "array",
       "items": [
--- a/addons/addon-base-raas/packages/base-raas-services/lib/schema/update-study.json
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/schema/update-study.json
@@ -23,12 +23,6 @@
       "type": "string",
       "description": "Leaving length and pattern blank to accommodate open data"
     },
-    "sha": {
-      "type": "string",
-      "maxLength": 64,
-      "pattern": "^([A-Fa-f0-9]{40})$",
-      "description": "A unique identifier for Open Data in MD5 hash, hexadecimal"
-    },
     "appRoleArn": {
       "type": "string",
       "maxLength": 2048,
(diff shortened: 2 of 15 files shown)


main/config/settings/.defaults.yml
```diff
@@ -179,3 +179,5 @@ versionDate: '2021-09-16'
 
 # Metadata provided with AWS SDK calls
 customUserAgent: 'AwsLabs/SO0144/${self:custom.settings.versionNumber}'
+
```
SanketD92, line 176: Could we point to a URL or include something in the example.yml file to show a list of all available tags in the open data registry?

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
```diff
@@ -497,17 +497,22 @@ class StudyService extends Service {
...
+        result = (
+          await this._query()
+            .index(this.categoryIndex)
+            .key('category', category)
+            .limit(1000)
+            .projection(fields)
+            .query()
+        ).filter(study => {
+          return openDataTagFilters.some(filterTag => study.tags.includes(filterTag));
+        });
```
jn1119, line 515: Just curious if this filter can be server side as opposed to client side?

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
```diff
@@ -497,17 +497,22 @@ class StudyService extends Service {
   }
 
   async list(requestContext, category, fields = []) {
```
jn1119, line 499: what is the behavior when during workspace provisioning an open data study is attempted to be attached that we have in DynamoDB but not included in the filters? That situation won't arise from the UI but someone can arbitrarily try to include any study from provision API directly.

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-study.json
```diff
@@ -34,6 +34,9 @@
       "maxLength": 2048,
       "pattern": "^(arn:aws[a-zA-Z-]*:iam::[0-9]{12}:role[/]{1}[a-zA-Z0-9-]+)$"
     },
+    "tags": {
+      "type": "array"
+    },
```
jn1119, line 33: Is there a test for this?

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
nguyen102: We discussed this offline and decided to go with filtering in the lamba function because filtering using filter expression is much more complex and the latency difference would be small.

addons/addon-base-raas/packages/base-raas-services/lib/schema/update-study.json
nguyen102: Updated tests in `study-service.test.js` to include `tags`. Without the change in this json the tests would fail.

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
nguyen102: Updated the create workspace API to not allow users to create workspace with an Open Study that is not in the list of valid Open Studies.

main/config/settings/.defaults.yml
nguyen102: Yep, I'll add that in the comment. The list of available tags is [here](https://github.com/awslabs/open-data-registry/blob/main/tags.yaml)

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
```diff
@@ -32,12 +32,6 @@
     "uploadLocationEnabled": {
       "type": "boolean"
     },
-    "sha": {
```
jn1119, line 35: Do we need to change OpenAPI documentation

main/config/settings/.defaults.yml
```diff
@@ -173,3 +173,6 @@ createServiceCatalogPortfolio: true
...
+# Available tags: https://github.com/awslabs/open-data-registry/blob/main/tags.yaml
```
jn1119, line 177: Probably worth mentioning that updating the filter doesn't result in unmounting of studies from existing workspaces

addons/addon-base-raas/packages/base-raas-services/lib/schema/create-study.json
nguyen102: updated the OpenApi docs

main/config/settings/.defaults.yml
nguyen102: added

jn1119 (approved):

PR #731, opened 2021-10-05
fix: Disable public IP address for Linux and Windows in AppStream

nguyen102:
Issue #, if available:

Description of changes:
fix: Disable public IP address for Linux and Windows in AppStream

**Test**
* AppStream Enabled
   * Checked in the AWS console that Linux and Windows instances does NOT have public IP addresses
   * Log into windows, egress store there, egress write succeed
   * Log into Linux, egress store there, egress write succeed
* AppStream Disabled
   * Checked in the AWS console that Linux and Windows instance does have public IP Addresses 
   * Log into Windows, read/write from study succeed
   * Log into Linux, read/write from study succeed
Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-windows-instance.cfn.yml
+++ b/addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-windows-instance.cfn.yml
@@ -266,7 +266,8 @@ Resources:
             KmsKeyId: !Ref EncryptionKeyArn
             DeleteOnTermination: true
       NetworkInterfaces:
-        - DeviceIndex: '0'
+        - AssociatePublicIpAddress: !If [ AppStreamEnabled, 'false', 'true' ]
+          DeviceIndex: '0'
           GroupSet:
             - !Ref SecurityGroup
             - !If
@@ -274,10 +275,6 @@ Resources:
               - Fn::ImportValue: !Sub "${SolutionNamespace}-SwbVPCDefaultSG"
               - !Ref "AWS::NoValue"
           SubnetId: !Ref Subnet
-        - !If
-          - AppStreamEnabled
-          - AssociatePublicIpAddress: false
-          - AssociatePublicIpAddress: true
       Tags:
         - Key: Name
           Value: !Join ['-', [Ref: Namespace, 'ec2-windows']]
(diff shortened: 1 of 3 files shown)


addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-windows-instance.cfn.yml
```diff
@@ -265,15 +266,20 @@ Resources:
             KmsKeyId: !Ref EncryptionKeyArn
             DeleteOnTermination: true
       NetworkInterfaces:
-        - AssociatePublicIpAddress: true
```
jn1119, line 268: Can't we just replace `true` with `AppStreamEnabled` flag?

jn1119 (commented):
Might as well update the RStudio template as well

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-windows-instance.cfn.yml
nguyen102: I just tried it now. When SWB tries to spin up the instance, CF fails with this error.
`Encountered unsupported boolean value AppStreamDisabled`

We want to use `AppStreamDisabled` because `AssociatePublicIpAddress` should be `true` only when  AppStream is disabled.

nguyen102:
updated to include rstudio as well.

addons/addon-base-raas/packages/base-raas-cfn-templates/src/templates/service-catalog/ec2-windows-instance.cfn.yml
jn1119: I actually meant using “not of AppStreamEnabled” flag or just the “AppStreamDisabled” flag. But it’s surprising that CFN resolves to unsupported Boolean value even though something like “!Ref AppStreamDisabled” should be a boolean

nguyen102: I was able to get it working by writing the code like this. Which convert the boolean to a string.

`- AssociatePublicIpAddress: !If [ AppStreamEnabled, 'false', 'true' ]`

jn1119 (approved):

PR #738, opened 2021-10-11
fix: build ami version bug

maghirardelli:
Issue #, if available: GALI-1167

Description of changes: add a step to call a script to add the versionNumber and versionDate to the stage file in the serverless plugin that is used to build the AMIs. Before, if the machine images were built before the environment was deployed, it would error out as not having the versionNumber in the settings. This script is the same that is used in environment-deploy.sh. I manually tested this by getting the expected error without the added method being called and seeing a successful creation of 4 AMIs with the added method. Additionally, I spun up an instance using the Linux AMI created to test that the build was successful and I could connect and use it.

Checklist:

AS review ticket id:


addons/addon-base-raas/packages/serverless-packer/index.js
```diff
@@ -112,6 +118,16 @@ class ServerlessPackerPlugin {
     });
     return varArgs;
   }
+
+  async addVersionInfo(stageName) {
+    this.serverless.cli.log('Adding versionNumber and versionDate to stage file');
+
+    try {
+      execSync(`cd ../../../ && ./scripts/get-release-info.sh ${stageName}`);
```
jn1119: Does this command also change the directory for end user who ran the original `build-image` command? If yes, can we run `../../../scripts/get-release-info.sh` instead?

maghirardelli: `get-release-info.sh` assumes it is being called from the `service-workbench-on-aws` folder because inside it has paths for `CHANGELOG.md` and the stage file folder hardcoded. What is the case that someone would can the "original" `build-image` command? Is that different than `pnpx sls build-image`?

jn1119 (dismissed):

addons/addon-base-raas/packages/serverless-packer/index.js
maghirardelli: `execSync` makes a child process for those commands so the end user's directory stays the same

Bingjiling (dismissed):
LGTM

Bingjiling (approved):

PR #743, opened 2021-10-12
fix: Various bugfixes

nguyen102:
Issue #, if available:

Description of changes:
* When there is a preexisting non TRE account being upgraded to a TRE account, the `awsAccountService.update` method is called. This method should share AppStream with the member account just like how `awsAccountService.create` does.
* PolicyWorkFlowLoopRunner should not depend on PermissionBoundaryPolicyStudyBucket

**Test**
* AppStream image was shared with member account when updating AWS account
* Deploying latest code to CFN succeeded

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-accounts-service.test.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/aws-accounts/__tests__/aws-accounts-service.test.js
@@ -392,6 +392,59 @@ describe('AwsAccountService', () => {
       encryptionKeyArn: 'arn:aws:kms::key/someKey',
     };
 
+    beforeEach(() => {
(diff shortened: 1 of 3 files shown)


jn1119 (approved):

jn1119 (commented):
Do we have unit tests where we ensure that `update` and `create` methods are calling `shouldShareAppStreamImageWithMemberAccount` and consequently `shareAppStreamImageWithMemberAccount` if sharing is possible?

nguyen102:
The unit tests check if we call `shareAppStreamImageWithMemberAccount` but not if we call `shouldShareAppStreamImageWithMemberAccount`. Calling `shareAppStreamImageWithMemberAccount` is the main logic that we want to check, since that's the important method that needs to be called given the correct state.

PR #767, opened 2021-10-21
test: Codepipeline to Github notification

nguyen102:
Issue #, if available:

Description of changes:
Wait for Codepipeline deployment and integration tests to finish before merging code to `mainline`. CodePipeline will send a status check to GH once the Pipeline successfully executes all steps. The resources to configure that is outlined in this [blog post](https://aws.amazon.com/blogs/devops/aws-codepipeline-build-status-in-a-third-party-git-repository/). We then have a custom script in our GitHub workflow that checks the status of the current commit before merging the code to `mainline`. 

**Tests**
I tested this code in my repo, and it worked as expected. [Link](https://github.com/nguyen102/service-workbench-on-aws/runs/3970547394?check_suite_focus=true) to what an example run would look like. 

Checklist:

AS review ticket id:


main/cicd/cicd-pipeline/config/infra/cloudformation.yml
```diff
@@ -358,6 +358,22 @@ Resources:
             Resource:
               - !Sub 'arn:aws:iam::${AWS::AccountId}:role/${self:custom.settings.namespace}-cicd-pipeline-AppDeployerRole-*'
               - !Sub 'arn:aws:iam::${AWS::AccountId}:policy/${self:custom.settings.namespace}-cicd-pipeline-PolicyAppDeployer-*'
+          - !If
+            - RunTestsAgainstTargetEnv
+            - Effect: 'Allow'
+              Action:
+                - servicecatalog:CreateProduct
```
SanketD92, line 402: Why do we need `servicecatalog` permissions for integration tests? I'd assume they're only needed for installing/uninstalling.

Also, you might want to pull in changes from develop (I've added similar perms for the `DeleteTargetEnv` step as well). Only some `servicecatalog` permissions require wildcard resources.

nguyen102: I was confused as to why the Integration tests were failing as well, but before this change I was getting this error when the tests were running.
```
"errorMessage": "User: arn:aws:sts::684277579687:assumed-role/int-va-sw-cicd-pipeline-AppDeployerRole-1XJ4CRKLF9QQT/AWSCodeBuild-be46b256-10b8-43bb-a2a4-2812e506b283 is not authorized to perform: servicecatalog:CreateProduct on resource: arn:aws:servicecatalog:us-east-1:684277579687:*/* because no identity-based policy allows the servicecatalog:CreateProduct action",
```

Upon digging into the code there are [create service catalog](https://github.com/awslabs/service-workbench-on-aws/blob/b99092458c7ee11f9b7540f7d1bbf898dd90744f/main/integration-tests/__test__/api-tests/common/workspace-service-catalogs/create-workspace-service-catalog.test.js#L60) tests as well as delete service catalog tests. 

Yep, I'll pull in your changes. Looks like we have similar permissions, and we both have conditional for when to add the permissions.

SanketD92: That makes sense. I just remembered we hadn't set up the previous CodePipeline instance with integration tests, which is why we didn't need these perms before.

scripts/check-commit-status/checkCommitStatus.js
```diff
@@ -0,0 +1,40 @@
...
+        // Latest commit status is at the top of the list
+        // https://docs.github.com/en/rest/reference/repos#list-commit-statuses-for-a-reference
+        const response = await octokit.rest.repos.listCommitStatusesForRef({owner, repo, ref: sha});
+        const commitStatuses = response.data.filter((status) => {return status.context === commitStatusToCheckFor});
+        if (commitStatuses.length > 0) {
+            latestStatus = commitStatuses[0].state;
+        }
+        await new Promise(resolve => setTimeout(resolve, waitTimePerApiCallInSeconds * 1000));
+        iteration++;
+        console.log('Iteration', iteration);
```
SanketD92: Could we add more details while we're logging? Specifically for the commit we're looking for.

nguyen102: Updated

SanketD92 (approved):
LGTM

PR #774, opened 2021-10-26
fix: s3 bucket versioning and https access

maghirardelli:
Issue #, if available: GALI-1202, GALI-1200

Description of changes: enable versioning on all buckets and add public policy statement to enforce https traffic only on all buckets and all items in buckets. Tested that versioning is enabled in the console of the buckets after but not before deployment with changes and that the bucket policy was updated to include protection for buckets after re-deployment.

Checklist:

AS review ticket id:


jn1119 (commented):
Have you also verified the change by bringing up workspaces and verified that study access works as expected from at minimum Windows and Linux workspaces?

nguyen102 (approved):
LGTM, assuming manual testings passed. Launching workspaces and checking you can access the studies in those workspaces.

PR #775, opened 2021-10-26
test: Add check for secrets in PRs

nguyen102:
Description of changes:
[Git Secrets](https://github.com/awslabs/git-secrets) is a software tool built by Amazon to prevent users from committing their credentials to GitHub.

Uses Git Workflow to add a check on  any published PRs. If `git secrets` catches that a credential was committed it will fail the check, and prints a message in the build log telling the user to revoke the exposed credentials.

Error Message is shown when check fails. Click [here](https://github.com/awslabs/service-workbench-on-aws/runs/4012581919?check_suite_focus=true) for example. Expand the `Print remediation message if git secret scan found a leaked secret` section


SanketD92 (approved):
LGTM

PR #782, opened 2021-10-29
fix: more secure traffic policy

maghirardelli:
Issue #, if available: GALI-1200

Description of changes: In the cloudformation templates, add another resource to enforce HTTPS traffic to S3 buckets provisioned by SWB. This fix only applies to the buckets that are static i.e. not changed by the user programmatically through the SWB UI later on. Those buckets will be fixed in another PR.

Testing: Deployed to both a TRE and non-TRE env and was able to make a study, launch a workspace with that study mounted, access the workspace and the contents of the study bucket.

Checklist:

AS review ticket id:

### DIFF
--- a/main/solution/backend/config/infra/cloudformation.yml
+++ b/main/solution/backend/config/infra/cloudformation.yml
@@ -181,11 +181,13 @@ Resources:
       PolicyDocument:
         Version: '2012-10-17'
         Statement:
-          - Sid: Deny requests that do not use TLS
+          - Sid: Deny requests that do not use TLS/HTTPS
             Effect: Deny
             Principal: '*'
             Action: s3:*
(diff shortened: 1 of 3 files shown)


main/solution/backend/config/infra/cloudformation.yml
```diff
@@ -437,6 +443,34 @@ Resources:
               StringNotEquals:
                 s3:signatureversion: 'AWS4-HMAC-SHA256'
 
+  EgressNotificationBucketPolicy:
```
jn1119, line 446: Were you able to verify that egress notifications are working as expected?

jn1119 (approved):

main/solution/backend/config/infra/cloudformation.yml
maghirardelli: Yes

PR #797, opened 2021-11-11
fix: Don't delete ALB if another instance is starting

nguyen102:
Issue #, if available:

Description of changes:
Setup: 
There is one instance. Launch a new instance. After check-launch-dependency passes for that instance, terminate the original instance.

This PR prevents SWB from terminating the ALB, because it thinks that there are no more instance relying on the ALB left.

We fix this in `terminate-launch-depency`. Before terminating the ALB we check if there any `PENDING` instances that is associated with a ProjectType that has the input parameter `ACMSSLCertARN`. 

Checklist:

AS review ticket id:


main/solution/infrastructure/config/infra/cloudformation.yml
```diff
@@ -2,6 +2,12 @@ Conditions:
   IsDev: !Equals ['${self:custom.settings.envType}', 'dev']
```
SanketD92, line 2: This fix already exists on `develop` now. Any reason why we needed it in this PR?

nguyen102: I added it because when I was testing the `fix-rstudio-alb-pentest` branch, it's needed to create the right HostedZone for RStudio. So I think it's better to merge this fix into `fix-rstudio-alb-pentest` so the bugfix is in the feature branch as well as develop. 

It doesn't hurt to include it, and it makes development easier as it'll allow installation of SWB to proceed correctly.

SanketD92 (approved):
LGTM

PR #802, opened 2021-11-15
fix: run TRE tests for develop merge

SanketD92:
Issue #, if available:
N/A

Description of changes:
Run integration tests for TRE env when changes merged onto develop.

AS review ticket id:


nguyen102 (approved):
Note this will run the tests in the TRE environment, but it won't block merge to `mainline` if a test fails. For now, it's ok to run the tests so we atleast know, but we'll want to invest time in fixing this so it'll block the merge to `mainline`.

PR #823, opened 2021-11-23
chore: integration test cleanup for dummy envs

SanketD92:
Issue #, if available:
GALI-1249

Description of changes:
Service Workbench integration tests upon completion attempt to terminate provisioned workspaces. Some of these test workspaces are based on a temporary workspace config, which do not terminate fully (unsuccessful termination of these workspaces is expected, and does not impact test outcome).

However these are stored in the database with the status `TERMINATING_FAILED` message, and since they were linked to a temporary config, they show up as errored environment cards on the SWB UI. This could cause confusion, and potentially other tests to fail which depend on no workspaces being in the `TERMINATING_FAILED` state.

In the `afterAll` block of tests creating such workspaces, add logic to remove their entries completely from DDB.

Checklist:

AS review ticket id:

### DIFF
--- a/main/integration-tests/__test__/api-tests/common/workspace-service-catalogs/connections/send-ssh-public-key-connection.test.js
+++ b/main/integration-tests/__test__/api-tests/common/workspace-service-catalogs/connections/send-ssh-public-key-connection.test.js
@@ -13,6 +13,7 @@
  *  permissions and limitations under the License.
  */
 
+const _ = require('lodash');
 const { runSetup } = require('../../../../../support/setup');
 const errorCode = require('../../../../../support/utils/error-code');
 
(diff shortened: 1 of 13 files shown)


nguyen102 (dismissed):

main/integration-tests/__test__/api-tests/common/workspace-service-catalogs/connections/send-ssh-public-key-connection.test.js
```diff
@@ -37,6 +40,11 @@ describe('Send SSH public key scenarios', () => {
 
   afterAll(async () => {
     await deleteDefaultServiceCatalogProduct(setup, productInfo);
+    await Promise.all(
+      _.map(dummyWorkspacesToDelete, async envId => {
```
nguyen102, line 44: Since `dummyWorkspacesToDelete` is an array, I don't think you need to use lodash. You can just do 
```
dummyWorkspacesToDelete.map(async(workspace) => {
    return deleteWorkspaceServiceCatalog({ aws: setup.aws, id: envId });
});
```

SanketD92: True, but I try to stick with lodash as a pattern due to its [performance](https://measurethat.net/Benchmarks/Show/2676/0/arrayprototypemap-vs-lodashmap#latest_results_block)

nguyen102 (approved):

PR #825, opened 2021-11-24
fix: Change build-image CLI argument to files

liamcoatman:
Issue #, if available:

Description of changes:

Change argument name from `file` to `files`, which is what is read on lines 56 and 58. I've tested this manually and verified that only machine image `foo` is built when I run `pnpx sls build-image --files foo.json -s dev`. 

Checklist:

AS review ticket id:


SanketD92 (commented):
Could you provide more details as to what didn't work earlier with the  "file" argument?

liamcoatman:
> Could you provide more details as to what didn't work earlier with the "file" argument?

As far as I could tell it it does nothing. An argument like `--file foo.json` is ignored and the default (to build all Packer files in the directory) is used instead.

SanketD92:
> > Could you provide more details as to what didn't work earlier with the "file" argument?
> 
> As far as I could tell it it does nothing. An argument like `--file foo.json` is ignored and the default (to build all Packer files in the directory) is used instead.

Was able to verify the `--files` flag was honored while the `--flag` flag wasn't.

Please pull the latest changes from `develop` by clicking on the "Update branch" button below. Also, please add a commit (could even be an [empty](https://stackoverflow.com/questions/20138640/pushing-empty-commits-to-remote) one) that starts with a conventional commit type.

SanketD92 (approved):

PR #826, opened 2021-11-24
fix: Add wait time for Launch rstudio test 

nguyen102:
Issue #, if available:

Description of changes:
RStudio launch test checks that all RStudio instances are terminated before it runs the test. When running RStudio launch test in a suite, there can be instances from previous tests still terminating when the RStudio launch test starts.

This PR adds a ten minute wait time to the beginning of the RStudio test. It will poll DDB to check if all RStudio instances have been terminated. If those instances are not terminated, the test will wait a maximum of 10 minutes before failing.

This will fix false failure like seen [here](https://github.com/awslabs/service-workbench-on-aws/runs/4314246641?check_suite_focus=true)

AS review ticket id:


SanketD92 (approved):
LGTM

PR #829, opened 2021-12-08
fix: cidr port range check

SanketD92:
Issue #, if available:
The "fromPort" and “toPort” parameters in the API for updating a workspace's CIDR does not throw errors when ports outside range 0-65535 are entered. Although the update request does not fail the changes do not persist.

Description of changes:
Explicitly throw errors when ports outside 0-65535 range are used for CIDR updates.

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
@@ -54,6 +54,14 @@ class EnvironmentScCidrService extends Service {
     const erroneousInputs = [];
     const ipv6Format = [];
     const protPortCombos = {};
+    if (
+      _.some(
+        updateRequest,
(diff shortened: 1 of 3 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-cidr-service.js
```diff
@@ -54,6 +54,14 @@ class EnvironmentScCidrService extends Service {
     const erroneousInputs = [];
     const ipv6Format = [];
     const protPortCombos = {};
+    if (
+      _.some(
+        updateRequest,
+        rule => rule.fromPort < 0 || rule.fromPort > 65535 || rule.toPort < 0 || rule.toPort > 65535,
```
jn1119: Can these ranges be added to JSON schema, so that we don't have to explicitly check the range?

SanketD92: Done.

jn1119 (approved):

PR #832, opened 2021-12-14
fix: force securetransport traffic only for buckets with dynamic bucket policies

maghirardelli:
Issue #, if available: GALI-1239

Description of changes: AppSec requires that all bucket policies enforce TLS/HTTPS traffic only. Within SWB, all bucket policies are made during deployment. Some bucket policies then dynamically change due to user interaction with the SWB UI (deploying workspaces, creating studies, etc.). These policies cannot be edited in the cloudformation templates because they overwrite the dynamic changes that took place and create a breaking change. This fix introduces a new plugin that adds the necessary security statement to the buckets with these dynamic policies.

Checklist:

- [x] If new dependencies have been added, have they been pinned to specific versions? not to specific versions--following the workspace:* convention

AS review ticket id:


addons/addon-edit-s3-bucket-policy/packages/edit-s3-bucket-policy/lib/steps/edit-s3-bucket-policy-service.js
```diff
@@ -0,0 +1,119 @@
...
+      let oldStatement;
+
+      if (replacePattern === 'HTTPS') {
+        // Get TLS statement and remove from policy object
+        const tlsStatement = oldStatementSid;
+        statements.forEach((statement, index, object) => {
+          if (statement.Sid === tlsStatement) {
+            this.log.info(`Found statement ${tlsStatement}. Removing.`);
+            oldStatement = statement;
+            object.splice(index, 1);
```
jn1119: It isn't generally considered a good practice to remove items inside a list while you are iterating over them. Seems like `statements.map` function would be a better fit here where you can iterate over each statement and replace the old statement with the new one if it matches `tlsStatement`. That way you get a new list which can be used to replace the old one.

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -230,7 +242,7 @@ class EnvironmentMountService extends Service {
           s3Policy.Statement = s3Policy.Statement.filter(
             statement => ![listSid, getSid, putSid].includes(statement.Sid),
           );
-          [listStatement, getStatement, putStatement].forEach(statement => {
+          [listStatement, getStatement, putStatement, secureTransportStatement].forEach(statement => {
```
jn1119: Add a unit test for this new statement?

main/config/settings/.defaults.yml
```diff
@@ -201,4 +202,7 @@ egressNotificationTopicName: ${self:custom.settings.namespace}-EgressTopic
 # If you toggle the AppStream feature from true to false and redeploy the whole solution, the backend stack deployment shold error out with following message:
 # Error validating existing stack policy: Unknown logical id 'LogicalResourceId/AppStream*' in statement {} - stack policies can only be applied to logical ids referenced in the template
 # Stack policies are applied here: addons/addon-stack-policy/packages/stack-policy/lib/steps/update-cfn-stack-policy.js
-isAppStreamEnabled: false
\ No newline at end of file
+isAppStreamEnabled: false
+
+# The name of the S3 bucket used to store uploaded study data
+studyDataBucketName: ${self:custom.settings.globalNamespace}-studydata
```
jn1119, line 212: Do we need `studyDataBucketName ` in other `.defaults.yml` files if we have it added here. I wondered that because `deploymentBucketName` is just initialized here once but referenced in cloudformation.yml of post-deployment and backend both

addons/addon-edit-s3-bucket-policy/packages/edit-s3-bucket-policy/lib/steps/edit-s3-bucket-policy-service.js
maghirardelli: Got it! Will try this alternative out.

main/config/settings/.defaults.yml
maghirardelli: Doesn't the fact that it is in the main `.defaults.yml` mean that all SDCs will have this value _unless_ it is overwritten in another SDC settings file? This is the only place that it is a parameter so this value will be used by all SDCs I thought.

addons/addon-edit-s3-bucket-policy/packages/edit-s3-bucket-policy/lib/steps/edit-s3-bucket-policy-service.js
maghirardelli: Edited to use `.map`

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
maghirardelli: created a helper method to generate these statements and added a unit test for that method per Jeet's advice

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
```diff
@@ -170,58 +170,14 @@ class EnvironmentMountService extends Service {
         // Get statements for listing and reading study data, respectively
         const statements = s3Policy.Statement;
         s3Prefixes.forEach(prefix => {
+          // eslint-disable-next-line prefer-const
+          let allStatements = this.getAllStatements(s3BucketName, prefix);
+          // eslint-disable-next-line no-unused-vars
```
jn1119: You probably don't need to suppress this warning

jn1119 (dismissed):

jn1119 (approved):

addons/addon-base-raas/packages/base-raas-services/lib/environment/environment-mount-service.js
maghirardelli: changed it to const and got rid of the no-unused-vars warning

PR #839, opened 2021-12-20
test: Terminate stopped rstudio instances

nguyen102:
Issue #, if available:

Description of changes:
* In E2E tests if we try to terminate too many workspaces as a time, we sometimes receive an error from the backend, and not all workspaces are terminated. We're adding a 3 second wait time between terminating workspaces

* In integration tests, if we find an RStudio instance that is stopped, we terminate it before starting the launch-rstudio test.

**Test:** Tested code against E2E test environment and code worked as expected.

Checklist:

AS review ticket id:


SanketD92 (approved):
LGTM

PR #843, opened 2021-12-20
fix: elb logging on

maghirardelli:
Issue #, if available: GALI-1330

Description of changes: turn on ELB logging for each ALB provisioned. **This change does require users to update permissions of member accounts for it to work. It is NOT a breaking change.** Additionally, only _new_ ALBs will have logging on. So, for users to see the effect of this change, they need to update the permissions of their member accounts in the SWB Console and terminate all RStudio workspaces (which will delete all existing ALBs). If users decide not to do this, their workspaces (RStudio and others) will continue to work, but they will not have logging on for new RStudio workspaces on existing ALBs.

Checklist:

AS review ticket id:


main/solution/infrastructure/config/infra/cloudformation.yml
```diff
@@ -11,6 +11,59 @@ Conditions:
 
 Description: (SO0144) Service Workbench on AWS Solution
 
+Mappings:
+  ELBAccountIDMapping:
```
SanketD92, line 15: Which accounts are these?

nguyen102: I think it's from this: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-logging-bucket-permissions

nguyen102 (approved):

main/solution/infrastructure/config/infra/cloudformation.yml
maghirardelli: Yes to Tim's link!

PR #844, opened 2021-12-21
fix: Allow onboarding member account in non AppStream supported regions

nguyen102:
Issue #, if available:

Description of changes:
If you wanted to deploy SWB with AppStream disabled in a non AppStream supported region, you were unable to do so. You will encounter this error in CloudFormation

`Template format error: Unrecognized resource types: [AWS::AppStream::StackFleetAssociation, AWS::AppStream::Stack, AWS::AppStream::Fleet]`

This code change allows SWB to be deployed in all regions by removing those AppStream specific resources. The resources will be only removed if AppStream is disabled. 

**Testing:** SWB was able to onboard member accounts in both AppStream supported and non supported regions. The upgrade process from an old `onboard-account` template to the current template was also tested. The upgrade process was smooth. There was no conflicts when using the SWB UI to upgrade the member account.

Checklist:

AS review ticket id:


addons/addon-base-raas/packages/base-raas-services/lib/cfn-templates/cfn-template-service.js
```diff
@@ -48,6 +54,19 @@ class CfnTemplateService extends Service {
 
   async getTemplate(name) {
     const entry = _.find(this.store, ['key', name]);
+    const isAppStreamEnabled = await this.settings.get(settingKeys.isAppStreamEnabled);
```
SanketD92: Could we use `settings.getBoolean` instead?

nguyen102: updated to use getBoolean

addons/addon-base-raas/packages/base-raas-services/lib/cfn-templates/cfn-template-service.js
```diff
@@ -48,6 +54,19 @@ class CfnTemplateService extends Service {
...
   async getTemplate(name) {
     const entry = _.find(this.store, ['key', name]);
+    const isAppStreamEnabled = await this.settings.getBoolean(settingKeys.isAppStreamEnabled);
+    if (!isAppStreamEnabled && name === 'onboard-account' && entry) {
+      const doc = yamlParse(entry.value);
+      delete doc.Resources.AppStreamFleet;
+      delete doc.Resources.AppStreamStack;
+      delete doc.Resources.AppStreamStackFleetAssociation;
+      delete doc.Outputs.AppStreamFleet;
+      delete doc.Outputs.AppStreamStack;
```
jn1119: Rather than deleting specific Resources was it possible to remove the ones where we have condition as `isAppStream`? That way any future changes to that template (if any) would be accounted for

nguyen102: updated to remove all resources/outputs with `isAppStream` or `isAppStreamAndCustomDomain` condition.

jn1119 (approved):

PR #854, opened 2022-01-07
chore: Remove internal auth logic

nguyen102:
Issue #, if available:

Description of changes:
chore: Remove internal auth logic

**Testing:** Deployed code to AWS. Tested that Cognito user can log in, launch/terminate workspaces, create studies, edit users. Tested that internal auth users can not log in. 

```
1. User logins attempted using internal authentication provider have to be denied (API: `/api/authentication/id-tokens`)
* Done. Internal Auth `tokenIssuerLocator` references `internalAuthenticationProviderService`, which has been removed
2. Ensure new internal auth users cannot be created.
* Done, `users-controller.js` API for creating new user is removed
3. Authentication code paths checking for internal authentication (code sprinkled across repo where auth provider is being checked for `internal`)
* Done, internal auth logic removed. `internalAuthenticationProviderService` also removed
4. Updates to unit tests that check for internal authentication logic.
* Done
5. Sanity check to ensure existing internal users cannot get access to any SWB resources once these changes are made (list APIs tested in PR).
* Done, Auth logic calls `internalAuthenticationProviderService/validateToken` which has been removed. The validation fails and therefore the user is not authenticated or authorized to proceed.
6. The `_system_` user context is not an actual user, although its namespace (`ns` value) should not be `internal` (change it to the Cognito native pool equivalent value). Code found here: `addons/addon-base/packages/services/lib/helpers/system-context.js`
* Done , changed it to `cognito`. There is currently no Cognito equivalent, so we get to choose the value we would like to use. In this case I chose `cognito`.
```
Checklist:

AS review ticket id:


addons/addon-base/packages/services/lib/helpers/system-context.js
```diff
@@ -31,7 +29,7 @@ function getSystemRequestContext() {
 
   const systemUid = '_system_';
   const systemUsername = '_system_';
-  const systemUserNamespace = internalAuthProviderId;
+  const systemUserNamespace = 'cognito';
```
nguyen102: I think we can choose our own preset value for this. An example `ns` value for an Auth0 user is `Auth0||||https://cognito-idp.eu-west-3.amazonaws.com/eu-west-3_U204NRMAr",`.

To keep things simple for a system user, I'm thinking we can just use `cognito`. Let me know if you prefer another default name for system users.

SanketD92: Just a thought - If this value doesn't need to be anything specific out of the actual authentication providers, then maybe we can just set it to `system` or `lambda` (since we won't be able to find this user in cognito anyways).

addons/addon-base-rest-api/packages/base-controllers/lib/users-controller.js
```diff
@@ -91,30 +64,16 @@ async function configure(context) {
     }),
   );
 
+  // TODO: Do we want to support changing user passwords? Currently this API is only supported for internal Auth
```
SanketD92, line 67: We can keep this API commented for now. 

For updating user passwords from the native user pool, we'll need to implement [admin-set-user-password](https://docs.aws.amazon.com/cli/latest/reference/cognito-idp/admin-set-user-password.html) later, if needed.

SanketD92 (dismissed):

addons/addon-base/packages/services/lib/helpers/system-context.js
nguyen102: Noted: renaming to `_system_` to be consistent with `systemUid` and `systemUserNamespace`

SanketD92 (approved):
LGTM

PR #857, opened 2022-01-11
feat: My study migration

maghirardelli:
Issue #, if available: GALI-1352

Description of changes: added new migration service and controller that has two APIs: 
1. **PUT /migrate/mystudies** - accepts a request body of studyId and uid mappings to migrate those StudyIds to be owned by those uids. Uids must be for non internal users. StudyIds must not already be owned by uids. 
2. **GET /migrate/mystudies** - lists all studies of the category My Studies. Both APIs can only be called by admin users. 
 
Unit tests have been written for all new/modified methods. Integ tests will come later after Sanket's native Cognito PR.

Checklist:

AS review ticket id:


addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
```diff
@@ -0,0 +1,89 @@
...
+    const studyOperationService = await this.service('studyOperationService');
+    const userService = await this.service('userService');
+    const studyService = await this.service('studyService');
+
+    const migrationResults = [];
+    await Promise.all(
+      migrationMappings.map(async item => {
+        const endUser = item.uid;
+        const studyId = item.studyId;
+        // console.log(result1);
```
SanketD92: redundant comment

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/migration-controller.js
```diff
@@ -0,0 +1,50 @@
...
+
+  // ===============================================================
+  //  GET /mystudies (mounted to /api/migrate)
+  // ===============================================================
+  router.get(
+    '/mystudies',
+    wrap(async (req, res) => {
+      const migrationService = await context.service('migrationService');
+      const requestContext = res.locals.requestContext;
+      const result = await migrationService.listMyStudies(requestContext);
```
SanketD92, line 42: Does the response make it evident which studies require migration, and which internal user the resource belonged to? These values in the response body will help a lot for admins to figure out what to migrate:

```
studyId: <sampleStudyId>,
uid: <sampleUidValue>,
username: <sampleUsername>,
authProvider: <respectiveAuthProviderId / 'internal'> 

```

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
```diff
@@ -0,0 +1,89 @@
...
+    this._query = () => dbService.helper.query().table(studiesTable);
+    this._deleter = () => dbService.helper.deleter().table(studiesTable);
+    this._scanner = () => dbService.helper.scanner().table(studiesTable);
+
+    this.categoryIndex = this.settings.get(settingKeys.categoryIndexName);
+  }
+
+  async migrateMyStudiesPermissions(requestContext, migrationMappings) {
+    if (!isAdmin(requestContext)) {
+      throw this.boom.forbidden("You don't have permission to migrate My Studies", true);
```
nguyen102: Let's be more specific and say "You need admin permissions to migrate My Studies"

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
```diff
@@ -0,0 +1,89 @@
...
+    const studyService = await this.service('studyService');
+
+    const migrationResults = [];
+    await Promise.all(
+      migrationMappings.map(async item => {
+        const endUser = item.uid;
+        const studyId = item.studyId;
+        // console.log(result1);
+        if (await userService.isInternalAuthUser(endUser)) {
+          throw this.boom.forbidden('You cannot migrate My Studies to internal user', true);
```
nguyen102: These errors should be a badRequest, not a permission error.
`this.boom.badRequest`

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        // check that end user is not already owner
+        if (await studyService.isStudyAdmin(requestContext, studyId, endUser)) {
+          throw this.boom.forbidden('You cannot migrate My Studies to current owner', true);
```
nguyen102: These errors should be a badRequest, not a permission error.
this.boom.badRequest.

`Current study is already owned by ${endUser}`

addons/addon-base-raas/packages/base-raas-services/lib/migration/__tests__/migration-service.test.js
```diff
@@ -0,0 +1,163 @@
...
+      // BUILD
+      const requestContext = createAdminContext();
+      const migrationMapping = [
+        { studyId: 'testStudy1', uid: 'testUser1' },
+        { studyId: 'testStudy2', uid: 'testUser2' },
+      ];
+
+      // OPERATE
+      const results = await service.migrateMyStudiesPermissions(requestContext, migrationMapping);
+      //   console.log(results);
```
nguyen102: delete

addons/addon-base-raas/packages/base-raas-services/lib/migration/__tests__/migration-service.test.js
```diff
@@ -0,0 +1,163 @@
...
+    });
+
+    it('should pass correct updateRequest', async () => {
+      // BUILD
+      const requestContext = createAdminContext();
+      const migrationMapping = [{ studyId: 'testStudy1', uid: 'testUser1' }];
+
+      // OPERATE
+      await service.migrateMyStudiesPermissions(requestContext, migrationMapping);
+      //   console.log(results);
```
nguyen102: delete

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        const result = await studyOperationService.updatePermissions(requestContext, studyId, {
+          usersToAdd: [{ uid: endUser, permissionLevel: 'admin' }],
+          usersToRemove: [{ uid: '*', permissionLevel: 'admin' }],
```
nguyen102, line 68: I'm not familiar with study permissions. What does a `uuid` of `*` mean? In what cases does a study have an owner of `*`?

maghirardelli: This is a new convention. The `*` means to remove all users like a wildcard. In the case of the My Studies, this will only be one user. But we don't necessarily know the current user so we need to put something. The logic in the methods file later adds whoever was the old admin of the My Study to be the value in place of `*`. 

TLDR: `*` is never the owner--just a wildcard placeholder.

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
maghirardelli: Good idea! Adopting this wording.

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
maghirardelli: Got it! Thank you!

nguyen102 (dismissed):

openapi.yaml
```diff
@@ -2147,6 +2147,34 @@ paths:
       responses:
         '200':
           description: 'Return success update message'
+  /api/migrate/mystudies:
```
rsmayda: nit: should apis be in kebab-case (my-studies)?

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
```diff
@@ -0,0 +1,113 @@
...
+        const endUser = item.uid;
+        const studyId = item.studyId;
+        if (await userService.isInternalAuthUser(endUser)) {
+          throw this.boom.badRequest('You cannot migrate My Studies to internal user', true);
+        }
+
+        // check that end user is not already owner
+        if (await studyService.isStudyAdmin(requestContext, studyId, endUser)) {
+          throw this.boom.badRequest(`Current study is already owned by ${endUser}`, true);
+        }
```
rsmayda: should this be a failure? effectively this request is to change the study ownership to `endUser` but this check is saying that `endUser` is already the owner. Please correct me if wrong!

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
```diff
@@ -0,0 +1,113 @@
...
+      throw this.boom.forbidden('You need admin permissions to list all My Studies in this environment', true);
+    }
+
+    const studyService = await this.service('studyService');
+    const userService = await this.service('userService');
+
+    const result = await this._query()
+      .index(this.categoryIndex)
+      .key('category', 'My Studies')
+      .limit(1000)
```
rsmayda, line 87: 1. Will this API be hit from the UI? 
2. How likely it is to hit the limit?

addons/addon-base-raas/packages/base-raas-services/lib/migration/__tests__/migration-service.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+      // CHECK
+      expect(studyOperationService.updatePermissions).toHaveBeenCalledWith(requestContext, 'testStudy1', {
+        usersToAdd: [{ uid: 'testUser1', permissionLevel: 'admin' }],
+        usersToRemove: [{ uid: '*', permissionLevel: 'admin' }],
+      });
+    });
+  });
+
+  describe('listMyStudies', () => {
```
rsmayda, line 156: Not testing happy path at all

addons/addon-base-raas/packages/base-raas-services/lib/study/helpers/entities/study-permissions-methods.js
```diff
@@ -52,15 +52,23 @@ function isAdmin(studyPermissionsEntity, uid) {
 function applyUpdateRequest(studyPermissionsEntity, updateRequest) {
   const entity = studyPermissionsEntity;
 
+  if (updateRequest.usersToRemove[0].uid === '*') {
+    // For mitigation of old internal users My Studies
+    const level = `${updateRequest.usersToRemove[0].permissionLevel}Users`;
+    const oldUser = entity[level][0];
+    updateRequest.usersToRemove = [{ uid: oldUser, permissionLevel: 'admin' }];
```
rsmayda, line 59: What if there are multiple "old" admins?

addons/addon-base-raas/packages/base-raas-services/lib/study/helpers/entities/__tests__/study-permissions-methods.test.js
```diff
@@ -55,5 +55,54 @@ describe('study permissions methods', () => {
         writeonlyUsers: ['u-4'],
       });
     });
+
+    it('should apply requested changes for migration', () => {
```
rsmayda, line 59: could you add test for multiple adminUsers

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
```diff
@@ -388,7 +388,8 @@ class StudyService extends Service {
       throw this.boom.forbidden('Permissions cannot be set for studies in the "Open Data" category', true);
     }
 
-    if (isMyStudies(studyEntity)) {
+    // If study entity is My Study and not part of a migration request, throw error
+    if (isMyStudies(studyEntity) && updateRequest.usersToRemove[0].uid !== '*') {
```
rsmayda: when updating permission we can now use `*`  for any update permissions call; is that the intention?

Also we are saying MyStudies MUST have  `*` in "toRemove" is that the intention?

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
```diff
@@ -1294,4 +1294,103 @@ describe('studyService', () => {
       expect(retVal).toEqual(false);
     });
   });
+
+  describe('updatePermissions', () => {
+    it('should throw error if My Study but not migration request', async () => {
```
rsmayda, line 1299: that isn't so clear it is probably clearer to say: 
```suggestion
    it('should throw error if My Study and usersToRemove is not *', async () => {
```

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
```diff
@@ -1294,4 +1294,103 @@ describe('studyService', () => {
       expect(retVal).toEqual(false);
     });
   });
+
+  describe('updatePermissions', () => {
```
rsmayda, line 1298: Could you add another use-case for using the `*` outside of the MyStudy context

openapi.yaml
maghirardelli: Will change.

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
maghirardelli: The migration is failing was my logic. The request is failing because what was requested was already up to date. Would you prefer a different boom error?

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
maghirardelli: 1. The API will not be hit from the UI.
2. This is the same limit on the method called from GET /studies (`list()` in `study-service.js`). If we are accepting this limit there, I thought it would be okay to accept the limit here.

addons/addon-base-raas/packages/base-raas-services/lib/study/helpers/entities/study-permissions-methods.js
maghirardelli: My Studies by design only have one admin. A user cannot share them and before now, could not transition ownership. Therefore, the AdminUsers list will only have one user ever and none of the other permissions lists will have any users as well.

addons/addon-base-raas/packages/base-raas-services/lib/study/helpers/entities/__tests__/study-permissions-methods.test.js
maghirardelli: See above comment, but there will never be multiple adminUsers as My Studies by design are only owned and can only be accessed by one user (formerly the one who created it; now, the one who created it or whoever it is migrated to).

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
maghirardelli: By design, we have never before allowed changing permissions on My Study because they are not sharable. So, yes, MyStudies MUST have `*` in "toRemove" because that is what the API controller passes in. It is the only API that does this.
Theoretically, a user could put `*` in `userToRemove` in an update permissions API call body which might cause unexpected behavior. We can either assume that won't happen because `*` has never been a valid input before or we can build some sort of check to make sure that the only API that can pass `*` to update permissions in `userToRemove` is the my studies migration API. What is acceptable/required in your opinion?

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
maghirardelli: I am going to combine both messages, as they *should* be synonymous with each other because the only time `usersToRemove` should be `*` is when it is a migration request.

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
maghirardelli: There should not be a use-case for using `*` outside of the MyStudy context.

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
rsmayda: I would put in logic to prevent the use of `*` unless we explicitly want that functionality

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
rsmayda: There is nothing from the code preventing this use-case from happening so it is a side-effect we need to test

addons/addon-base-raas/packages/base-raas-services/lib/migration/migration-service.js
rsmayda: since the request is both `endUser` and `studyId` for a user to action this error they would need both, right?

maghirardelli: Good catch. I'll include both.

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
maghirardelli: Got it! Will do.

addons/addon-base-raas/packages/base-raas-services/lib/migration/__tests__/migration-service.test.js
maghirardelli: Added this test

addons/addon-base-raas/packages/base-raas-services/lib/study/__tests__/study-service.test.js
maghirardelli: Removed the part that relies on `*` as migration indicator and disallowed `*` to be part of the update request so there is no need to test anymore

addons/addon-base-raas/packages/base-raas-services/lib/study/helpers/entities/study-permissions-methods.js
```diff
@@ -71,6 +71,13 @@ function applyUpdateRequest(studyPermissionsEntity, updateRequest) {
   });
 }
 
+function isValidUpdateRequest(updateRequest) {
+  // Check to make sure the wildcard is not used as a uid in usersToRemove
+  const wildcardEntries = updateRequest.usersToRemove.filter(entry => entry.uid === '*');
```
rsmayda: I think using `any` would make this simpler

```suggestion
  return updateRequest.usersToRemove.any(entry => entry.uid === '*');
```

addons/addon-base-raas/packages/base-raas-services/lib/study/study-permission-service.js
```diff
@@ -337,6 +338,14 @@ class StudyPermissionService extends Service {
         { studyEntity, studyPermissionsEntity },
       );
 
+      // Ensure only migration requests can have wildcards in the updateRequest
+      if (!('isMigration' in requestContext) && !isValidUpdateRequest(updateRequest)) {
```
rsmayda: I think this is a bit clearer
```suggestion
      if (!requestContext.isMigration && !isValidUpdateRequest(updateRequest)) {
```

addons/addon-base-raas/packages/base-raas-services/lib/study/study-service.js
```diff
@@ -389,7 +389,7 @@ class StudyService extends Service {
     }
 
     // If study entity is My Study and not part of a migration request, throw error
-    if (isMyStudies(studyEntity) && updateRequest.usersToRemove[0].uid !== '*') {
+    if (isMyStudies(studyEntity) && !('isMigration' in requestContext)) {
```
rsmayda:
```suggestion
    if (isMyStudies(studyEntity) && !requestContext.isMigration) {
```

addons/addon-base-raas/packages/base-raas-services/lib/study/helpers/entities/study-permissions-methods.js
maghirardelli: I think Javascript uses `some` for this purpose. It did make it simpler! Thanks for pointing out there may be a more efficient array method.

rsmayda (approved):

PR #867, opened 2022-01-18
chore: remove add local user from UI 

nguyen102:
Issue #, if available:

Description of changes:
* Remove update internal auth user api
* Remove `Add Local User` UI component
* Fix `AddSingleUser.test.js` test. This test was calling the wrong UI component

There's two versions of this code, the [first one](https://github.com/awslabs/service-workbench-on-aws/blob/feat-internal-auth-dep/addons/addon-base-rest-api/packages/base-controllers/lib/users-controller.js#L67) was removed, but we need to remove the second one as well. 

Checklist:

AS review ticket id:


addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/users-controller.js
```diff
@@ -74,22 +74,23 @@ async function configure(context) {
     }),
   );
 
+  // TODO: Do we want to support changing user passwords? Currently this API is only supported for internal Auth
```
SanketD92: I think there's no need to support password change, if login API is gone anyways.

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/users-controller.js
```diff
@@ -74,22 +74,23 @@ async function configure(context) {
...
-      const requestContext = res.locals.requestContext;
-      const username = req.params.username;
-      const { password } = req.body;
-
-      // Save password salted hash for the user in internal auth provider (i.e., in passwords table)
-      await dbPasswordService.savePassword(requestContext, { username, password });
-      res.status(200).json({ username, message: `Password successfully updated for user ${username}` });
-    }),
-  );
+  // router.put(
```
SanketD92: Could we also update API changes in the openapi file?

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/users-controller.js
nguyen102: deleted

addons/addon-base-raas/packages/base-raas-rest-api/lib/controllers/users-controller.js
nguyen102: good idea. Updated OpenApi file.

SanketD92 (approved):
LGTM

PR #889, opened 2022-01-26
test: Add Cognito creds for Integration and E2E tests

nguyen102:
Issue #, if available:

Description of changes:
* Updated e2e tests creds that is stored in Github config files
* We still need to update S3 config file for Int Tests and Github Secrets. However, that cannot be done until we merge the feature branch to `develop`. If we update those configs now, the current tests will not run correctly. 

AS review ticket id:


SanketD92 (approved):
LGTM

PR #898, opened 2022-02-03
fix: Reduce scope of list users API for non admin users

nguyen102:
Issue #, if available:

Description of changes:
Reduce the number of fields that is returned by the list users API when the requester is a `researcher`. We will only return these fields `['firstName', 'lastName', 'email', 'uid', 'username']`.

**Testing:**  
* Checked that researchers can still edit the user permissions in `Org studies`
* Checked the researchers can still navigate to `Workspace` page and launch workspaces

Checklist:

AS review ticket id:


SanketD92 (approved):
LGTM

PR #900, opened 2022-02-05
chore: Migrate api integ tests

maghirardelli:
Issue #, if available: GALI-1375

Description of changes: add integration tests to test api/migrate/my-studies GET and PUT. Test cases:
Call list my studies (GET)
    Should fail if request not by admin
    Should return message that there is nothing to migrate if admin
Migrate my study to non internal user (PUT)
    Should fail if request not by admin
    Should succeed if end user is not internal and should fail if attempting to transfer to same non internal user

Checklist:

AS review ticket id:

### DIFF
--- a/main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
+++ b/main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
@@ -65,7 +65,7 @@ describe('Migration for My Studies scenarios', () => {
       await researcherSession.resources.studies.create({ id: studyId, name: studyId, category: 'My Studies' });
       await researcherSession.resources.studies.study(studyId);
 
-      const body = [{ studyId: studyId, uid: defaultUser.uid }];
+      const body = [{ studyId, uid: defaultUser.uid }];
       await expect(researcherSession.resources.migration.migrateMyStudyOwnership(body)).rejects.toMatchObject({
         code: errorCode.http.code.forbidden,
       });
@@ -77,7 +77,7 @@ describe('Migration for My Studies scenarios', () => {
       await adminSession.resources.studies.create({ id: studyId, name: studyId, category: 'My Studies' });
       await adminSession.resources.studies.study(studyId);
 
-      const body = [{ studyId: studyId, uid: defaultUser.uid }];
+      const body = [{ studyId, uid: defaultUser.uid }];
       const expectedContents = {
         category: 'My Studies',
         id: studyId,
--- a/main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
+++ b/main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
@@ -19,21 +19,15 @@ const errorCode = require('../../../../support/utils/error-code');
 describe('Migration for My Studies scenarios', () => {
   let setup;
   let adminSession;
-  let defaultUser;
-  let defaultUserSession;
+  let targetUser;
+  let targetUserSession;
 
   beforeAll(async () => {
     setup = await runSetup();
     adminSession = await setup.defaultAdminSession();
     // Create cognito user to migrate ownership to
-    // Proper way that currently doesn't work:
-    // const username = setup.gen.username();
-    // defaultUser = adminSession.resources.users.defaults({ username });
-    // defaultUser = await await adminSession.resources.users.create(defaultUser);
-    // await adminSession.resources.users.user(defaultUser.uid);
-    // Workaround:
-    defaultUserSession = await setup.createResearcherSession();
-    defaultUser = defaultUserSession.user;
+    targetUserSession = await setup.createResearcherSession();
+    targetUser = targetUserSession.user;
   });
 
   afterAll(async () => {
@@ -48,10 +42,8 @@ describe('Migration for My Studies scenarios', () => {
       });
     });
 
-    it('should return message that there is nothing to migrate', async () => {
-      await expect(adminSession.resources.migration.listInternaAuthUserMyStudies()).resolves.toStrictEqual(
-        'No My Studies are owned my internal users. No migration needed.',
-      );
+    it('should return empty list if nothing to migrate', async () => {
+      await expect(adminSession.resources.migration.listInternaAuthUserMyStudies()).resolves.toStrictEqual([]);
     });
   });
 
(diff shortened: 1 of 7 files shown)


main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+  describe('Migrating My Studies', () => {
+    let studyId;
+
+    it('should fail if request not by admin', async () => {
+      const researcherSession = await setup.createResearcherSession();
+      // Create My Study
+      studyId = setup.gen.string({ prefix: `create-my-study-test` });
+      await researcherSession.resources.studies.create({ id: studyId, name: studyId, category: 'My Studies' });
+      await researcherSession.resources.studies.study(studyId);
```
nguyen102: what does this line do? It gets the study, but doesn't store the value of the study?

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+const { runSetup } = require('../../../../support/setup');
+const errorCode = require('../../../../support/utils/error-code');
+
+describe('Migration for My Studies scenarios', () => {
+  let setup;
+  let adminSession;
+  let defaultUser;
+  let defaultUserSession;
```
SanketD92: Suggested renames: `targetUser` and `targetUserSession`
Reason: `default` resources are those that are already created in SWB and their IDs are provided via config file.

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
```diff
@@ -0,0 +1,95 @@
...
+    setup = await runSetup();
+    adminSession = await setup.defaultAdminSession();
+    // Create cognito user to migrate ownership to
+    // Proper way that currently doesn't work:
+    // const username = setup.gen.username();
+    // defaultUser = adminSession.resources.users.defaults({ username });
+    // defaultUser = await await adminSession.resources.users.create(defaultUser);
+    // await adminSession.resources.users.user(defaultUser.uid);
+    // Workaround:
+    defaultUserSession = await setup.createResearcherSession();
```
SanketD92: This is the right way to create users, follows the pattern we use in almost all of our integration test suites (therefore, wouldn't call this a workaround either).

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
```diff
@@ -0,0 +1,95 @@
...
+  let adminSession;
+  let defaultUser;
+  let defaultUserSession;
+
+  beforeAll(async () => {
+    setup = await runSetup();
+    adminSession = await setup.defaultAdminSession();
+    // Create cognito user to migrate ownership to
+    // Proper way that currently doesn't work:
+    // const username = setup.gen.username();
```
SanketD92: remove comments

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
```diff
@@ -0,0 +1,95 @@
...
+    it('should fail if request not by admin', async () => {
+      const researcherSession = await setup.createResearcherSession();
+      await expect(researcherSession.resources.migration.listInternaAuthUserMyStudies()).rejects.toMatchObject({
+        code: errorCode.http.code.forbidden,
+      });
+    });
+
+    it('should return message that there is nothing to migrate', async () => {
+      await expect(adminSession.resources.migration.listInternaAuthUserMyStudies()).resolves.toStrictEqual(
+        'No My Studies are owned my internal users. No migration needed.',
```
SanketD92: I probably missed this in the implementation PR, but list methods should ideally return back a list. In this case it should be an empty list.

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
```diff
@@ -0,0 +1,95 @@
...
+    });
+  });
+
+  describe('Migrating My Studies', () => {
+    let studyId;
+
+    it('should fail if request not by admin', async () => {
+      const researcherSession = await setup.createResearcherSession();
+      // Create My Study
+      studyId = setup.gen.string({ prefix: `create-my-study-test` });
```
SanketD92: The prefix should attempt to describe the test case, in this case it should be `migrate-my-study-test`.

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+      const body = [{ studyId, uid: defaultUser.uid }];
+      await expect(researcherSession.resources.migration.migrateMyStudyOwnership(body)).rejects.toMatchObject({
+        code: errorCode.http.code.forbidden,
+      });
+    });
+
+    it('should succeed if end user is non internal and fail on reattempt after success', async () => {
+      // Create My Study
+      studyId = setup.gen.string({ prefix: `create-my-study-test` });
```
SanketD92: same comment for prefix

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
maghirardelli: This was following the convention I saw in other integration tests of using `.create` and then `.study`. I believe `create` calls the API and creates the resource and `study` creates a Study object within the resources object that can be referenced later. But I didn't end up referencing them later so I can take these lines out. The tests still work with the lines taken out. Thanks for pointing it out!

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
maghirardelli: Will change.

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
maghirardelli: Got it. I thought I was going to switch to creating the user after the user create method was fixed. So I left this here for when that was in place. I will stick to just using the session logic.

main/integration-tests/__test__/api-tests/common/migration/my-study-migration.test.js
maghirardelli: I thought we discussed that this message was helpful to the admin users who had to migrate these resources. Should the API be returning an empty list with no helpful message like this?

SanketD92: I don't think a message is required, the right place for this would rather be in this API's description field in the openapi file. Also we're explaining how to use this API in our release [documentation](https://github.com/awslabs/service-workbench-on-aws/pull/892/files#diff-3e9263816b528a0d25a0bf4c15d2d8721460ce24df019793b240a4c146f88186) as well.

The regular list studies API (not migrate list) can always be used to get a glimpse of their current state of study resources. Moreover, if the user has reached a point where they get back an empty list with this migrate list API - then they've already used this correctly, or they don't need to use it at all.

maghirardelli: The documentation mentions this message. I can edit that and include that change in this PR as well. Changing to empty list everywhere!

SanketD92 (approved):
LGTM

PR #903, opened 2022-02-07
fix: Handle workflow-trigger-service StepFunction execution failure

nguyen102:
Issue #, if available:
Handle exceptions in `api/workflows/:workflowId/v/:versionId/trigger` API.

Description of changes:
When calling `StepFunction.startExecution` for the trigger workflow API catch any exception that might occur. 

This is the only part of the code in [triggerWorkflow](https://github.com/awslabs/service-workbench-on-aws/blob/cab312c4bea930e58dd3aee0038ad681cae16a8a/addons/addon-base-workflow/packages/base-workflow-core/lib/workflow/workflow-trigger-service.js#L39) that needs to be protected from unhandled exception. User inputs for the API is validated by `jsonSchemaValidationService`. If the user input does not pass validation a 400 error is returned to the user. After the user inputs has been validated, there are no malformed requests that could cause an unhandle exception. The only unhandle exception is if `StepFunction.startExecution` fails. 

Exception that could occur before:
500 HTTP Status code
```
{
    "requestId": "",
    "code": "ExecutionLimitExceeded",
    "message": "Something went wrong"
}
```

Exception is now masked
500 HTTP Status code
```
{
    "requestId": "",
    "code": "internalError",
    "message": "Something went wrong"
}
```

The error is masked in the API response, but additional details for debugging will be logged in CWL.

Checklist:

AS review ticket id:


addons/addon-base-workflow/packages/base-workflow-core/lib/workflow/workflow-trigger-service.js
```diff
@@ -84,7 +84,17 @@ async function triggerStepFunctions({ instance, meta, input }) {
...
-  const data = await sf.startExecution(params).promise();
+  let data = {};
+  try {
+    data = await sf.startExecution(params).promise();
+  } catch (e) {
+    throw this.boom.internalError(
+      `Step Function could not start execution for State Machine ${stateMachineArn} with params ${JSON.stringify(
+        params,
+      )}. Error code: ${e.code}`,
+      false,
```
SanketD92, line 95: optional: Why not also display the error message on SWB UI?

```suggestion
      true,
```

SanketD92 (dismissed):
Left a comment, LGTM otherwise.

addons/addon-base-workflow/packages/base-workflow-core/lib/workflow/workflow-trigger-service.js
nguyen102: The security ticket ask us to not reveal error codes in the response. Setting this value to false doesn't show the error code in the API response, but still logs it in Cloudwatch Logs. Logging the error in CWL will help the developer debug the issue.

SanketD92 (approved):

PR #918, opened 2022-02-11
fix: pre-deployment checks work with fresh deployment

maghirardelli:
Issue #, if available:

Description of changes: Added check to only run the internal auth upgrade readiness check if the install was not a fresh deployment

Checklist:

AS review ticket id:


addons/addon-internal-auth-dep-check/packages/internal-auth-dep-check/lib/steps/internal-auth-dep-check-service.js
```diff
@@ -42,6 +42,20 @@ class InternalAuthDepCheckService extends Service {
   async execute() {
     // Keep track of resources blocking upgrade
     const blockers = {};
+    const usersTable = this.settings.get(settingKeys.usersTableName);
+
+    // check not fresh install
+    try {
```
SanketD92, line 48: Could you also add a unit test for this case?

SanketD92, line 48: Could you also add a unit test for this case?

addons/addon-internal-auth-dep-check/packages/internal-auth-dep-check/lib/steps/internal-auth-dep-check-service.js
```diff
@@ -42,6 +42,20 @@ class InternalAuthDepCheckService extends Service {
...
+      await this.dbService.dynamoDb.describeTable({ TableName: usersTable }).promise();
+    } catch (e) {
+      if (e.code === 'ResourceNotFoundException') {
+        this.log.info('This is first time deployment, no resources exist.');
+        return [];
```
SanketD92: Why return empty list? (why not simply `return;` ?)

maghirardelli: following the convention in the other pre-dep check! I can simply `return;` though!

nguyen102 (approved):

SanketD92 (approved):

PR #942, opened 2022-03-07
fix: Throw HTTP Status 429 error when there are too many get Sagemaker Presigned URL requests

nguyen102:
Issue #, if available:

Description of changes:
When multiple requests are made to retrieve a Sagemaker URL, we can not process more than one request at a time.

We'll now throw a HTTP Status Code 429 error, with the following body
```
429 error
{
    "requestId": "",
    "code": "tooManyRequests",
    "message": "Please wait 30 seconds before requesting Sagemaker URL"
}
```

Before the fix this was what we returned
```
500 error
{
    "requestId": "",
    "code": "internalError",
    "message": "Could not obtain a lock"
}
```

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
+++ b/addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
@@ -383,46 +383,54 @@ class EnvironmentScConnectionService extends Service {
 
   async createPrivateSageMakerUrl(requestContext, envId, connection, presign_retries = 10) {
     const lockService = await this.service('lockService');
-    const signedURL = await lockService.tryWriteLockAndRun({ id: `${envId}presign` }, async () => {
-      if (!(_.toLower(_.get(connection, 'type', '')) === 'sagemaker')) {
-        throw this.boom.badRequest(
-          `Cannot generate presigned URL for non-sagemaker connection ${connection.type}`,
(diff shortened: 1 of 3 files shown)


addons/addon-base-raas/packages/base-raas-services/lib/environment/service-catalog/environment-sc-connection-service.js
```diff
@@ -383,46 +383,54 @@ class EnvironmentScConnectionService extends Service {
...
+          const oldPolicyDocument = decodeURIComponent(currentPolicyResponse.PolicyDocument);
+          await this.putRolePolicy(iam, connection, oldPolicyDocument);
+        }
+      });
+    } catch (e) {
+      if (this.boom.is(e, 'internalError') && e.message === 'Could not obtain a lock') {
+        throw this.boom.tooManyRequests('Please wait 30 seconds before requesting Sagemaker URL', true);
       }
-    });
+      throw e;
```
SanketD92, line 432: Are we throwing the stack trace itself? Rather could we throw a boom error with just the `e.code` if required?

nguyen102: This line will rethrow what was already thrown. In the code, it looks like that will be `boom` errors. That means there won't be a stack trace.

SanketD92 (approved):

PR #1028, opened 2022-08-31
fix: remove permission for Guest to list users

nguyen102:
Issue #, if available:

Description of changes:
fix: remove permission for Guest to list users

Checklist:

AS review ticket id:


kpark277 (commented):
Does this break any e2e tests? So guests can still reach the Users tab, but they will just get an empty table?

kpark277 (dismissed):

addons/addon-base/packages/services/lib/user/user-authz-service.js
```diff
@@ -49,6 +51,13 @@ class UserAuthzService extends Service {
   }
 
   // Protected methods
+  async authorizeList(requestContext) {
+    if (requestContext.principal.userRole === 'guest') {
```
Bingjiling: Is there a differentiation between internal v.s. external guest here?

SanketD92: Yes, both are different. @nguyen102 should we add `internal guest` here as well?

nguyen102: Good catch. I updated the code to deny access for internal-guest as well

SanketD92 (approved):

PR #1031, opened 2022-09-02
fix: allow Guest users to log in

nguyen102:
Issue #, if available:

Description of changes:
By default when a user logs in, the UI call the GET `/users` API. Since we're now restricting Guest roles for calling the GET `/users` API, the UI returns an error. When the `userStore` can't get the users list it is supposed to set the user list to be an empty array. The code didn't properly handle that use case, and this PR fixes that. 

**Testing**
* Guest can log in and create SSH key (only functionality that Guest can actually do)
* Researcher can log in and view/edit study permissions
* Admin can log in and edit studies. Can view and edit users

Checklist:

-  [x] Have you written new tests for your core changes, as applicable?

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/users/UsersStore.js
+++ b/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/users/UsersStore.js
@@ -33,7 +33,12 @@ const UsersStore = BaseStore.named('UsersStore')
 
     return {
       async doLoad() {
-        const users = (await getUsers()) || [];
+        let users = [];
+        try {
+          users = await getUsers();
+        } catch (e) {
+          console.error('Could not get users', e);
+        }
         self.runInAction(() => {
           users.forEach(user => {
             const userModel = User.create(user);
--- a/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/users/UsersStore.js
+++ b/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/users/UsersStore.js
@@ -37,7 +37,10 @@ const UsersStore = BaseStore.named('UsersStore')
         try {
           users = await getUsers();
         } catch (e) {
-          console.error('Could not get users', e);
+          console.error(
+            'Could not get users. This is expected if you have a Guest role, but is not expected behavior for any other roles',
+            e,
+          );
         }
         self.runInAction(() => {
           users.forEach(user => {


addons/addon-base-raas-ui/packages/base-raas-ui/src/models/users/UsersStore.js
```diff
@@ -33,7 +33,12 @@ const UsersStore = BaseStore.named('UsersStore')
 
     return {
       async doLoad() {
-        const users = (await getUsers()) || [];
```
nguyen102, line 36: `await getUsers()` needs to be surrounded by a `try, catch` otherwise the error will bubble up and an error screen will be displayed. 
You are not authorized to list users

awsdrbalaji:
Let's make sure we run test scenarios manually or automated to make sure there are no side effects of this fix.

addons/addon-base-raas-ui/packages/base-raas-ui/src/models/users/UsersStore.js
```diff
@@ -33,7 +33,12 @@ const UsersStore = BaseStore.named('UsersStore')
...
+        let users = [];
+        try {
+          users = await getUsers();
+        } catch (e) {
+          console.error('Could not get users', e);
```
maghirardelli: So when a guest is logged in, this will log an error to the logs but it is not an error. 

Plus, if a non guest logs in and for whatever reason the `getUsers()` call fails, it will not bubble to the UI, which is not the right behavior for a valid error. Is there any way to add these conditions?

nguyen102: This code replicates the intended behavior of the original code by setting the users list to `[]` if `getUsers` fails. 

For non guest users, they should be able to call the user list API successfully and get the list of users. In the case that they can't, they're not going to be able to do anything with the error shown in the UI, so that's why I think logging the error to the console is preferable. That's what I saw was being done with the [StudiesStore](https://github.com/awslabs/service-workbench-on-aws/blob/9e224ff1aee34ccbdea14f330c7c840a9a39b125/addons/addon-base-raas-ui/packages/base-raas-ui/src/models/studies/StudiesStore.js#L135). 

To resolve the Guest use case, maybe I can specify in the `error` message that if they're a Guest, it is expected that they won't be able to call the list users API?

SanketD92 (approved):

maghirardelli (approved):
Thanks for updating the message! Looks good!

PR #1035, opened 2022-09-14
chore: Request package replace

FernandoAranda:
Issue #, if available:

Description of changes:
Replace request package with fetch 
Checklist:
- [x ] Have you successfully deployed to an AWS account with your changes?
- [x ] Have you successfully tested with your changes locally?

AS review ticket id:


addons/addon-base-rest-api/packages/services/package.json
```diff
@@ -50,4 +50,4 @@
       "pre-commit": "pretty-quick --staged --pattern '**/*.*(js|jsx)'"
     }
   }
-}
+}
```
nguyen102, line 53: Nitpick: Add extra new lines in the files so GH doesn't complain

nguyen102 (commented):
Looks good. Can you also update the threshold specified in this file?
https://github.com/awslabs/service-workbench-on-aws/blob/mainline/scripts/check-dependency-vulnerabilities.sh

addons/addon-base-rest-api/packages/services/package.json
FernandoAranda: whenever I save the file with a new line it removes it, I will see what tool is doing that

FernandoAranda:
> Looks good. Can you also update the threshold specified in this file? https://github.com/awslabs/service-workbench-on-aws/blob/mainline/scripts/check-dependency-vulnerabilities.sh

should I set it to 4/0 as it was before?

FernandoAranda:
> Looks good. Can you also update the threshold specified in this file? https://github.com/awslabs/service-workbench-on-aws/blob/mainline/scripts/check-dependency-vulnerabilities.sh

Done!

nguyen102 (approved):

PR #1049, opened 2022-10-10
fix: add aws config to workflow

maghirardelli:
Issue #, if available:

Description of changes:
Adding a step to the GH Actions that explicitly sets the default values so the default AWS profile is created for serverless to work in the pipeline

Tested in my [sandbox GH Actions repo](https://github.com/maghirardelli/service-workbench-on-aws-github-actions/blob/develop/.github/workflows/configure.yml) first.

Checklist:

AS review ticket id:


aws-tyler (approved):
seems reasonable. fingers crossed!!

PR #1051, opened 2022-10-11
fix: add steps in pipeline to create aws creds file

maghirardelli:
Issue #, if available:

Description of changes:
Trying to fix the pipelines so that the new Serverless V3 syntax behaves with the aws credentials configuration within the pipeline environments.
This is option 2 where we are adding steps to manually create `~/.aws/credentials` files in the pipeline environments so that they can use the `default` profile within Serverless V3. The GH Actions pipeline step has been tested and works. The CodePipeline step has not been tested.

Checklist:

AS review ticket id:


main/cicd/cicd-pipeline/config/buildspec/buildspec.yml
```diff
@@ -21,6 +21,16 @@ phases:
       - echo "Downloading environment config"
       - ./scripts/download-env-config.sh "$DEPLOYMENT_BUCKET" "$ENV_NAME"
       - printf "\n\n"
+      - echo "Creating ~/.aws/credentials"
```
nguyen102, line 24: I think you need to add something similar for `buildspec-int-test.yml`

.github/workflows/deploy-integ-appstream-egress.yml
```diff
@@ -36,6 +36,10 @@ jobs:
           go-version: 1.13
```
nguyen102, line 36: On line 33 and the other lines that specify node-version: 12, let's use node v16.
https://github.com/actions/setup-node

main/cicd/cicd-pipeline/config/buildspec/buildspec.yml
maghirardelli: There are no `sls` commands in the script that build runs([run-integration-tests.sh](https://github.com/awslabs/service-workbench-on-aws/blob/mainline/scripts/run-integration-tests.sh)). Why do you think we need to do something similar?

.github/workflows/deploy-integ-appstream-egress.yml
maghirardelli: Updated!

main/cicd/cicd-pipeline/config/buildspec/buildspec.yml
nguyen102: nevermind, you're right. Resolving this comment.

nguyen102 (approved):

PR #1157, opened 2023-03-21
fix: BYOB Study Permission Issue fix

jxuamazon:
Issue #, if available: 1067 

Description of changes:
Current implementation send the whole list of all users who have permission to the study for process. Should be just the delta (added/moved) users. 

When backend gets the user list, it will remove permission for all users first and then add the new list of uses back with permissions (delta).  However, when the last user is removed, the IAM role in the source bucket account also gets deleted, while that role is still being used by users within workspaces. 

The fix will only process the delta users, and the role will not be deleted , if there are still users attached to that role. 

**One Caveat - important** , please use this  to evaluate the security impact :   if an user is currently using this study in an active workspace, this user's access to the study will not be revoked when this user's permission is removed. After the workspace is restarted, then the user will lose access to the study.   So when admins remove permission for a user, they need to stop/start the workspaces that's using the study for that user.  **This needs to be documented.**    

Checklist:

AS review ticket id:


maghirardelli:
Hi @jxuamazon! We have reviewed and discussed this PR and **decided we will not be merging in this PR**. This is due to a few factors. We are not going to merge this PR since it changes the security posture of the BYOB feature and this is a documented limitation of the BYOB feature (as pointed out in the comments of issue #1067 ).

Thank you for working with us!

Issue #1004, opened 2022-06-22
[Bug] Cannot mount cross region open data


dengmingtong:
**Describe the bug**
Suppose my swb is in us-east-1 region, I create a linux workspace and try to mount https://registry.opendata.aws/hecatomb/ open data which is hosted on S3 us-west-2 region.

The workspace can be created successfully, but the hecatomb data cannot be mounted to the EC2 instance.

I think it should be a common bug when swb try to mount cross region data.
I did some debug as below:
Below is S3Mount parameter in cloudformation
[{"id":"hecatomb","readable":true,"writeable":false,"bucket":"hecatombdatabases","prefix":"/"}]

in the main/solution/post-deployment/config/environment-files/bin/mount_s3.sh file, there is code as below:
        if [ "$s3_role_arn" == "null" ]
        then
            printf 'Mounting internal study "%s" at "%s"\n' "$study_id" "$study_dir"
            goofys --region $region --acl "bucket-owner-full-control" "${s3_bucket}:${s3_prefix}" "$study_dir"

The region is the workspace region, not s3 bucket region. 
I think this cause the error of mounting data.

Could you try to fix this bug?

Thanks
Mingtong

ssvegaraju:
Hi @dengmingtong, thanks for bringing up this issue! I've added this to our team's backlog.

SanketD92:
Hi @dengmingtong, have you enabled TRE (AppStream) in your SWB deployment? Cross-region study mounting is not supported in TRE environments.

dengmingtong:
No, I did not enable TRE feature. You can repeat this issue when you mount another regions data.

dengmingtong:
any update?

tdmalone:
This still appears to be a problem in the latest version of SWB; it has already spawned two possible duplicate issues too, in #1013 and #1045.

Appears to be quite an oversight that mounting of open data is only supported in the same region...

BoPeng:
I have also encountered this bug. There are 119 open data on SWB but none of them appears to be available to `us-east-2`.  Please

1. If this is a limitation of SWB, can we list the region of the datasets so that we know what datasets are available? 
2. @SanketD92 mentioned TRE (AppStream) could solve this issue. Could you elaborate how AppStream helps and how to enable AppStream?

Edit: The following is the configuration flag about AppStream, which says enable AppStream is a one-way process and we should proceed with caution. It would be helpful if someone can point me to the section in the documentation that talks about AppStream.

[...]

PR #1179, opened 2023-04-19
Mount s3 bucket without option --region

BoPeng:
#1004

Description of changes:

Since `goofys` will fail to mount s3 bucket if the wrong `--region` is specified, and will usually be able to detect `--region` automatically, this patch calls `goofys` without option `--region` and calls `goofys` again with option `--region` if it fails.

Checklist:

AS review ticket id:


aws-tyler:
Hi Bo! We're going to need to wait on merging this so I'm going to close it for now and re-open it when we're ready. Thanks for your patience with this.

PR #1186, opened 2023-04-28
[chore] Add conditionals for TRE permission boundaries

aws-tyler:
Added check to cloudformation templates to only include boundary permission updates if AppStream is enabled.

Testing (in Linux, Windows, and SageMaker):
* Verified that credentials being used OUTSIDE of the AppStream environment could not access S3 data.
* Verified that credentials being used INSIDE of the AppStream environment could access studies.


SanketD92 (approved):
Could you also mention in the PR description the testing performed to verify these changes?

PR #1209, opened 2023-07-18
Support chunking with batchGet in getter.js

### LINKED ISSUE #1180: [Bug] A user cannot be in more than 100 studies at a time
**Describe the bug**

If a user is in more than 100 studies as an admin or a user, the Studies page will fail to load for that user, giving a validation error from DynamoDB BatchGetItem in the returned JSON. From what I understand, the first query which finds which studies a user has access to, succeeds. Only the fetch of the details for these studies fails.

**To Reproduce**

1. Add a user as an admin to 100 studies
2. Try to load the Studies page

**Expected behavior**

As an administrator using SWB, I would expect the Studies page to be displayed.

**Versions (please complete the following information):**

We are using a fork from an earlier version, but the code responsible for this issue has not changed.

**Additional context**

I have been able to patch the issue on our deployment by breaking up the BatchGetItem requests into chunks of 100.

The code which performs the get batch get operation in your main branch:

https://github.com/awslabs/service-workbench-on-aws/blob/1d2fe14e04218070aa7034a9868002643da88bc6/addons/addon-base/packages/services/lib/db/getter.js#L121-L141

Our temporary patch for this:

```js
  async get() {
    if (this.params.Key && this.params.Keys)
      throw new Error('DbGetter <== only key() or keys() may be called, not both');

    let data = [];
    if (this.params.Key) data = (await this.client.get(this.params).promise()).Item;
    else if (this.params.Keys) {
      // BatchGet supports a maximum of 100 keys, since this becomes
      // possible within HIC TRE, we want to break it up into chunks.
      for (let keySet of _.chunk([ ...this.params.Keys ], 100)) {
	this.params.Keys = keySet;

	const batchParams = { RequestItems: {} };
	batchParams.RequestItems[this.params.TableName] = { ...this.params };

	if (this.params.ReturnConsumedCapacity) {
          batchParams.RequestItems.ReturnConsumedCapacity = this.params.ReturnConsumedCapacity;
          delete batchParams.RequestItems[this.params.TableName].ReturnConsumedCapacity;
	}

	data.push(...(await this.client.batchGet(batchParams).promise()).Responses[this.params.TableName]);
      }
    }

    return unmarshal(data);
  }
```

I suspect BatchGetItem is the wrong approach and that a DynamoDB Scan might be more appropriate.

Thanks

AaronJackson:
Fixes #1180

> If a user is in more than 100 studies as an admin or a user, the Studies page will fail to load for that user, giving a validation error from DynamoDB BatchGetItem in the returned JSON. From what I understand, the first query which finds which studies a user has access to, succeeds. Only the fetch of the details for these studies fails.

Description of changes:

This enables chunking of DynamoDB BatchGet requests if there are more than 100 results. I'm not sure if this is the most appropriate solution as I think a Scan should be used, but it was a quick fix for our SWB deployment.

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base/packages/services/lib/db/getter.js
+++ b/addons/addon-base/packages/services/lib/db/getter.js
@@ -122,19 +122,24 @@ class DbGetter {
     if (this.params.Key && this.params.Keys)
       throw new Error('DbGetter <== only key() or keys() may be called, not both');
 
-    let data;
+    let data = [];
     if (this.params.Key) data = (await this.client.get(this.params).promise()).Item;
     else if (this.params.Keys) {
-      // BatchGet
-      const batchParams = { RequestItems: {} };
-      batchParams.RequestItems[this.params.TableName] = { ...this.params };
+      // BatchGet supports a maximum of 100 keys, since this becomes
+      // possible within HIC TRE, we want to break it up into chunks.
+      for (let keySet of _.chunk([ ...this.params.Keys ], 100)) {
+	this.params.Keys = keySet;
 
-      if (this.params.ReturnConsumedCapacity) {
-        batchParams.RequestItems.ReturnConsumedCapacity = this.params.ReturnConsumedCapacity;
-        delete batchParams.RequestItems[this.params.TableName].ReturnConsumedCapacity;
+	const batchParams = { RequestItems: {} };
+	batchParams.RequestItems[this.params.TableName] = { ...this.params };
+
+	if (this.params.ReturnConsumedCapacity) {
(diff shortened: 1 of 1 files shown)


maghirardelli:
Thank you for reaching out. Your request has been added to our backlog for research computing solutions. Our
team curates these requests for fit with our solutions vision on a regular basis. Please watch this space for new
updates. If you require immediate assistance, please reach out to your AWS account team. Please note that
security issues should be reported directly to AWS Security at [aws-security@amazon.com](mailto:aws-security@amazon.com).

PR #1215, opened 2023-07-26
feat: add IMDSv2 support on AMI creation

maghirardelli:
Issue #, if available:

Description of changes:
This adds IMDSv2 support to the AMIs created by running `pnpx sls build-image -s <STAGE>`. To apply these changes to your environment, rerun this command update the AMI IDs in the Workspace Type Configs with the newly created AMIs.

Checklist:

AS review ticket id:

### DIFF
--- a/addons/addon-base-raas/packages/serverless-packer/index.js
+++ b/addons/addon-base-raas/packages/serverless-packer/index.js
@@ -33,7 +33,7 @@ class ServerlessPackerPlugin {
     this.commands = {
       'build-image': {
         usage: 'Build an AMI using packer',
-        lifecycleEvents: ['build'],
+        lifecycleEvents: ['build', 'add-imds-v2-support'],
         options: {
           files: {
             usage:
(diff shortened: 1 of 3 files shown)


addons/addon-base-raas/packages/serverless-packer/index.js
```diff
@@ -89,6 +91,71 @@ class ServerlessPackerPlugin {
...
+    const customSettings = this.serverless.service.custom.settings;
+    if (customSettings.enableAmiSharing) {
+      customSettings.awsProfile = customSettings.devopsProfile;
+    }
+    const awsProfile = customSettings.awsProfile;
+    const awsRegion = customSettings.awsRegion;
+
+    // For each AMI, add IMDSv2 support
+    return Promise.all(
+      _.forEach(amis, async ami => {
```
SanketD92: Use `_.map` instead of `_.forEach` when dealing with async [operations](https://stackoverflow.com/questions/68026591/what-is-the-difference-between-async-await-foreach-and-promise-all-map)

janeyuaws (dismissed):

addons/addon-base-raas/packages/serverless-packer/index.js
maghirardelli: good point we've dealt with this causing bugs in the past so I will align with this

SanketD92 (approved):

Issue #1237, opened 2023-11-04
[Misc] babelrc config to make deployment a bit quieter


mrgum:
**What's on your mind?**
I deploy swb in codebuild as part of a code pipeline so I get the full logs rather than the scrolling thing you get in cli, there are hundreds, possibly thousands of lines like these repeated

```
addons/addon-base-ui/packages/base-ui build: Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-property-in-object since the "loose" mode option was set to "true" for @babel/plugin-proposal-class-properties.
addons/addon-base-ui/packages/base-ui build: The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
addons/addon-base-ui/packages/base-ui build:    ["@babel/plugin-proposal-private-property-in-object", { "loose": true }]
addons/addon-base-ui/packages/base-ui build: to the "plugins" section of your Babel config.
```

This can be fixed by using .babelrc containing

```
{
  "plugins": [
    ["@babel/plugin-proposal-class-properties", { "loose": true }],
    ["@babel/plugin-proposal-private-methods", { "loose": true }],
    ["@babel/plugin-proposal-private-property-in-object", { "loose": true }]
  ],
  "presets": [
    "@babel/preset-react",
    ["@babel/preset-env", {
      "loose": true
    }]
  ]
}
```

PR on its way

I know little to nothing about react/babel etc so please can someone who does review :)

**Versions (please complete the following information):**
 - Release Version installed git mainline

SanketD92:
Thank you for reaching out. Your request has been added to our backlog for research computing solutions. Our
team curates these requests for fit with our solutions vision on a regular basis. Please watch this space for new
updates. If you require immediate assistance, please reach out to your AWS account team. Please note that
security issues should be reported directly to AWS Security at [aws-security@amazon.com](mailto:aws-security@amazon.com).

mrgum:
the babel config I used above breaks part of the UI and reading up I think loose set false is what should be used so either
```
{
  "plugins": [
    ["@babel/plugin-proposal-class-properties", { "loose": false }],
    ["@babel/plugin-proposal-private-methods", { "loose": false }],
    ["@babel/plugin-proposal-private-property-in-object", { "loose": false }]
  ],
  "presets": [
[...]

PR #1239, opened 2023-11-07
@babel/plugin-proposal-class-properties loose should be false

mrgum:
Issue #1237

Description of changes:
This change removes the loose="true" setting on @babel/plugin-proposal-class-properties which was not being applied anyway and caused thousands of warning lines in the logs

Checklist:

- [x] Is this change also required on the AWS Solution version?


mrgum:
this makes some tickboxes untickable so I withdraw it

